PR #3, opened 2016-06-02
Fix building on Windows

gabrielesvelto:
This is a quick hack to ensure rust-minidump builds correctly under Windows. It's not pretty since minidump_format.rs is generated but it works. Let me know if you think there's a better approach to tackle the issue with type sizes across different architectures.


luser:
Yeah, this isn't fantastic but it's workable for now. Longer-term either rust-bindgen needs to get smarter about `stdint.h` (by just mapping `XintN_t` -> `XN`), or minidump_format.rs should become hand-authored, which might be nice, since the generated code isn't great and some of the stuff inherited from C sucks.

PR #4, opened 2017-12-24
Use static dispatch for Minidump::reader.

frewsxcv:

PR #5, opened 2017-12-24
Use non-static parameterized lifetime instead of requiring static lifetime.

frewsxcv:


frewsxcv:
Alternative to https://github.com/luser/rust-minidump/pull/4

luser:
Thanks! I like #4 slightly better, so let's just do that.

Issue #17, opened 2018-09-18
Support aarch64 context matching Microsoft's definition


luser:
Breakpad has made some changes to rename its aarch64 context structs and flags:
https://chromium.googlesource.com/breakpad/breakpad/+/627ef0cb9c705444da6ca7515bc5ca6b3169e1aa%5E%21/

Microsoft is shipping Windows on aarch64 devices now, so the Windows SDK has a definition for an aarch64 context, and they're going to add support for that. We should similarly support those structs.

PR #39, opened 2019-02-07
add aarch64 support

froydnj:
This series gets me to the point where I can `minidump_stackwalk` and have at least the crashing thread line up with what crash-stats tells me.  I have an in-progress frame pointer unwind for aarch64, but that's not lining up with what crash-stats says, so we'll leave that bit for later.


luser:
This fixes #17

minidump-common/src/format.rs
```diff
@@ -856,11 +856,15 @@ pub struct FLOATING_SAVE_AREA_ARM64 {
 /// in WinNT.h.
```
luser: We should change the doc comment here to match reality.

luser:
Thanks!

PR #40, opened 2019-02-15
use new-style `?` syntax instead of `try!`

froydnj:
The new style was used in a couple places; this change makes things more uniform.


luser:
Thanks! I started using `?` at some point but never got around to going back and changing most of the existing code.

PR #54, opened 2019-05-29
handle "old" aarch64 exception contexts

froydnj:
Turns out these are necessary to get reasonable things to happen for Android minidumps.

### DIFF
--- a/minidump-common/src/format.rs
+++ b/minidump-common/src/format.rs
@@ -845,9 +845,9 @@ pub enum ArmRegisterNumbers {
 /// aarch64 floating point state
 #[derive(Clone, Pread, SizeWith)]
 pub struct FLOATING_SAVE_AREA_ARM64 {
+    pub regs: [u128; 32usize],
     pub fpsr: u32,
     pub fpcr: u32,
-    pub regs: [u128; 32usize],
 }
 
 /// An aarch64 (arm64) CPU context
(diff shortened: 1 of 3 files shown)


minidump-common/src/format.rs
```diff
@@ -845,9 +845,9 @@ pub enum ArmRegisterNumbers {
 /// aarch64 floating point state
 #[derive(Clone, Pread, SizeWith)]
 pub struct FLOATING_SAVE_AREA_ARM64 {
```
luser: FYI, Breakpad defines two versions of this struct, an "old" and a "new":
https://chromium.googlesource.com/breakpad/breakpad/+/88d8114fda3e4a7292654bd6ac0c34d6c88a8121/src/google_breakpad/common/minidump_cpu_arm64.h#125
https://chromium.googlesource.com/breakpad/breakpad/+/88d8114fda3e4a7292654bd6ac0c34d6c88a8121/src/google_breakpad/common/minidump_cpu_arm64.h#74

The old one has `regs` first in this struct, the new one has it last. Presumably the latter matches the struct from Microsoft's SDK. I suppose ideally this code would handle both but realistically I have seen very little usage of fp register data from minidumps historically.

minidump-common/src/format.rs
```diff
@@ -842,6 +843,27 @@ pub enum ArmRegisterNumbers {
     ProgramCounter = 15,
 }
 
+/// aarch64 floating point state (old)
+#[derive(Clone, Copy, Pread, SizeWith)]
+pub struct FLOATING_SAVE_AREA_ARM64_OLD {
```
luser: Oh, sorry, I was looking at a single changeset and didn't see this in the next commit...

PR #55, opened 2019-05-29
change get-minidump-instructions to disassemble from crashing ip

froydnj:
The current implementation disassembles the entire memory region
containing the crashing ip.  But the entire region might not be
disassemblable (sp?): the region might start in the middle of a variable
length instruction or might contain just plain garbage in the case of
fixed width instructions.  And the underlying disassembler seems to give
up entirely in the case of stuff it can't disassemble, rather than
trying to move on.

Instead, let's start disassembling from the byte indicated by the
crashing ip.  This will at least produce reasonable output until we can
change the underlying disassembler or correctly call into the
disassembler to make it continue in the presence of garbage.

### DIFF
--- a/minidump-tools/src/lib.rs
+++ b/minidump-tools/src/lib.rs
@@ -230,8 +230,8 @@ pub fn get_minidump_instructions() -> Result<(), Error> {
     println!("Faulting instruction pointer: {:#x}", ip);
     info!("Disassembling {} bytes starting at {:#x} of arch {:?}",
           memory.bytes.len(), memory.base_address, arch);
-    disasm::disasm_bytes(&memory.bytes,
-                         memory.base_address,
+    disasm::disasm_bytes(&memory.bytes[ip as usize - memory.base_address as usize..],
+                         ip,
                          arch,
                          color.unwrap_or(Color::Auto),
                          Some(ip),


luser:
This sucks. :frowning_face: Does the disassembler fail in a way that's detectable so that the code could try the previous behavior first and then try the new behavior if that fails? When I've used this tool in the past having the context surrounding the crashing instruction has been hugely valuable.

froydnj:
> This sucks. frowning_face Does the disassembler fail in a way that's detectable so that the code could try the previous behavior first and then try the new behavior if that fails? When I've used this tool in the past having the context surrounding the crashing instruction has been hugely valuable.

I agree, this is a hack. :(  I was helping Kannan with some arm64 troubles and the first thing he requested after I fixed this was "can I see instructions before the crashing instruction?"

`rust-disasm` uses [`Capstone::disassm_all`](https://docs.rs/capstone/0.6.0/capstone/struct.Capstone.html#method.disasm_all) which (well, at least in 0.6) returns an [`Instructions`](https://docs.rs/capstone/0.6.0/capstone/struct.Instructions.html) instance which claims to tell you how many instructions were disassembled.  For the aarch64 case, I could see keeping track of how much of the buffer you've disassembled and repeatedly calling `disasm_all` (or maybe `disasm_count`) until you reach the end of the buffer.

I don't have good ideas for x86ish processors (or arm + thumb, for that matter).

luser:
It's definitely a PITA. If we have function symbols I guess we could look for the start of the current function and if that falls within the memory region we have, start there. There really is no good answer for variable-length ISAs if we don't know a starting point. (Presumably a real debugger would have access to the actual binaries so this is much easier!)

PR #98, opened 2020-09-30
Re-export scroll::Endian

flub:
This is a public field on the Minidump struct, but the API currently
forces you to explicitly pull in scroll to use it.  It seems
reasonable for the crate to export the types which which is uses in
public APIs.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -32,6 +32,8 @@ extern crate scroll;
 #[cfg(test)]
 extern crate test_assembler;
 
+pub use scroll::Endian;
+
 mod context;
 mod iostuff;
 mod minidump;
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -40,6 +42,6 @@ pub mod synth_minidump;
 pub mod system_info;
 
 pub use iostuff::Readable;
+pub use minidump::*;
```
flub, line 45: This was re-ordered by rust-fmt which my editor runs automatically on every save.  Would you mind if I made PRs to format the entire project and add CI config to keep running cargo fmt, clippy, tests?

luser: That seems fine, yes. I've been historically lazy about ensuring formatting and such on my own personal projects. In my defense this project *does* predate a lot of Rust tooling. :)

luser (approved):
This is reasonable, thanks!

PR #99, opened 2020-10-08
Change multi_structs macro to compile on 1.47.0

jan-auer:
This is a quick workaround to fix compilation on Rust 1.47.0, which was released to the stable channel today. It seems that custom derives no longer resolve the syntax tree correctly when declared in a macro that uses `ty` matchers. When updated to `tt`, it works however.

I yet have to pin it down and then potentially report this upstream.

Fixes https://github.com/luser/rust-minidump/issues/100


luser (approved):
Thanks!

jan-auer:
Also fixing this in https://github.com/m4b/scroll/pull/75, which would make this PR redundant.

PR #102, opened 2020-10-13
Fixes for updated scroll dependency

flub:
The new scroll dependency introduced in
https://github.com/luser/rust-minidump/pull/101 has slightly
simplified traits.  Adapt the code to work with these.


luser (approved):
You would think I might have known this given that I am the one that made that change in scroll. :laughing:

luser:
Thank you!

PR #103, opened 2020-10-13
Fixes for new reqwest version

flub:
https://github.com/luser/rust-minidump/pull/97 updated the reqwest
version which by default uses async code now.  Migrate to the blocking
client feature instead.


luser:
Thanks! I should have actually looked at this more closely but I think I merged that update on top of the scroll bustage...

PR #106, opened 2020-10-15
Run cargo fmt and add ci

flub:
This adds some generic CI configuration and runs cargo-fmt on all the
code.  No code changes made.

For now still allow warnings, since there are a lot of them.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,97 @@
+name: Rust CI
+
+on:
+  pull_request:
+  push:
+    branches:
+      - master
+  schedule:
+    - cron: '11 7 * * 1,4'
+
+# env:
+#   RUSTFLAGS: -Dwarnings
+#   RUSTDOCFLAGS: -Dwarnings
+
+jobs:
+
+  fmt:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          profile: minimal
+          components: rustfmt
+          override: true
+      - uses: actions-rs/cargo@v1
+        with:
+          command: fmt
+          args: --all -- --check
+
+  clippy:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - uses: actions-rs/toolchain@v1
+        with:
+            toolchain: stable
+            profile: minimal
+            components: clippy
+            override: true
+      - uses: actions-rs/clippy-check@v1
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          args: --workspace --tests --examples
+
+  docs:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: nightly
+          profile: minimal
+          components: rust-docs
+          override: true
+      - uses: actions-rs/cargo@v1
+        with:
+          command: doc
+          args: --workspace --no-deps
+
+  test:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest, macOS-latest]
+        rust: [nightly, beta, stable]
+    steps:
+      - uses: actions/checkout@master
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: ${{ matrix.rust }}
+          profile: minimal
+          override: true
+      - uses: actions-rs/cargo@v1
+        with:
+          command: test
+          args: --workplace --tests --examples --bins
+  
+  doctest:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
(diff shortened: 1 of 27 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,106 @@
...
+          toolchain: nightly
+          profile: minimal
+          components: rust-docs
+          override: true
+      - uses: swatinem/rust-cache@v1
+        with:
+          key: ${{ github.job }}
+      - uses: actions-rs/cargo@v1
+        with:
+          command: doc
```
luser, line 62: If this isn't going to also run the doc upload script like the Travis config does then it's not really worth running this. I'm fine with dropping that entirely, I don't look at the nightly docs and I doubt anyone else does. docs.rs exists for releases and it's simple enough to run `cargo doc` locally for master.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,106 @@
...
+          toolchain: ${{ matrix.rust }}
+          profile: minimal
+          override: true
+      - uses: swatinem/rust-cache@v1
+        with:
+          key: ${{ github.job }}
+      - uses: actions-rs/cargo@v1
+        with:
+          command: test
+          args: --workspace --doc
```
luser, line 105: Is this split out to a separate job because that's how it currently is in the Travis config? I'm fuzzy on the details but looking at it I think it's only written that way because the `docmatic` crate had some problems, and I believe I'm only using that to test the examples in the `README`:

What would you think about swapping that out for the `doc-comment` crate (it's trivial to use), running `cargo test --all` in the `test` job above, and removing this separate doctest job?
https://docs.rs/doc-comment/0.3.3/doc_comment/

luser (changes requested):
A few minor things but otherwise I'm happy to land this! I haven't actually used GitHub Actions with any of my repos—do I need to do something to enable them here or will landing this be enough to enable them?

Thanks for doing this work! As someone who has done a whole lot of build automation and CI work professionally I know it's a thankless task. :)

.github/workflows/ci.yml
flub: Oh, I didn't realise there's a travis config :)  I probably agree that nightly docs aren't that important, I also didn't know they existed, I imagine most people use cargo doc for this.

The intention of this step is to ensure there are no bad links or formatting issues in the doc comments, together with the `RUSTDOCFLAGS=-Dwarnings` env this will fail if there are any warnings in the docs building.

.github/workflows/ci.yml
flub: I would prefer one job as well, but I couldn't figure out the flags that run everything at once.  It seems there are some tests that this runs which aren't run in the main test job.  I don't know the doc-comment crate, I'll check it out and see whether I can get rid of this step

flub:
> A few minor things but otherwise I'm happy to land this! I haven't actually used GitHub Actions with any of my repos—do I need to do something to enable them here or will landing this be enough to enable them?

There's Settings -> Actions where the actions can be controlled, but this defaults to on I think.  I think they're not running here because they come from a PR in a different repo and it probably only uses a config actually in this repo.  So once this is merged it should work again.
In the mean time I switched this back to draft and am actually trying the changes out at https://github.com/flub/rust-minidump/pull/1 as there they are actually being run.  As you see there's still some tweaking to be done ;)

> Thanks for doing this work! As someone who has done a whole lot of build automation and CI work professionally I know it's a thankless task. :)

Heh, the bulk of this is figured out from other projects already so it's not too bad :)

flub:
This now works on https://github.com/flub/rust-minidump/pull/1 and is a bit simpler.  There are two sections which are commented out: the env section to make rust stricter and the clippy section.  I'm aiming to enable those in the future as well but they require code changes and I wanted to keep those limited to just rustfmt changes for this PR.

The docs job doesn't actually do anything useful yet while RUSTDOCFLAGS isn't enabled, the purpose is to stop PRs from breaking intra-doc links or other syntax.  This runs on the nightly compiler as that allows the new-style intra-doc links which is so much nicer to use and seem to work on docs.rs as well.  If you don't like that we can change that to stable again.

The main tests run on stable and nightly, could also add beta.  Maybe more interesting would be to only run nightly (and maybe beta) for the scheduled jobs and not for PRs so that nightly breakages don't stop normal work.  On the other hand that's not supposed to happen often.

Lastly I added a step which runs tests in examples and binaries as well.  Currently there are no such tests, so happy to remove that if you prefer.

I can squash the commits to something more reasonable if you prefer as well.

.github/workflows/ci.yml
luser: For most of my crates I think I just run `cargo test --all` and it runs everything I've ever cared about? I'm not sure if there's something covered by `cargo doc` that doesn't get covered there.

luser (approved):
This looks good enough, your reasoning is sound and I'm not going to nitpick you further over tiny details on this lovely volunteer work you've done. I'll merge it as-is, I'm not overly concerned about the commit history being pristine on small projects like this. If you have additional improvements in the future I will gladly merge those as well!

PR #107, opened 2020-10-19
Enable all CI

flub:
Clippy tricked me into thinking there wouldn't be much it complained about... but that was a lie.

Most of this is boring but nice to get clippy's consistency I guess.  A few things stand out though:

* I added a few `Default` impls because clippy wanted those.  That's a public API addition so perhaps deserves some more careful consideration.  We could disable these warnings if better for now.  Though if `::new()` were to change on these it would also be a public API change so perhaps than no longer being able to use `Default` or it having different semantics at that point is fine.

* I added a few `allow(clippy::should_implement_trait)` attributes, clippy complains that those methods have expectations in common traits (`std::ops::Add` in this case) and should either impl that trait or use different names.  I think using different names is probably better here, but that breaks the public API so I'm not doing that.

Happy to squash etc if you'd like, feel free to be critical.

### DIFF
--- a/minidump-processor/tests/test_processor.rs
+++ b/minidump-processor/tests/test_processor.rs
@@ -82,7 +82,7 @@ fn test_processor() {
         assert_eq!(raw.eip, 0x0040429e);
         assert_eq!(*valid, MinidumpContextValidity::All);
     } else {
-        assert!(false, "Wrong context type");
+        panic!("Wrong context type");
     }
 
     // Check thread 0, frame 3.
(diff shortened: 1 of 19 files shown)


minidump-processor/tests/test_processor.rs
```diff
@@ -82,7 +82,7 @@ fn test_processor() {
         assert_eq!(raw.eip, 0x0040429e);
         assert_eq!(*valid, MinidumpContextValidity::All);
     } else {
-        assert!(false, "Wrong context type");
+        panic!("Wrong context type");
```
luser, line 85: I'm not morally opposed to this but `panic!` just reads oddly in the context of tests. (I realize that `assert!` uses `panic!` internally.)

src/minidump.rs
```diff
@@ -767,6 +768,12 @@ impl MinidumpModuleList {
     }
 }
 
+impl Default for MinidumpModuleList {
```
luser, line 771: This seems fine. I'm not sure that there are really valid semantics for `Default` here, but considering that `new` already hands out an empty list this isn't much different.

luser:
Thanks! I'm struck by how much of this is code that wouldn't have worked in older versions of Rust (which were current when I wrote the bulk of the code). It's interesting to see the evolution of Rust in such a direct way!

flub:
> I'm struck by how much of this is code that wouldn't have worked in older versions of Rust (which were current when I wrote the bulk of the code). It's interesting to see the evolution of Rust in such a direct way!

Yeah, I've noticed that myself as well.  The language moves surprisingly quickly!

PR #109, opened 2020-10-19
Update to 2018 edition

flub:
As you seem to be happy updating the code, I've adopted this to the 2018 edition.  Other than the strictly required `cargo fix --edition` this also replaces the `extern crate` statements with explicit `use` statements.

I realise this is a bit of a gratuitous change, so I won't mind at all if you'd rather not update the edition.


luser (approved):
Thank you! Sorry for leaving this open for so long.

PR #110, opened 2020-10-19
Add missing debug implementations

flub:
This adds Debug derives to most of the library public API.  I have on
occasion missed not having these.

Derives are for now probably good enough, they can be improved as
needed.

----
This in on top of the edition-2018 branch so includes those commits as long as they're not merged.  Only the last commit here is relevant.  It's easy enough to rebase if that branch won't be merged so just let me know.


luser:
If you rebase this I'm happy to merge it (or you can do so yourself since I just sent you an invitation to be a collaborator).

PR #111, opened 2020-10-21
Fix memory region and stream lifetimes

flub:
Self can directly be of lifetime 'a, constraining it to a different
lifetime does not seem to gain any more flexibility.  This simplifies
the signature and thus use.

I don't think there are backwards compatibility concerns with this,
everything which was possible should still be possible.  The lifetime
of the returned stream is still the same 'a as before.

### DIFF
--- a/src/minidump.rs
+++ b/src/minidump.rs
@@ -927,24 +927,24 @@ Memory
 
 /// An iterator over `MinidumpMemory`s.
 #[allow(missing_debug_implementations)]
-pub struct MemoryRegions<'b, 'a> {
-    iter: Box<dyn Iterator<Item = &'b MinidumpMemory<'a>> + 'b>,
+pub struct MemoryRegions<'iter, 'data> {
+    iter: Box<dyn Iterator<Item = &'iter MinidumpMemory<'data>> + 'iter>,
 }
 
-impl<'b, 'a> Iterator for MemoryRegions<'b, 'a>
+impl<'iter, 'data> Iterator for MemoryRegions<'iter, 'data>
 where
-    'a: 'b,
+    'data: 'iter,
 {
-    type Item = &'b MinidumpMemory<'a>;
+    type Item = &'iter MinidumpMemory<'data>;
 
-    fn next(&mut self) -> Option<&'b MinidumpMemory<'a>> {
+    fn next(&mut self) -> Option<&'iter MinidumpMemory<'data>> {
         self.iter.next()
     }
 }
 
-impl<'a> MinidumpMemoryList<'a> {
+impl<'mdmp> MinidumpMemoryList<'mdmp> {
     /// Return an empty `MinidumpMemoryList`.
-    pub fn new() -> MinidumpMemoryList<'a> {
+    pub fn new() -> MinidumpMemoryList<'mdmp> {
         MinidumpMemoryList {
             regions: vec![],
             regions_by_addr: RangeMap::new(),
@@ -952,7 +952,7 @@ impl<'a> MinidumpMemoryList<'a> {
     }
 
     /// Create a `MinidumpMemoryList` from a list of `MinidumpMemory`s.
-    pub fn from_regions(regions: Vec<MinidumpMemory<'a>>) -> MinidumpMemoryList<'a> {
+    pub fn from_regions(regions: Vec<MinidumpMemory<'mdmp>>) -> MinidumpMemoryList<'mdmp> {
         let regions_by_addr = regions
             .iter()
             .enumerate()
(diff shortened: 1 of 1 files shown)


flub:
This is not wrong, but it exposes more subtle underlying problems.  This ends up exposing that the lifetime returned is the lifetime of the Minidump and not of the data.  I think this can be made clearer a bit with lifetime names etc so will try and improve this some more.

flub:
This is now updated to explain what goes on with the lifetime of streams.  Plus it now also fixes a bug in the iteration of the memory regions which did get the wrong lifetimes.  I've given some lifetimes more descriptive names in an attempt to make them clearer.  This is probably not overly consistent with the rest of the crate and a hard call of when to do this and when not, so I'd be fine with reverting that part.

src/minidump.rs
```diff
@@ -866,35 +866,35 @@ Memory
...
+    type Item = &'iter MinidumpMemory<'data>;
 
-    fn next(&mut self) -> Option<&'b MinidumpMemory<'a>> {
+    fn next(&mut self) -> Option<&'iter MinidumpMemory<'data>> {
         self.iter.next()
     }
 }
 
-impl<'a> MinidumpMemoryList<'a> {
+impl<'mdmp> MinidumpMemoryList<'mdmp> {
```
flub, line 887: If this naming is adopted perhaps this should be done for all of the stream types.  On the other hand perhaps the doc comment on `.get_stream()` is sufficient and we should revert this rename?

src/minidump.rs
```diff
@@ -907,21 +907,26 @@ impl<'a> MinidumpMemoryList<'a> {
...
+    /// the lifetime of the [Minidump] struct ('mdmp).
+    pub fn iter<'slf>(&'slf self) -> MemoryRegions<'slf, 'mdmp> {
         MemoryRegions {
             iter: Box::new(self.regions.iter()),
         }
     }
 
     /// Iterate over the memory regions in order by memory address.
-    pub fn by_addr<'b>(&'b self) -> MemoryRegions<'a, 'b> {
+    pub fn by_addr<'slf>(&'slf self) -> MemoryRegions<'slf, 'mdmp> {
```
flub, line 929: This is actually the bug fix, these were swapped (a bit hard to see due to the rename)

src/minidump.rs
```diff
@@ -907,21 +907,26 @@ impl<'a> MinidumpMemoryList<'a> {
...
     }
 
     /// Iterate over the memory regions in the order contained in the minidump.
-    pub fn iter<'b>(&'b self) -> MemoryRegions<'a, 'b> {
+    ///
+    /// The iterator returns items of [MinidumpMemory] as `&'slf MinidumpMemory<'mdmp>`.
+    /// That is the lifetime of the item is bound to the lifetime of the iterator itself
+    /// (`'slf`), while the slice inside [MinidumpMemory] pointing at the memory itself has
+    /// the lifetime of the [Minidump] struct ('mdmp).
+    pub fn iter<'slf>(&'slf self) -> MemoryRegions<'slf, 'mdmp> {
```
flub, line 922: This is actually the bug fix, the lifetimes of `MemoryRegions` where swapped.  The fix is harder to see due to the rename, but it's easier to read this correctly with the new names.

src/minidump.rs
luser: This seems very sensible and I would be happy with that. With a single named lifetime things are usually fine but as you've demonstrated here, adding a second one without naming them clearly leaves room for confusion!

src/minidump.rs
```diff
@@ -1581,11 +1586,14 @@ where
     /// [`MinidumpStream`][stream] trait, this method allows reading
     /// the stream data as a specific type.
     ///
+    /// Note that the lifetime of the returned stream is bound to the lifetime of the this
+    /// `Minidump` struct itself and not to the lifetime of the data backing this minidump.
+    /// This is a consequence of how this struct relies on [Deref] to access the data.
```
luser, line 1591: It seems like we can fix this to actually do the sensible thing, right? We'd just have to add another named lifetime and use that for `&self`, but use `'a` (which should probably be renamed to `'mdmp` as you've done elsewhere) to correctly tie the `MinidumpStream`'s lifetime to the data. That will probably also require telling rustc about the relation between the two lifetimes.

luser (approved):

src/minidump.rs
flub: Whoops, this PR completely escaped me.  Apologies.

Don't you describe how the lifetimes were organised before this change?  And there the lifetime of the reference returned with deref is always the same lifetime as the Minidump struct itself has IIUC.  The Minidump struct does tie all these lifetimes together with the lifetime specifiers on the deref and phantomdata.

PR #116, opened 2020-11-09
Use intra-doc links in the format module

flub:
This turns a number of names into new-style intra-doc links.  These
need the nightly compiler currently, but doc.rs already supports this
and we also test for this in the CI with the nightly compiler.

Leaves existing doc links intact.


minidump-common/src/format.rs
```diff
@@ -673,11 +673,15 @@ pub enum ExceptionCodeMac {
     SIMULATED = 0x43507378,
 }
 
-/// Valid bits in a `context_flags` for `ContextFlagsCpu`
+/// Valid bits in a `context_flags` for [`ContextFlagsCpu`]
```
luser, line 676: I might reword this to be "Valid bits in [`ContextFlagsCpu::context_flags`]"

minidump-common/src/format.rs
```diff
@@ -806,8 +810,8 @@ pub struct CONTEXT_AMD64 {
     /// This is defined as a union in the C headers, but also
     /// ` MAXIMUM_SUPPORTED_EXTENSION` is defined as 512 bytes.
     ///
-    /// Callers that want to access the underlying data can use `Pread` to read either
-    /// an `XMM_SAVE_AREA32` or `SSE_REGISTERS` struct from this raw data as appropriate.
+    /// Callers that want to access the underlying data can use [`Pread`] to read either
```
luser, line 813: Does this link to `Pread` work given that it comes from the `scroll` crate?

luser (approved):
Thanks for these changes, this seems like it'll be much nicer! This doesn't cause any issues with `cargo doc` on a non-nightly compiler, does it?

I'm not sure why, but a URL in a doc comment that you didn't change is now causing a failure, can you fix it while you're here? Looks like you just need to wrap the URL in `<>` to make rustdoc happy.

Happy to merge this once CI is green.

minidump-common/src/format.rs
flub: I can't actually see that target when building the docs, I haven't used this API yet so not sure what's is supposed to point at.

minidump-common/src/format.rs
flub: Yes, that links correctly to other crates (as long as you don't build with `--no-deps`)

flub:
> Thanks for these changes, this seems like it'll be much nicer! This doesn't cause any issues with `cargo doc` on a non-nightly compiler, does it?

On stable those links get rendered as plain text and with the square brackets around them.  So you do get the weird looking brackets around it on stable.

> I'm not sure why, but a URL in a doc comment that you didn't change is now causing a failure, can you fix it while you're here? Looks like you just need to wrap the URL in `<>` to make rustdoc happy.

Oops, fixed

luser:
Sorry I forgot to merge this after that conversation. Thanks for the patch!

PR #122, opened 2020-12-15
Use stable rustc for docs CI

flub:
The intra-doc linking now works on rust stable.  No need to use
nightly anymore.

Also update caching action which is simpler to use in more recent
versions.


luser:
Thanks!

PR #124, opened 2020-12-17
Fix Rust Edition 2018 idioms

jan-auer:
This is a follow-up to #109, applying `cargo fix --edition-idioms` and removing all `extern crate` declarations.

cc @flub

### DIFF
--- a/examples/dumpmodules.rs
+++ b/examples/dumpmodules.rs
@@ -6,9 +6,6 @@ use std::ffi::OsStr;
 use std::io::Write;
 use std::path::Path;
 
-extern crate minidump;
-extern crate minidump_common;
-
 use minidump::*;
 use minidump_common::traits::Module;
 
--- a/src/minidump.rs
+++ b/src/minidump.rs
@@ -304,7 +304,7 @@ fn read_string_utf16(
 }
 
 /// Convert `bytes` with trailing NUL characters to a string
-fn string_from_bytes_nul(bytes: &[u8]) -> Option<Cow<str>> {
+fn string_from_bytes_nul(bytes: &[u8]) -> Option<Cow<'_, str>> {
     bytes
         .split(|&b| b == 0)
         .next()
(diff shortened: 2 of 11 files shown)


src/minidump.rs
```diff
@@ -304,7 +304,7 @@ fn read_string_utf16(
 }
 
 /// Convert `bytes` with trailing NUL characters to a string
-fn string_from_bytes_nul(bytes: &[u8]) -> Option<Cow<str>> {
+fn string_from_bytes_nul(bytes: &[u8]) -> Option<Cow<'_, str>> {
```
luser, line 307: Is this really a thing that `cargo fix` does? I don't see why lifetime elision isn't a perfectly sensible thing here.

examples/dumpmodules.rs
```diff
@@ -6,8 +6,8 @@ use std::ffi::OsStr;
 use std::io::Write;
 use std::path::Path;
 
-extern crate minidump;
-extern crate minidump_common;
+
```
luser, line 9: I'm guessing you did a find-and-replace here but didn't remove the newlines? :sweat_smile:

luser (approved):
Please remove those blank lines before merging. I don't love the weird lifetime annotation changes, but if that's what `cargo fix` does I don't have a really compelling argument against. Is there a justification somewhere for why it does that? If not, maybe we can file a bug to make it not do that?

src/minidump.rs
jan-auer: Correct. There's some information on the reasoning behind that on [the anonymous lifetime](https://doc.rust-lang.org/edition-guide/rust-2018/ownership-and-lifetimes/the-anonymous-lifetime.html#more-details) docs. Essentially, for references it is clear that there is an elided lifetime, but it is not for types. Therefore, one should annotate the anonymous lifetime always, to show that the type is not `'static`.

jan-auer:
I should look at my own diffs before submitting. Luckily, even the linter complained about those empty lines.

examples/dumpmodules.rs
jan-auer: Guessed correctly. Shame on me, really.

src/minidump.rs
luser: Okay, that's very reasonable. Thanks for digging up the link!

PR #130, opened 2021-01-29
Fix a panic in location_slice

jan-auer:
If `loc.data_size` is the equivalent of a negative number, then in production
builds the end index will wrap around and be smaller than the start index.
The `slice::get` method conveniently checks for all that already, so we can use
that instead.

Note that I left the cast for `end` outside of the addition, since it is
convenient to see the wraparound happen.


flub (approved):

luser:
This probably would have been a good place to use [`checked_add`](https://doc.rust-lang.org/nightly/std/primitive.u32.html#method.checked_add), I guess?

jan-auer:
You're right, especially since the current code can still panic when compiled with `debug_assertions`.

PR #137, opened 2021-03-30
Parse the MISC_INFO_5 format in minidumps

### LINKED ISSUE #16: Support MINIDUMP_MISC_INFO_5
This should be straightforward now given the `multi_structs` macro I wrote:
https://github.com/luser/rust-minidump/blob/e73f08c09f6177eae07f6f4f028367c61f42de1f/minidump-common/src/format.rs#L3472

Gankra:
Fixes #16

This is a bit messy because I misunderstood the scope of #16 and started working towards actually walking through the XSAVE entries that MISC_INFO_5 describes before realizing that wasn't actually desired at the moment. So this has some extra "useless" definitions that cover what the meaning of the enabled_features bitfield is, but I figured I'd leave it there in case it's ever desirable to properly process that data.

### DIFF
--- a/minidump-common/src/format.rs
+++ b/minidump-common/src/format.rs
@@ -1351,7 +1351,7 @@ pub enum PlatformId {
 /// This struct matches the [Microsoft struct][msdn] of the same name.
 ///
 /// [msdn]: https://msdn.microsoft.com/en-us/library/windows/desktop/ms724950(v=vs.85).aspx
-#[derive(Debug, Clone, Pread, SizeWith)]
+#[derive(Debug, Clone, Default, Pread, SizeWith, PartialEq, Eq)]
 pub struct SYSTEMTIME {
     pub year: u16,
     pub month: u16,
@@ -1377,6 +1377,20 @@ pub struct TIME_ZONE_INFORMATION {
     pub daylight_name: [u16; 32],
     pub daylight_date: SYSTEMTIME,
     pub daylight_bias: i32,
+}
+
+impl Default for TIME_ZONE_INFORMATION {
+    fn default() -> Self {
+        Self {
+            bias: 0,
+            standard_name: [0; 32],
+            standard_date: SYSTEMTIME::default(),
+            standard_bias: 0,
+            daylight_name: [0; 32],
+            daylight_date: SYSTEMTIME::default(),
+            daylight_bias: 0,
+        }
+    }
 }
 
 /*
@@ -1424,7 +1438,7 @@ multi_structs! {
     /// This struct matches the [Microsoft struct][msdn] of the same name.
     ///
     /// [msdn]: https://docs.microsoft.com/en-us/windows/desktop/api/minidumpapiset/ns-minidumpapiset-_minidump_misc_info_2
-    pub struct MINIDUMP_MISC_INFO2 {
+    pub struct MINIDUMP_MISC_INFO_2 {
         pub processor_max_mhz: u32,
         pub processor_current_mhz: u32,
         pub processor_mhz_limit: u32,
@@ -1435,7 +1449,7 @@ multi_structs! {
     /// Miscellaneous process and system information
     ///
     /// This struct matches the struct of the same name from minidumpapiset.h.
-    pub struct MINIDUMP_MISC_INFO3 {
+    pub struct MINIDUMP_MISC_INFO_3 {
         pub process_integrity_level: u32,
         pub process_execute_flags: u32,
         pub protected_process: u32,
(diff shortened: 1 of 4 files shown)


Gankra:
Oh also my naming scheme is all over the place because I don't quite get where SCREAMING_CASE and RustCase should be used in this file. Presumably something relating to the official headers and breakpad but MISC_INFO_3+ aren't in the headers so...?

Gankra:
Oh also I still need to test this -- need to dig up a minidump that has this data.

minidump-common/src/format.rs
```diff
@@ -1389,33 +1389,74 @@ multi_structs! {
...
+        const XSTATE_AVX512_ZMM_H           = 1 << (XStateFeatureIndex::AVX512_ZMM_H as usize);
+        const XSTATE_AVX512_ZMM             = 1 << (XStateFeatureIndex::ACK512_ZMM as usize);
+        const XSTATE_IPT                    = 1 << (XStateFeatureIndex::XSTATE_IPT as usize);
+        const XSTATE_LWP                    = 1 << (XStateFeatureIndex::XSTATE_LWP as usize);
+    }
+}
+
+#[derive(Clone, Copy, Pread, SizeWith)]
+pub struct XSTATE_FEATURE {
+    pub offset: u32,
```
jan-auer, line 1449: It would be great to assume some of Crashpad's doc comments:

- https://crashpad.chromium.org/doxygen/structXSTATE__FEATURE.html
- https://crashpad.chromium.org/doxygen/structXSTATE__CONFIG__FEATURE__MSC__INFO.html

minidump-common/src/format.rs
```diff
@@ -1389,33 +1389,74 @@ multi_structs! {
...
+    ///
+    /// This struct matches the struct of the same name from minidumpapiset.h.
+    pub struct MINIDUMP_MISC_INFO5 {
+        pub xstate_data: XSTATE_CONFIG_FEATURE_MSC_INFO,
+        pub process_cookie: u32,
+    }
+    // TODO?: read xstate_data.enabled_features and process the extra XSAVE
+    // sections that should follow MISC_INFO_5 entries as documented in
+    // Volume 1, Chapter 13 of the "Intel 64 and IA-32 Architectures
+    // Software Developer’s Manual".
```
jan-auer, line 1404: This is a great idea. 

If I understand @luser's intentions correctly, he prefers to keep these types close to the binary format. Considering this, you could provide helper methods that look into `enabled_features` and then return an `Option<XStateFeature>` by `XStateFeatureIndex`.

jan-auer (commented):
Thanks for this patch! 

I agree with your naming concerns, and it seems like you've found good middle grounds that we can go with.

If you rebase or merge `master`, all unrelated clippy and rustc errors should be resolved.

minidump-common/src/format.rs
gabrielesvelto: We'd like to implement that in a follow-up patch to keep complexity limited and also because we still have to gather a bunch of minidumps to test the feature (I've gathered minidumps  with 8 different combinations of XSTATE flags but they don't cover all of them).

Gankra: let's leave this for future work, since it potentially involves some restructuring of how we read this type out that I need to think through more.

minidump-common/src/format.rs
Gankra: I made my own versions since these refer a lot to their own APIs

Gankra:
I've also added an extra underscore to the MISC_INFO_X names to both match the windows headers and luser's original implementation of this.

minidump-common/src/format.rs
jan-auer: Could you outline your plans? It looks like all you need is one additional method that does roughly this:

```rust
impl XSTATE_CONFIG_FEATURE_MSC_INFO {
    pub fn feature(&self, index: XStateFeatureIndex) -> Option<XSTATE_FEATURE> {
        let index = index as usize;

        XStateFeatures::from_bits(1 << index)
            .filter(|f| self.enabled_features.contains(*f))
            .and_then(|_| self.features.get(index))
            .copied()
    }
}
```

minidump-common/src/format.rs
```diff
@@ -1386,37 +1386,91 @@ multi_structs! {
...
+    pub context_size: u32,
+    /// The bit `enabled_features[i]` indicates that `features[i]` contains valid data.
+    pub enabled_features: XStateFeatures,
+    /// The offset and size of each XSAVE entry inside the XSAVE context.
+    pub features: [XSTATE_FEATURE; 64],
+}
+
+/// Several known entries in `XSTATE_CONFIG_FEATURE_MSC_INFO.features`.
+#[repr(usize)]
+pub enum XStateFeatureIndex {
```
jan-auer, line 1428: Let's also add the usual trait implementations here.

jan-auer (commented):
Thanks for the updates! This looks great and is G2G once tests have been added. See my one comment below.

luser:
I had a WIP patch for this from a while back, I can't remember why I never finished it though: https://github.com/luser/rust-minidump/commit/6708884864c0526aa1909aa92f729738bafc4509

The naming conventions are mostly "match the official Microsoft struct names when possible", but there are a bunch of structs that don't originate from Microsoft headers, so it's more convention than anything.

minidump-common/src/format.rs
```diff
@@ -1386,37 +1386,91 @@ multi_structs! {
...
+#[derive(Clone, Copy, Pread, SizeWith)]
+pub struct XSTATE_FEATURE {
+    /// This entry's offset from the start of the context (in bytes).
+    pub offset: u32,
+    /// This entry's size (in bytes).
+    pub size: u32,
+}
+
+// For whatever reason Pread array derives use 0u8.into() instead of Default to
+// create an initial array to write into. Weird.
```
luser, line 1468: This is probably just "Default not implemented for arrays of arbitrary length".

Gankra:
I accidentally got my branches mixed up so the printing patches in my other PR are now folded into this one. I've updated the commit message to reflect the current state.

The new printing output now looks like this:

```
MINIDUMP_MISC_INFO
  size_of_info                 = 1364
  flags1                       = 3f7
  process_id                   = 12144
  process_create_time          = 0x605a4f4d 2021-03-23 20:27:57
  process_user_time            = 350
  process_kernel_time          = 457
  processor_max_mhz            = 3701
  processor_current_mhz        = 3701
  processor_mhz_limit          = 3701
  processor_max_idle_state     = 2
  processor_current_idle_state = 2
  process_integrity_level      = 8192
  process_execute_flags        = 3a
  protected_process            = 0
  time_zone_id                 = 2
  time_zone                    =
    bias          = 480
    standard_name = Pacific Standard Time
    standard_date = 0000-11-01 02:00:00:000000000
    standard_bias = 0
    daylight_name = Pacific Daylight Time
    daylight_date = 0000-03-02 02:00:00:000000000
    daylight_bias = -60
  build_string                 = 19041.1.amd64fre.vb_release.191206-1406
  dbg_bld_str                  = dbgcore.i386,10.0.19041.546
  xstate_data                  =
    feature  0 - LEGACY_FLOATING_POINT :  offset    0, size  160
    feature  1 - LEGACY_SSE            :  offset  160, size  256
    feature  2 - GSSE_AND_AVX          :  offset  576, size  256
  process_cookie               = 11984912
```

Gankra:
Oh right: I removed the bitflags impl from the enabled_features field because it didn't really do anything and bitflags isn't designed for situations where unspecified bits can be set. So now it's a bare integer that's interpreted by the iterator impl I added.

Gankra:
Apparently you can mess up a rebase so hard github just closes the issue because it thinks your PR was merged lol

Gankra:
Is there anything else that needs to be done to push this forward?

minidump-common/src/format.rs
```diff
@@ -1446,37 +1460,133 @@ multi_structs! {
...
+
+/// An iterator of all the enabled features in an XSTATE_CONFIG_FEATURE_MSC_INFO.
+#[derive(Debug)]
+pub struct XstateFeatureIter<'a> {
+    info: &'a XSTATE_CONFIG_FEATURE_MSC_INFO,
+    idx: usize,
+}
+
+impl<'a> Iterator for XstateFeatureIter<'a> {
+    type Item = (usize, &'a XSTATE_FEATURE);
```
jan-auer, line 1526: Two questions here:

1. Since `XSTATE_FEATURE` is `Copy`, can you just return the owned instance?
2. The index would always be one of `XStateFeatureIndex`, would it make sense to use that type instead?

minidump-common/src/format.rs
jan-auer: This still applies, could you add at least `Clone, Copy`, and depending on what you think makes sense also `Ord, Eq, Hash`?

minidump-common/src/format.rs
```diff
@@ -1446,37 +1460,133 @@ multi_structs! {
...
+impl XSTATE_CONFIG_FEATURE_MSC_INFO {
+    /// Gets an iterator of all the enabled features.
+    pub fn iter(&self) -> XstateFeatureIter {
+        XstateFeatureIter { info: self, idx: 0 }
+    }
+}
+
+/// An iterator of all the enabled features in an XSTATE_CONFIG_FEATURE_MSC_INFO.
+#[derive(Debug)]
+pub struct XstateFeatureIter<'a> {
```
jan-auer, line 1520: This structure is spelled `Xstate` vs the index enum that is spelled `XState`. Even though it's just a minor difference, it would be great consolidating this.

jan-auer (approved):
Looks good to me, especially the updates to minidump_dump and SynthMinidump. Noticed three minor points in final review, other than that this is ready to merge.

Thanks!

minidump-common/src/format.rs
Gankra: It can't be an XStateFeatureIndex because there can be unknown indices. We could fill up the enum with Unknown1, Unknown2... but I don't think that's particularly helpful. We *could* yield `Option<XStateFeatureIndex>` but that's a bit weird, and may be limiting if the user of our library has extra info about index mappings.

Gankra:
Should be good to go.

jan-auer:
Thank you!

PR #138, opened 2021-04-02
Fix acronym and panic warnings

jan-auer:
Rust and Clippy 1.51.0 have new default warnings:

 - `clippy::from_over_into` prefers a `From` impl, which comes with a blanked `Into` impl for symmetry.
 - `clippy::upper_case_acronyms` enforces [RFC 430 naming guidelines](https://rust-lang.github.io/api-guidelines/naming.html#casing-conforms-to-rfc-430-c-case).
 - `rustc_lint::non_fmt_panic` warns against non-format strings as first
    argument to `panic!` and related macros. In Rust 2021, this will become a hard
    error.

This includes breaking changes in the signature of some public types.

### DIFF
--- a/breakpad-symbols/src/lib.rs
+++ b/breakpad-symbols/src/lib.rs
@@ -491,10 +491,8 @@ impl Symbolizer {
             debug!("locate_symbols for {}: {}", module.code_file(), res);
             self.symbols.borrow_mut().insert(k.clone(), res);
         }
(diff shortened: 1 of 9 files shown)


flub (approved):
If we're fine with breaking the public API then lgtm

jan-auer:
The public API has changed already since the last release (`0.1.0`), so as long as there is no objection to following RFC 430, we should be good to break.

PR #139, opened 2021-04-02
Fix reading annotation object strings

jan-auer:
While most of Crashpad's UTF8 strings are terminated with a NUL byte,
the string variant of annotation objects is not. The Breakpad parser for
CrashpadInfo ignores this, never validates the NUL byte, and doesn't
even implement parsing of annotation objects.

To make intentions more clear, this patch splits UTF8 string parsing
into two methods: one that validates the terminating NUL byte, and one
that does not.

### DIFF
--- a/src/minidump.rs
+++ b/src/minidump.rs
@@ -346,18 +346,27 @@ fn read_string_utf16(
     }
 }
 
-fn read_string_utf8<'a>(
+#[inline]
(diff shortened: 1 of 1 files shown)


nornagon (approved):
Thanks :)

flub (approved):

PR #140, opened 2021-04-02
Print CrashpadInfo in minidump_dump

jan-auer:
Implements support for `MinidumpCrashpadInfo` in `minidump_dump`, mostly
equivalent to Breakpad's reference implementation. Breakpad does not
support annotation objects, which we print in a similar fashion.

This is the first time we have to interpret and print the `GUID` type.
Note that the interpretation of these GUIDs is endianness dependent,
which means that byte order of the first three fields is inverted for
little-endian _Minidumps_ independent of the host endianness.

### DIFF
--- a/minidump-common/src/format.rs
+++ b/minidump-common/src/format.rs
@@ -11,6 +11,8 @@
 #![allow(non_upper_case_globals)]
 #![allow(clippy::upper_case_acronyms)]
 
+use std::fmt;
+
 use bitflags::bitflags;
 use enum_primitive_derive::Primitive;
 use scroll::{Endian, Pread, SizeWith};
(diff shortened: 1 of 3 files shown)


src/bin/minidump_dump.rs
```diff
@@ -57,6 +57,11 @@ fn print_minidump_dump(path: &Path) {
                 breakpad_info.print(stdout).unwrap();
             }
             // TODO: MemoryInfoList
+            if let Ok(crashpad_info) = dump.get_stream::<MinidumpCrashpadInfo>() {
+                crashpad_info.print(stdout).unwrap();
+            } else {
+                write!(stdout, "MinidumpCrashpadInfo cannot print invalid data").unwrap();
```
nornagon, line 63: This should probably print nothing if there's no `MinidumpCrashpadInfo` stream in the dump? But for that we'd need to distinguish between "no stream present" and "stream present, but parsing failed"

jan-auer: Good catch. Would you believe me that I was thinking exactly the same while AFK? 😄

minidump-common/src/format.rs
```diff
@@ -411,6 +413,26 @@ pub struct GUID {
     pub data4: [u8; 8],
 }
 
+impl fmt::Display for GUID {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
```
flub, line 417: `guid_to_string` already exists and probably there should only be one impl as this endianness is really hard to get right.

afaik this will print the "wrong" guids if the target is LE.  This is the same as breakpad does so makes sense for now.

flub (approved):

PR #141, opened 2021-04-05
Print out all of the MISC_INFO fields

Gankra:
This includes some additional extensions to the macros that manipulate MISC_INFO fields to handle versions and simple printing cases. `misc_accessors!` has been changed to always yield an Option, and always yield references. 

Yielding references avoids very large copies for the more enormous fields, at the cost of making it slightly more awkward to use the smaller POD values. Adding special casing for reference-vs-value seemed like a needless mess, but it can be done if desired.

Making everything an Option is the result of adding support for fields which e.g. don't exist in MISC_INFO_2 but do exist in MISC_INFO_3. This is an "unnecessary" regression for precisely the first two fields of MISC_INFO which are always guaranteed to exist. Adding special casing for those two seemed like a needless mess, but it can be done if desired.

This builds on #137 (commit included until it lands).

I am a bit uncertain about how I handle SYSTEMTIME instances here. They don't seem to be "normal" SYSTEMTIME's (e.g. in the one minidump I'm testing with the year is 0). Also I just drop the day-of-the-week entry since that seems uninteresting.

### DIFF
--- a/minidump-common/src/format.rs
+++ b/minidump-common/src/format.rs
@@ -1425,6 +1425,7 @@ pub struct XSTATE_CONFIG_FEATURE_MSC_INFO {
 
 /// Several known entries in `XSTATE_CONFIG_FEATURE_MSC_INFO.features`.
 #[repr(usize)]
+#[derive(Debug)]
 pub enum XStateFeatureIndex {
     LEGACY_FLOATING_POINT = 0,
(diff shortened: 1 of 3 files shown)


Gankra:
Did some cleanup on this to fixup most of my annoyances with my first pass. Initial comment updated to reflect the current state.

PR #143, opened 2021-04-19
gracefully fail on invalid dates

nornagon:
This panics on some minidumps, so instead default to the empty string if the
date is invalid.

### DIFF
--- a/src/minidump.rs
+++ b/src/minidump.rs
@@ -326,15 +326,18 @@ fn format_time_t(t: u32) -> String {
 
 fn format_system_time(time: &md::SYSTEMTIME) -> String {
     // Note this drops the day_of_week field on the ground -- is that fine?
-    let date = NaiveDate::from_ymd(time.year as i32, time.month as u32, time.day as u32);
-    let time = NaiveTime::from_hms_milli(
-        time.hour as u32,
-        time.minute as u32,
-        time.second as u32,
-        time.milliseconds as u32,
-    );
-    let datetime = NaiveDateTime::new(date, time);
-    datetime.format("%Y-%m-%d %H:%M:%S:%f").to_string()
+    if let Some(date) = NaiveDate::from_ymd_opt(time.year as i32, time.month as u32, time.day as u32) {
+        let time = NaiveTime::from_hms_milli(
+            time.hour as u32,
+            time.minute as u32,
+            time.second as u32,
+            time.milliseconds as u32,
+        );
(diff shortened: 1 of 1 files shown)


src/minidump.rs
```diff
@@ -326,15 +326,20 @@ fn format_time_t(t: u32) -> String {
...
-        time.minute as u32,
-        time.second as u32,
-        time.milliseconds as u32,
-    );
-    let datetime = NaiveDateTime::new(date, time);
-    datetime.format("%Y-%m-%d %H:%M:%S:%f").to_string()
+    if let Some(date) =
+        NaiveDate::from_ymd_opt(time.year as i32, time.month as u32, time.day as u32)
+    {
+        let time = NaiveTime::from_hms_milli(
```
jan-auer, line 332: Since we're already on it, we could also use `from_hms_milli_opt` or `date.and_hms_milli_opt` here.

src/minidump.rs
```diff
@@ -326,15 +326,20 @@ fn format_time_t(t: u32) -> String {
...
+            time.hour as u32,
+            time.minute as u32,
+            time.second as u32,
+            time.milliseconds as u32,
+        );
+        let datetime = NaiveDateTime::new(date, time);
+        datetime.format("%Y-%m-%d %H:%M:%S:%f").to_string()
+    } else {
+        String::new()
```
jan-auer, line 341:
```suggestion
        "<invalid date>".to_owned()
```

jan-auer (commented):
Thank you. Out of curiosity, which values is this failing for in practice?

If we want to support printing these, we could also remove the use of `chrono` here and just use plain string formatting to print out the components.

cc @Gankra

Gankra:
Oh thanks, totally missed the _opt versions

luser:
> If we want to support printing these, we could also remove the use of `chrono` here and just use plain string formatting to print out the components.

I'm not sure this code (which I wrote) is getting much value out of using `chrono`, but we are using it elsewhere for datetime handling so removing the dependency doesn't seem likely.

jan-auer:
In that case, I'd suggest writing the date directly here as we're not getting value out of the validation. Can still keep `chrono` in other places where it actually matters.

nornagon:
> Thank you. Out of curiosity, which values is this failing for in practice?

year=0, month=0, day=0

Gankra:
ping @luser we should just land this, eh?

PR #144, opened 2021-04-19
Make MinidumpStream::read take &Minidump so that streams can depend on other streams

Gankra:
This is required for several things, like the CPU context of a MinidumpThread requiring MiscInfo to know the format of the XSAVE data; or a potential stackwalker for a MinidumpThread requiring ModuleList access to lookup data on instruction addresses.

As of this PR the functionality is unused, but we need to establish *how* stream dependencies are resolved sometime, and I figured it would be good to pull it out into its own PR for discussion. 

This PR implements the simplest method. It's not a problem that we pass an immutable reference to the Minidump when reading a stream because that operation already doesn't mutate the parent minidump.

Streams depending on eachother can potentially lead to infinite recursion, but this should be trivially caught in basic testing, and I'm not aware of any cyclic relationships like that.

### DIFF
--- a/src/minidump.rs
+++ b/src/minidump.rs
@@ -112,7 +112,12 @@ pub trait MinidumpStream<'a>: Sized {
     /// `bytes` is the contents of this specific stream.
     /// `all` refers to the full contents of the minidump, for reading auxilliary data
     /// referred to with `MINIDUMP_LOCATION_DESCRIPTOR`s.
-    fn read(bytes: &'a [u8], all: &'a [u8], endian: scroll::Endian) -> Result<Self, Error>;
+    fn read<T: Deref<Target = [u8]> + 'a>(
+        minidump: &'a Minidump<'a, T>,
+        bytes: &'a [u8],
+        all: &'a [u8],
+        endian: scroll::Endian,
+    ) -> Result<Self, Error>;
 }
 
 /// CodeView data describes how to locate debug symbols
@@ -851,7 +856,8 @@ impl Default for MinidumpModuleList {
 impl<'a> MinidumpStream<'a> for MinidumpModuleList {
     const STREAM_TYPE: MINIDUMP_STREAM_TYPE = MINIDUMP_STREAM_TYPE::ModuleListStream;
 
-    fn read(
+    fn read<T: Deref<Target = [u8]> + 'a>(
+        _minidump: &'a Minidump<'a, T>,
         bytes: &'a [u8],
         all: &'a [u8],
         endian: scroll::Endian,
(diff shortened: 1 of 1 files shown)


Gankra:
[Example of why modules are needed for stackwalking](https://searchfox.org/mozilla-central/rev/46a67b8656ac12b5c180e47bc4055f713d73983b/toolkit/crashreporter/google-breakpad/src/processor/stackwalker.cc#306-315)

luser:
> or a potential stackwalker for a MinidumpThread requiring ModuleList access to lookup data on instruction addresses

I'm not sure I find this example compelling—the existing stackwalking code already takes a `MinidumpModuleList` argument, we're just not currently passing it down to `get_caller_frame`:
https://github.com/luser/rust-minidump/blob/d4562d55acc60bd8f102e17b46344fcca81bd715/minidump-processor/src/stackwalker/mod.rs#L72

More generally, yeah, I agree this is sort of a pain. Breakpad sidesteps this by just having every stream object hold on to a `Minidump*` because it's C++ and YOLO:
https://chromium.googlesource.com/breakpad/breakpad/+/c484031f1f199ee53567241426efffee49008f82/src/google_breakpad/processor/minidump.h#129

Would this work better if we refactored the API to be in terms of `Rc<Minidump>`, so that streams could just hold on to a reference and use it later? The `MinidumpStream` API is already wonky because when reading a stream there can be references to arbitrary offsets in the minidump, so I wound up just punting and passing the entire contents as a parameter:
https://github.com/luser/rust-minidump/blob/d4562d55acc60bd8f102e17b46344fcca81bd715/src/minidump.rs#L115

Gankra:
Oh! I completely missed the minidump-processor crate, interesting. So stackwalking shouldn't be handled by the core stream reading, but instead be a post-process step by minidump-processor? Interesting.

Using the MISC_INFO_5 info is still a usecase though, unless you *also* want that to be a post-process.

luser:
I don't have a great answer here, and the API design of what's currently in this repo hasn't had enough real-world usage to call it well-designed, so I'm open to  changes like this if they make it easier to do the work you actually need to do. Are you working on a stackwalker implementation currently, or is that work you plan to do? Either way, if this change (or something like it) makes that work easier then I'm in favor of letting you do whatever works best for you. I'm happy to explain the design choices here as well as I can (do note that a lot of this crate was written as Babby's First Rust Project) if that helps you.

Gankra:
Yea we're currently looking at beefing this up until it can replace firefox's use of breakpad/minidump-stackwalk, and part of that is fleshing out the stackwalker to do scanning/cfi stuff. I had started prototyping it out as part of reading the MinidumpThreadList since that just made intuitive sense to me, but I'm fine moving that logic to minidump-processor, where I see you've already sketched out a skeleton that mimics the design of breakpad's stackwalker.

luser:
> where I see you've already sketched out a skeleton that mimics the design of breakpad's stackwalker

It will probably not surprise you to learn that much of the design of this family of crates was heavily influenced by my experience with the Breakpad codebase. :slightly_smiling_face: I think Breakpad is generally a very solid project, it's just hamstrung by being implemented in C++.

Gankra:
Gonna close this for now. Now that I understand the code better, it seems it's only potentially needed by MISC_INFO_5 handling, which is a bridge we'll cross when we need to. (not high on my priorities)

PR #160, opened 2021-05-11
First stab at MINIDUMP_UNUSED_MODULE_LIST

flub:
This doesn't work yet.


flub:
superseded by #174

Gankra:
Oh geez, I forgot you started this 😰

flub:
> Oh geez, I forgot you started this 😰

so did i 😄  
I hadn't spotted that the stream itself was actually a different format and that the issue was i had in this PR.  so not much is lost :)

PR #165, opened 2021-05-12
Handle null stack RVAs

### LINKED ISSUE #164: Handle null stack RVAs
Sometimes windows makes the RVAs for the stack null, indicating we need to look the memory up in the memory list: https://searchfox.org/mozilla-central/rev/27722db2f164add7047d7db03169966cb806e927/toolkit/crashreporter/google-breakpad/src/processor/minidump_processor.cc#268-270

(Currently we erroneously accept RVA==0, mapping garbage in for the stack)

Gankra:
fixes #164

This one is a bit hacky...

### DIFF
--- a/src/minidump.rs
+++ b/src/minidump.rs
@@ -889,6 +889,13 @@ impl<'a> MinidumpMemory<'a> {
         desc: &md::MINIDUMP_MEMORY_DESCRIPTOR,
         data: &'a [u8],
     ) -> Result<MinidumpMemory<'a>, Error> {
+        if desc.memory.rva == 0 || desc.memory.data_size == 0 {
+            // Windows will sometimes emit null stack RVAs, indicating that
+            // we need to lookup the address in a memory region. It's ok to
+            // emit an error for that here, the thread processing code will
+            // catch it.
+            return Err(Error::MemoryReadFailure);
+        }
         let bytes = location_slice(data, &desc.memory).or(Err(Error::StreamReadFailure))?;
         Ok(MinidumpMemory {
             desc: *desc,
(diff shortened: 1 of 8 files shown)


src/minidump.rs
```diff
@@ -889,6 +889,13 @@ impl<'a> MinidumpMemory<'a> {
         desc: &md::MINIDUMP_MEMORY_DESCRIPTOR,
         data: &'a [u8],
     ) -> Result<MinidumpMemory<'a>, Error> {
+        if desc.memory.rva == 0 || desc.memory.data_size == 0 {
```
luser, line 892: The `location_slice` function called below should probably be fixed to handle this. It's my "turn an RVA into a byte slice" utility function, so fixing it there would fix this for most places that want to use an RVA like this.

src/minidump.rs
```diff
@@ -1058,13 +1065,6 @@ impl<'a> MinidumpStream<'a> for MinidumpMemoryList<'a> {
         // read memory contents for each region
         let mut regions = Vec::with_capacity(descriptors.len());
         for raw in descriptors.into_iter() {
-            if raw.memory.data_size == 0
-                || raw.memory.data_size as u64 > (u64::max_value() - raw.start_of_memory_range)
-            {
-                // Bad size.
-                // TODO: just drop this memory, keep the rest?
```
luser, line 1065: It's probably worth addressing this at some point. I spent a lot of time fixing things in Breakpad so that we could get *some* useful data out of minidumps where parts of them were broken. Being able to read the memory list even if some of the entries are corrupt would be useful!

src/minidump.rs
Gankra: I originally did that but it made `test_crashpad_info` explode. There's a few places where `rva=0` appears to be legit, but never for MinidumpMemory specifically.

src/minidump.rs
Gankra: I agree things should be more robust, but, I'd rather punt this for now as I have way bigger fish to fry first.

Gankra: Actually I changed my mind, this is trivial to implement and seems uncontroversial to me.

PR #178, opened 2021-05-19
Implement CrashReason for Windows/Linux/MacOs

### LINKED ISSUE #22: Actually implement CrashReason
We've got a `CrashReason` type, but `CrashReason::from_exception` is just a stub currently:
https://github.com/luser/rust-minidump/blob/e73f08c09f6177eae07f6f4f028367c61f42de1f/src/minidump.rs#L1317

Breakpad has a `GetCrashReason` function that takes the exception information + the OS from the system info stream and returns specific exception information (albeit as a string):
https://chromium.googlesource.com/breakpad/breakpad/+/54fa71efbe50fb2b58096d871575b59e12edba6d/src/processor/minidump_processor.cc#682

We should copy this logic to provide more useful crash info.

Gankra:
Fixes #22

Note that this punts on properly outputting NTSTATUS/FAST_FAIL values for now, because @gabrielesvelto was planning a bunch of work in that area anyway.

This is a big gross mess, made much worse by massive indentation and very long type names. I don't *love* the hack of using Debug output since it's Not Guaranteed, but the alternative is just such an enormous amount of messy hard-to-maintain copy-pasting.

### DIFF
--- a/src/minidump.rs
+++ b/src/minidump.rs
@@ -255,7 +255,43 @@ pub struct MinidumpBreakpadInfo {
 /// The reason for a process crash.
 #[derive(Copy, Clone, Debug, PartialEq)]
 pub enum CrashReason {
-    Unknown,
+    /// A Mac/iOS error code with no other interesting details.
+    MacGeneral(md::ExceptionCodeMac, u32),
+    MacBadAccessKern(md::ExceptionCodeMacBadAccessKernType),
+    MacBadAccessArm(md::ExceptionCodeMacBadAccessArmType),
+    MacBadAccessPpc(md::ExceptionCodeMacBadAccessPpcType),
+    MacBadAccessX86(md::ExceptionCodeMacBadAccessX86Type),
+    MacBadInstructionArm(md::ExceptionCodeMacBadInstructionArmType),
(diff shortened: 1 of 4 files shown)


src/minidump.rs
```diff
@@ -2007,13 +2284,76 @@ impl fmt::Display for CrashReason {
...
+        }
+
+        fn write_fast_fail(_f: &mut fmt::Formatter<'_>, _fast_fail: u64) -> fmt::Result {
+            // TODO: implement this
+            Ok(())
+        }
+
+        // OK this is kinda a gross hack but I *really* don't want
+        // to write out all these strings again, so let's just lean on Debug
+        // repeating the name of the enum variant!
```
luser, line 2301: This seems fine. If you want to pull in a proc macro crate that offers `#[derive(Display)]` for enums at some point, feel free.

luser (approved):
This seems like it must have been awful to write, thanks for tackling it! Having something working here is better than nothing. We can always make improvements down the road.

Gankra:
> This seems like it must have been awful to write, thanks for tackling it!

God it really was, even though it was just copying the logic from breakpad... so many cases to write out.

PR #179, opened 2021-05-20
Gardening

Gankra:
Just some random stuff I felt like doing.


Gankra:
This updates everything but `nom` which will be a nightmare and also isn't worth doing given it's all code we want to toss in favour of symbolic.

luser:
> This updates everything but `nom` which will be a nightmare and also isn't worth doing given it's all code we want to toss in favour of symbolic.

I have like 3 stale branches where I went down this road but got frustrated and gave up so you're making the right choice here. We can probably tell dependabot to never notify us about nom again?

PR #181, opened 2021-05-20
Move minidump-processor's binary to a separate minidump-stackwalk crate

### LINKED ISSUE #180: minidump-processor's binary should have a Cargo.lock
Not sure why it doesn't / how to make it use one. Builds need to be properly reproducible.

Gankra:
This way it can be properly published on crates.io and pulled down with `cargo install minidump-stackwalk`.

This will require publishing a new version of all the crates, whose versions I've already bumped. Once all the others have been published, I intend to publish the minidump-stackwalk crate to reserve the name and verify all the dependencies work.

I believe this will also make it easier to pin a Cargo.lock to the binary (which I'll do in a followup PR once everything is published).

Oh and because there's 3 things called "minidump-stackwalk" I'm going to start referring to this as "rust-minidump-stackwalk".

Fixes #180


Gankra:
cc @luser 

* does this seem reasonable
* can you add me as an owner to these crates for publishing?

Gankra:
I think *maybe* what wants to happen here is to push rust-minidump down to the same level as all the other crates, so that it's not a "library" workspace -- need to read more.

Gankra:
New description:

* Move minidump down into a subdirectory so the workspace is virtual
* Move minidump-processor's binary to its own top-level crate
* Add the Cargo.lock to the repo (remove it from .gitignore)

This more accurately reflects the nature of the repo. The `minidump` crate
is in no way the "dominant" one. As a nice side-effect, this means running
`cargo test` or any other command will now properly apply to all the crates,
instead of just `minidump` or the one you're currently cd'd into.

All crates are also bumped to 0.3.0 to be published, so that we can claim the
minidump-stackwalk name.

luser:
> I think _maybe_ what wants to happen here is to push rust-minidump down to the same level as all the other crates, so that it's not a "library" workspace -- need to read more.

Many of the design decisions in this repository can be traced back to it being literally the first Rust project I started after taking a Rust training at a Mozilla All Hands. :slightly_smiling_face: It started out as just a single crate, then at some point I decided I wanted to split pieces out, but I never really reorganized it all.

minidump-stackwalk/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
+name = "minidump-stackwalk"
+version = "0.3.0"
+authors = ["Ted Mielczarek <ted@mielczarek.org>"]
```
luser, line 4: Do feel free to add yourself in the authors list of any of these crates if you'd like!

minidump-stackwalk/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+license = "MIT"
+readme = "README.md"
+documentation = "https://docs.rs/minidump-stackwalk"
+homepage = "https://github.com/luser/rust-minidump"
+repository = "https://github.com/luser/rust-minidump"
+edition = "2018"
+
+[badges]
+travis-ci = { repository = "luser/rust-minidump" }
```
luser, line 13: I guess this is obsolete, eh? Presumably there's a GitHub Actions equivalent?

Gankra:
`cargo install minidump-stackwalk` now works 🎊

Issue #159, opened 2021-05-10
Replace breakpad-symbols with something based on symbolic


Gankra:
See also the discussion in https://github.com/getsentry/symbolic/issues/31

So far I don't actually have any complaints with breakpad-symbols, it's done exactly what I needed. But I get the sense that in implementing CFI-based unwinding (#26) it's going to behoove us to actually do the legwork to move to the more mature/maintained Symbolic crates.

Things this entails:

* Reading text-based breakpad files from the filesystem
  * Fetching those files [from mozilla's servers](https://tecken.readthedocs.io/en/latest/download.html)
    * And caching those results
  * or if we want, some other debug info format we migrate dump_syms to generating
* Supporting queries on the symbols at a given address
  * For symbolicating the stack
  * For detecting if a potential return address is actually in a function (to make stack scanning more robust)
* Supporting lookup of the unwinding instruction at a given address
  * Not sure what the output of this query would be...
    * Could give us raw breakpad cfi bytecode text we parse/interpret?
    * Could give us pre-parsed bytecode we interpret?
    * Could let us pass in a "context" and have it output a new context? (making the unwinder a black box)
* Potentially: support digesting native pdb/dwarf-cfi/compact-unwinding tables too
  * Not needed for the #153 MVP, but would be desirable for later milestones of https://github.com/getsentry/symbolic/issues/375
  * Specifically necessary for richer client-side unwinding using native OS symbols
  * For unwinding instructions, this creates a decision point:
    * Could squish the formats into some common bytecode
    * Could expose them more natively via an enum
    * Could hide the details by making the unwinder a black box
* Anything else?

I currently don't know the symbolic APIs well enough to say how much it can just be used out-of-the-box and how much we'll need to build wrappers / new impls. In the meantime I'm going to prototype out x64 CFI unwinding using breakpad-symbols, to figure out the shape of the API I need/want here.

PR #185, opened 2021-06-01
support wasm and async

nornagon:
This is a bit of a WIP PR, and I'm a Rust newbie so sorry for any obvious
mistakes. With these changes I was able to successfully symbolicate a crash in
the browser using a "FetchSymbolSupplier" to fetch symbols from a
CORS-supporting symbol server (see code below).

There are two main changes here:

1. Elide non-wasm-compatible code using `#[cfg(not(target_arch = "wasm32"))]`.
   The only place this is necessary is for usages of the `memmap` crate.
2. Make SymbolSupplier `async`, to support using `fetch()` in the browser.

Of these, the second change is far more invasive. I'd happily change the
approach if there were a less invasive way to allow fetching symbols in the
browser--for example, there might be a way to have a "cooperative" scheduler
for fetching symbols that allowed all the rust code to be synchronous, e.g. by
returning something like `Err(MissingSymbolFile(debug_id = ..., name = ...))`
when a symbol file is missing, so that then the "driver" can go fetch that
symbol asynchronously, and once that's done, retry processing with that symbol
file being synchronously available, and possibly receiving another "symbol file
missing" error & repeating the process.

Is this a change that would potentially be considered mergeable? Either in part
or in whole?

```rust
/// An implementation of `SymbolSupplier` that loads Breakpad text-format symbols from HTTP
/// URLs over fetch().
pub struct FetchSymbolSupplier {
    /// URLs to search for symbols.
    urls: Vec<String>,
}

impl FetchSymbolSupplier {
    pub fn new(
        urls: Vec<String>,
    ) -> FetchSymbolSupplier {
        let urls = urls
            .into_iter()
            .map(|mut u| {
                if !u.ends_with('/') {
                    u.push('/');
                }
                u
            })
            .collect();
        FetchSymbolSupplier {
            urls,
        }
    }
}

async fn fetch_symbol_file(
    base_url: &str,
    rel_path: &str,
) -> Result<Vec<u8>, Error> {
    let url = base_url.to_owned() + rel_path;
    let mut opts = RequestInit::new();
    opts.method("GET");
    opts.mode(RequestMode::Cors);
    opts.cache(RequestCache::ForceCache);

    let request = Request::new_with_str_and_init(&url, &opts).unwrap();

    let window = web_sys::window().unwrap();
    let resp_value = JsFuture::from(window.fetch_with_request(&request)).await.map_err(|_| Error::IoError)?;

    assert!(resp_value.is_instance_of::<Response>());
    let resp: Response = resp_value.dyn_into().unwrap();

    let buf = JsFuture::from(resp.array_buffer().unwrap()).await.unwrap();

    let typebuf = js_sys::Uint8Array::new(&buf);
    let mut body = vec![0; typebuf.length() as usize];
    typebuf.copy_to(&mut body[..]);
    Ok(body.to_vec())
}

#[async_trait(?Send)]
impl SymbolSupplier for FetchSymbolSupplier {
    async fn locate_symbols(&self, module: &dyn Module) -> SymbolResult {
        log("here");
        // Check local paths first.
        if let Some(rel_path) = relative_symbol_path(module, "sym") {
            for ref url in self.urls.iter() {
                if let Ok(buf) =
                    fetch_symbol_file(url, &rel_path).await
                {
                    return SymbolFile::from_bytes(&buf)
                        .map(SymbolResult::Ok)
                        .unwrap_or_else(SymbolResult::LoadError);
                }
            }
        }
        SymbolResult::NotFound
    }
}
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -79,6 +79,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "async-trait"
+version = "0.1.50"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0b98e84bbb4cbcdd97da190ba0c58a1bb0de2c1fdf67d159e192ed766aeca722"
+dependencies = [
+ "proc-macro2 1.0.27",
+ "quote 1.0.9",
+ "syn 1.0.72",
+]
+
+[[package]]
 name = "atty"
 version = "0.2.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -145,6 +156,7 @@ checksum = "cf1de2fe8c75bc145a2f577add951f8134889b4795d47466a54a5c846d691693"
 name = "breakpad-symbols"
 version = "0.3.0"
 dependencies = [
+ "async-trait",
  "failure",
  "log",
  "minidump-common",
@@ -533,12 +545,28 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3dcaa9ae7725d12cdb85b3ad99a434db70b468c09ded17e012d86b5c1010f7a7"
 
 [[package]]
+name = "futures"
+version = "0.3.15"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0e7e43a803dae2fa37c1f6a8fe121e1f7bf9548b4dfc0522a42f34145dadfc27"
+dependencies = [
+ "futures-channel",
+ "futures-core",
+ "futures-executor",
+ "futures-io",
+ "futures-sink",
+ "futures-task",
+ "futures-util",
+]
+
+[[package]]
 name = "futures-channel"
 version = "0.3.15"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e682a68b29a882df0545c143dc3646daefe80ba479bcdede94d5a703de2871e2"
 dependencies = [
  "futures-core",
+ "futures-sink",
 ]
 
 [[package]]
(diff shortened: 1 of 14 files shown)


Gankra:
Oh wow! So anything involving breakpad_symbols is a bit of a dead end, because I'm hoping to do #159 some time this month, and that will result in all of breakpad_symbols being deprecated (and deleted?).

I'm less certain on the async stuff. Could you elaborate a bit more on the need for an async API in the browser? What are you trying to build?

Hey @jan-auer and @swatinem -- is symbolic like, at all compatible with being compiled to wasm and used in the browser?

nornagon:
@Gankra ah, cool, thanks for the heads up :)

The main reason we need an async API for fetching symbols in the browser is because it's not possible to do anything synchronous involving the network in the browser. The only available networking APIs are asynchronous, and webassembly is not currently "pausable" without arcane voodoo like https://github.com/WebAssembly/binaryen/blob/main/src/passes/Asyncify.cpp. I'm not aware of systems which make use of this sort of thing in Rust to "fake" the asynchronous parts, but maybe they exist?

Gankra:
Sorry I mean like, what are you trying to build? Is this for some particular product? A hobby project? (Trying to figure out how much energy I can justify putting into getting this to work -- it would help if y'all were committing to some ongoing maintenance of this feature.)

nornagon:
My goal is to build a tool for easy symbolication of [Electron](/electron/electron)'s crash dumps, allowing the user to drag a crash dump into a web page and get a symbolicated crash.

I support Electron as part of my full-time work, and this would fall under that category, so I'm happy to contribute some ongoing energy to maintaining this as long as Electron remains my full-time job. However, as mentioned, I'm far from a Rust expert, so my assistance might not always be terribly useful 😅

Gankra:
Oh wow! Ok. Uhh gonna ping my manager (Mozilla) about this since this seems like a fruitful collaboration.

mstange:
You may also be interested in the [`profiler-get-symbols` project](https://github.com/mstange/profiler-get-symbols), which only handles symbolication but not unwinding / minidump parsing. It is designed for situations where you have build artifacts from the to-be-symbolicated binary stored on your local machine. It does not communicate with a symbol server.

nornagon:
@mstange I specifically want to fetch symbols from a symbol server, so unfortunately that's not terribly useful to me.

Gankra:
Ok so currently this is a bit of a difficult thing to review/merge for a few reasons:

* Someone needs to write a test harness that will compile this to wasm and run it under wasmtime so that it doesn't break
* All the code this touches is being heavily rewritten
  * Most notably breakpad-symbols will ideally be replaced with symbolic, which is a much bigger thing with much more complex of a design and actual downstream users that will presumably make an async-ification a large project.
* I don't think I have the time to spare this month ensuring this works / doesn't break / makes sense

My main question is: what's your timeline like for this crashreporting project? 

Is this something you need to get done this week? Something you expect to take a year? Short-term I think you'll just need to fork us. Long-term we could probably help get this done in collaboration with the symbolic folks.

nornagon:
It's not urgent. Probably the best thing to do for now is to keep wasm in mind as you do the symbolic rewrite / other rewrites, so hopefully it doesn't get significantly _harder_ to do this in future. I'm OK with closing out this PR for now and trying again once the rewrites land.

Gankra:
SGTM

PR #199, opened 2021-06-07
Make a shim for toggling between breakpad and symbolic

Gankra:
To test the symbolic mode I recommend just flipping the "default-features" value in minidump-processor's Cargo.toml

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1038,7 +1038,6 @@ name = "minidump-stackwalk"
 version = "0.3.0"
 dependencies = [
  "addr2line 0.15.2",
- "breakpad-symbols",
  "clap",
  "log",
  "minidump",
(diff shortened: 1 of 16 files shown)


Gankra:
cc @swatinem

Gankra:
I guess I should properly explain this:

All the symbolizing stuff has been moved to a new module at `minidump-processor/src/symbols.rs`, which contains two versions of a private internal `symbols_shim` module whose contents are re-exported.

One version contains an implementation based on `breakpad-symbols`, the other contains an implementation based on `symbolic` (just stubbed out, all `unimplemented!()`). They are selected via minidump-processor's `breakpad-syms` and `symbolic-syms` features (`breakpad-syms` enabled by default).

Gankra:
So "all" we need to do to slot in `symbolic` is fill in the `symbolic-syms` version of the `symbols_shim` module. Of course, if symbolic would like the interface between the rust-minidump and itself to be significantly different, we can reshape it as needed. This is just what the current API is.

Swatinem:
CC @loewenheim since he will probably work on this soon.

Gankra:
I'll land this as-is, y'all can mess with it however you want.

Issue #155, opened 2021-05-04
Produce readable version of error codes


Gankra:
Stubbing out an issue for this, see https://github.com/getsentry/symbolic/issues/375#issuecomment-829520432 for details.

Gankra:
@gabrielesvelto has volunteered to do this on top of https://github.com/luser/rust-minidump/pull/178

PR #203, opened 2021-06-11
Print out readable values when encountering known Windows errors

gabrielesvelto:
The values were extracted from the ntstatus.h and winerror.h files in
the Windows 10 SDK version 10.0.19041.0.

This fixes issue #155


Gankra (approved):
This seems super straight-forward and reasonable. I really appreciate that you included how to regenerate these monsters. Mildly tempted to banish all this error code stuff to a separate file but I can do that in a followup if I get really annoyed while trying to read format.rs.

I don't think this necessarily needs tests, but I'd certainly love to get them if you have something in mind.

(you need to run cargo fmt on it though)

gabrielesvelto:
Yeah, I wanted to add some tests, just to make sure we handle all the corner-cases correctly. I'll try to do that tomorrow. You'll have noticed that I split out the types in different enums because some can only appear in the exception code but others will appear in the last error value with different names but the same values.

gabrielesvelto:
After double-checking I noticed that I should be adding an exception stream writer to the synthesizer (easy) but that would also require introducing thread contexts and that's a lot of unrelated stuff... so I'll shamelessly land this w/o tests and file an issue for them.

gabrielesvelto:
Actually I can't merge this myself, @Gankra can you do it for me please?

luser:
> This seems super straight-forward and reasonable. I really appreciate that you included how to regenerate these monsters. Mildly tempted to banish all this error code stuff to a separate file but I can do that in a followup if I get really annoyed while trying to read format.rs.

At one point the whole of `format.rs` was generated by running bindgen against the Breakpad headers. I switched to hand-editing it and matching the type names to Microsoft's headers a while back. If there are bits that are easily regenerated from headers/etc it probably does make sense to keep them in a different file just to make that process easier.

Issue #25, opened 2018-09-18
Implement an exploitability heuristic like Breakpad


luser:
Breakpad can calculate an exploitability rating based on heuristics for Windows and Linux minidumps:
https://chromium.googlesource.com/breakpad/breakpad/+/54fa71efbe50fb2b58096d871575b59e12edba6d/src/processor/exploitability_win.cc
https://chromium.googlesource.com/breakpad/breakpad/+/54fa71efbe50fb2b58096d871575b59e12edba6d/src/processor/exploitability_linux.cc

We rely on this for Firefox crashes, so it would be good to implement something that is fairly similar.

Gankra:
Ok this looks pretty straightforward to do, just mind-numbing copy-paste/translation of the relevant heuristics.

luser:
The Breakpad implementation uses an x86 disassembler for x86 crashes. This is super cool and useful but also it only works on x86.

I ported my `get-minidump-instructions` tool from C++ to Rust and wrote a little [`disasm`](https://github.com/luser/rust-disasm) crate that wraps the `capstone` crate (which itself wraps the C library) which might be a useful starting point if you wanted to go down that route:
https://github.com/luser/rust-minidump/blob/8352988600a865b03a5c178c90a91a143c57c35a/minidump-tools/src/lib.rs#L224

ixi's `yaxpeax` set of decoder crates look great, they might also be a useful tool for instruction decoding in this way. ([x86, for example](https://crates.io/crates/yaxpeax-x86))

Gankra:
I talked to @iximeow and they gave me [this useful example](https://github.com/iximeow/regmap/blob/no-gods-no-/src/regmap.rs#L423-L443). She also seemed pretty interested in integrating rust-minidump into yaxpeax's debugging tools, so this seems like a beneficial symbiotic relationship to pursue.

Gankra:
I started poking at this, but mccr8 mentioned in matrix that he doesn't think this info is very useful. Shelving the work until I can poll more people at mozilla about it.

Gankra:
Some followup from dveditz: right now the number of false "oh no super scary exploit" reports is too high. We have plenty of crash reports and not enough people to look at them all. It would be better if it was less sensitive so that the set of things it flags is a more reliable signal.

Also apparently it tends to flag MOZ_CRASH as serious and that's obvious wrong (I'm guessing the way we crash looks like an invalid instruction or something, which the exploitability analysis thinks is super scary).

[...]

PR #206, opened 2021-06-14
Minimal exploitability impl

Gankra:
Notably missing:

* anything that involves disassembling and analyzing the binary
* linux checks based on /proc/self/maps

I'm kinda bleh on how I just add the raw error code to the ProcessState -- maybe I should just pass it down directly? Or CrashReason should always make it easy to get?

### DIFF
--- a/minidump-processor/src/exploitability.rs
+++ b/minidump-processor/src/exploitability.rs
@@ -1,0 +1,271 @@
+use crate::ProcessState;
+use log::{trace, warn};
+use minidump::format::{
+    ExceptionCodeLinux, ExceptionCodeWindows, ExceptionCodeWindowsAccessType, NtStatusWindows,
+    WinErrorWindows,
+};
+use minidump::system_info::Os;
+use minidump::{CrashReason, MinidumpMemoryList};
+use num_traits::FromPrimitive;
+
+/// An estimate of how exploitable a crash is.
+pub enum Exploitability {
+    /// The crash is likely to be an exploitabile memory corruption.
+    High,
+    /// The crash might be an exploitable memory corruption.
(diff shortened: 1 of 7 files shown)


minidump-processor/src/exploitability.rs
```diff
@@ -0,0 +1,273 @@
...
+                // Generally these are all benign
+                // (IN_PAGE_ERROR *sounds* scary but it's basically the OS failing
+                // to load a swapped out page from the disk/network, which is about
+                // as scary as a power failure.)
+                exploitability_weight += TINY_BUMP;
+            }
+            EXCEPTION_ILLEGAL_INSTRUCTION
+            | EXCEPTION_FLT_INVALID_OPERATION
+            | EXCEPTION_PRIV_INSTRUCTION => {
+                // Illegal instructions are an indication that control flow
```
luser, line 134: FYI, maybe you already know this, but on x86-64 LLVM emits `ud2` for its `abort`, so this is what a Rust `panic!` winds up as. It might be worth special-casing this one situation even if you don't have a full disassembler involved? I suspect you should be able to just compare the faulting instruction against a fixed set of bytes for any given architecture to check.

Gankra: Yeah I had a vague expectation that this sort of thing is why the existing exploitability is often terrified of asserts. It's in my mental backlog to refine the analysis but people don't really pay this much attention these days, so I'm not holding myself to a very high standard for this code right now.

Gankra:
rebased and made it use the new ProcessorOptions infra, in case we want to land it as experimental.

Gankra:
Closing, see #25 for context.

PR #222, opened 2021-07-31
Print out the exception code of Linux signals sent from userspace / kernel correctly

### LINKED ISSUE #221: Print out Linux signals sent from the kernel / userspace correctly
Linux signals that have an additional exception flag (SIGBUS, SIGILL, SIGSEGV and SIGFPE) have a number of signal-specific exception codes (BUS_ADRERR for example). All these signals can also be sent either from userspace via `kill()`/`pthread_kill()` or from a random place in the kernel. In these two cases the exception code is respectively `SI_USER` (value 0) or `SI_KERNEL` (value 0x80). Here's a few examples seen in the wild:
* [cbc62af6-5283-49a2-94b0-3d3a20210731](https://crash-stats.mozilla.org/report/index/cbc62af6-5283-49a2-94b0-3d3a20210731)
* [df527006-130b-4f60-8833-1c34b0210731](https://crash-stats.mozilla.org/report/index/df527006-130b-4f60-8833-1c34b0210731)
* [d958ed59-f6f6-445e-8ca9-bd2cd0210729](https://crash-stats.mozilla.org/report/index/d958ed59-f6f6-445e-8ca9-bd2cd0210729)
* [cadfd20f-05b0-4cfa-847e-b05610210728](https://crash-stats.mozilla.org/report/index/cadfd20f-05b0-4cfa-847e-b05610210728)
* [a163244f-4a80-4990-9ed4-3b9ee0210724](https://crash-stats.mozilla.org/report/index/a163244f-4a80-4990-9ed4-3b9ee0210724)

gabrielesvelto:
This fixes #221

### DIFF
--- a/minidump-common/src/format.rs
+++ b/minidump-common/src/format.rs
@@ -6291,6 +6291,7 @@ pub enum ExceptionCodeLinux {
 // These values come from asm-generic/siginfo.h
 #[derive(Copy, Clone, PartialEq, Debug, Primitive)]
 pub enum ExceptionCodeLinuxSigillKind {
+    SI_USER = 0,
     ILL_ILLOPC = 1,
     ILL_ILLOPN = 2,
     ILL_ILLADR = 3,
@@ -6299,10 +6300,12 @@ pub enum ExceptionCodeLinuxSigillKind {
     ILL_PRVREG = 6,
     ILL_COPROC = 7,
     ILL_BADSTK = 8,
+    SI_KERNEL = 0x80,
 }
 
 #[derive(Copy, Clone, PartialEq, Debug, Primitive)]
 pub enum ExceptionCodeLinuxSigfpeKind {
+    SI_USER = 0,
     FPE_INTDIV = 1,
     FPE_INTOVF = 2,
     FPE_FLTDIV = 3,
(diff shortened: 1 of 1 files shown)


luser:
I browsed through the kernel source and couldn't quite wrap my head around how this is intended to be used in practice. At first glance I thought these would be bitflags, but they seem instead to just be catch-all values for when there's not a specific `si_code`?

gabrielesvelto:
The kernel sets `SI_USER` as the `si_code` for all signals generated from userspace via `kill()`, see [this code](https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/kernel/signal.c#n3704). `SI_KERNEL` is pretty much the same but for signals raised from the kernel, see [this code](https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/kernel/signal.c#n1150). There is a description of this in the documentation for `signal_generate()` [here](https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/include/trace/events/signal.h#n37).

Gankra (approved):

Gankra:
looks like the clippy failures are just new more aggressive lints..?

gabrielesvelto:
> looks like the clippy failures are just new more aggressive lints..?

Looks like it. Should I roll the fix into this PR or should we fix it separately?

PR #240, opened 2021-09-23
STACK WIN unwinding

Gankra:
This seems to get the implementation working well on some examples in socorro I tested.

Notable problems I found and fixed:

* I was too aggressively rejecting overlapping STACK WIN address ranges. It seems very common that the `length` of a STACK WIN record just covers the entirety of the function(?) and subsequent entries mask over it. So we have to truncate the lengths as we parse them to clean this up.
* It seems that looking up STACK WIN entries based on the callee's raw `$eip` is incorrect, and that you need to subtract 1? (Which is what adjust_instruction does, so really we should just be using that value instead of eip?)
* I was deriving grand_callee_parameter_size from the value provided in FUNC records, but it appears that STACK WIN records have the more accurate value.

Also I added in a quick and dirty more fancy version of the address_seems_valid that actually looks up the function record for the module, but it's not quite right because it doesn't have a case for "it's in the module but we couldn't load any function records". I should probably delete the code, I only wrote it while I was desperately trying to figure out what was breaking.

Surprisingly, I now don't have any examples which seem to need the accursed scanning or other hacks that breakpad is stuffed with for STACK WIN cfi!

TODO:

* [ ] test more socorro cases
* [x] write tests
* [x] don't forward callee's registers as aggressively (doesn't seem to cause problems, but breakpad doesn't say those registers are valid anymore)


Gankra:
reminder if anyone wants to test this out on some examples that I have instructions on how to use this with tecken here: https://github.com/luser/rust-minidump/tree/master/minidump-stackwalk#analyzing-firefox-minidumps

Gankra:
Hmm the extra `-1` offset I applied makes a ton of the breakpad x86 unittests break (both STACK WIN and STACK CFI). Seems there's something more subtle going on that I need to work out.

The reason I convinced myself I needed the -1 was that I was seeing a few cases where we were one-past-the-end of the record we were supposed to be using. Perhaps I'm still not computing/handling the length right?

luser:
> Surprisingly, I now don't have any examples which seem to need the accursed scanning or other hacks that breakpad is stuffed with for STACK WIN cfi!

Hah, amazing! I'm sure you could grovel through blame and bugzilla for examples, but all the actual crash reports are assuredly lost to the mists of time. Given that this is all 32-bit Windows-specific, maybe it's not worth the effort?

Gankra:
Ah right, having slept on it I realized of course that we don't adjust the `context` frame's instruction, and that's the logic I need to make the `-1` work properly in all cases!

Gankra:
PR complete, I think?

gabrielesvelto:
I'd say yes! I'm going to throw some particularly nasty minidumps at it to look at the results but the few I've already tried all looked very good already.

PR #249, opened 2021-10-20
Massively cleanup the interface for minidump-stackwalk

Gankra:
* Add lots of help messages
* Make trace logging way less noisy
* Introduce defaults for --symbols-tmp and --symbols-cache

symbols-tmp is still ignored, but we now implicitly use std::env::temp_dir via the temp_file library
As the new help message for symbols-tmp explains, a different tmp location is needed to atomically swap
files into the cache so that multiple instances can share the cache.

To improve portability/usability, symbols-cache now defaults to std::env::temp_dir()/rust-minidump-cache.
This is especially valuable on Windows where the equivalent of /tmp is in a hidden folder in the User
directory. I primarily work on Windows so this resolves a large friction point. (rust-minidump is largely
very portable, this was the last papercut I knew of!)

This deterministic default potentially introduces some weird attacks where someone writes their own
fake symbol files into the cache but we don't trust the files anyway so I think it's fine?

See rust's documentation for how to modify std::env::temp_dir (lots of options).


Gankra:
cc @willkg

Is this --symbols-tmp solution adequate for your purposes? I'm assuming socorro just use /tmp normally and doesn't actually care where the temp download files go.

breakpad-symbols/src/lib.rs
```diff
@@ -310,12 +310,23 @@ impl HttpSymbolSupplier {
...
     let base = path.parent().ok_or_else(|| {
         io::Error::new(io::ErrorKind::Other, format!("Bad cache path: {:?}", path))
     })?;
     fs::create_dir_all(&base)?;
-    let mut f = File::create(path)?;
-    f.write_all(contents)?;
+
+    let mut temp = tempfile::NamedTempFile::new()?;
+    temp.write_all(contents)?;
+    temp.persist(path)?;
```
willkg, line 329: I think this doesn't work with the way we've got Socorro processor set up. Let me do a brain dump....

We run the Socorro processor in a Docker container with a volume mount for the disk cache which has both the `symbols-cache` and `symbols-tmp` directories in it. Thus those directories aren't on the same mount as `/tmp`.

The `symbols-cache` directory is managed by the disk cache manager which is a separate process that runs in the container and uses inotify to watch for changes in the directory and maintains the files in a LRU fashion. We don't want it watching `symbols-tmp` because then partially-written files kick off inotify events and cache updates and evictions and everything gets harder to reason about. Fun times.

So what we want is something like:

1. The downloader creates a temp file with part of the filename random to write the bytes of the sym file into. The random part asserts some minor security goodness. I think it also makes the code writing the file easier to deal with since we can ignore the "what if something else is already writing this temp file but the final file isn't there?" scenario.
2. Once the temp file is completely written out and closed, the downloader tries to rename the temp file to the destination final file name. Because it's on the same volume, the downloader can do a rename which (I'm fairly sure) is atomic. This "moves" the file from `symbols-tmp` to `symbols-cache`. The disk cache manager will see the file (inotify ftw) and update the LRU and evict things as appropriate. If there's already a file at the destination, the downloader can be a little sad it did all that work, delete the temp file, and move on with life--everything is in a known good state.

Given that, I think we want to pass along the `symbols_tmp_path` and use `new_in`. Also, I think we want to use `persist_noclobber` and then I think we should be good. We need to pass along the `symbols-tmp` path and then tweak these three lines like this (barring mediocre Rust code):

```rust

let mut temp = tempfile::NamedTempFile::new_in(symbols_tmp_path)?;
temp.write_all(contents)?;
temp.persist_noclobber(path)?;
```

If any of this fails, then that's fine--we can still use the sym file we downloaded. It just doesn't get cached. Since each subsequent line depends on the success of the previous lines, we might want to check and just drop out if we hit a failure.

Having said that, the processor runs this as a subprocess, so it this fails and the stackwalk process dies, the processor will make note of that in the processed crash report and move on.

Having said that, while I keep track of exit codes in Grafana, I'm not currently looking at the `processor_notes` field because it's a mess and hard to track in aggregate. That's covered in [bug 1626811](https://bugzilla.mozilla.org/show_bug.cgi?id=1626811).

I think that's all correct. If it's not, it'd be great to be corrected because I'm basing a few things on this mental model of the world.

breakpad-symbols/src/lib.rs
```diff
@@ -326,14 +357,17 @@ fn fetch_symbol_file(
...
 ) -> Result<Vec<u8>, Error> {
     let url = base_url.join(rel_path)?;
     debug!("Trying {}", url);
     let mut res = client.get(url).send()?.error_for_status()?;
     let mut buf = vec![];
     res.read_to_end(&mut buf)?;
     let local = cache.join(rel_path);
-    match save_contents(&buf, &local) {
+
+    // Ensure the cache exists
```
willkg, line 369: I don't understand how this comment matches the following line. Shouldn't the cache directory exist already at this point?

willkg (approved):
This looks good to me. My Rust is infantile at best, though, so I can't review that short of declaring it superb by fiat.

PR #259, opened 2021-10-27
Add some symbol statistics to the JSON output.

Gankra:
Broken into two parts:

* refactoring and setup of the symbols API
* actually implementing some of the stats in the JSON

This ended up being a fairly big refactor of breakpad_symbols to be a bit more idiomatic, at least in the facade.

### DIFF
--- a/breakpad-symbols/src/lib.rs
+++ b/breakpad-symbols/src/lib.rs
@@ -49,6 +49,17 @@ pub use crate::sym_file::{CfiRules, SymbolFile};
 
 mod sym_file;
 
+/// Statistics on the symbols of a module.
+#[derive(Default, Debug)]
+pub struct SymbolStats {
+    /// If the module's symbols were downloaded, this is the url used.
+    pub symbol_url: Option<String>,
+    /// If the symbols were found and loaded into memory.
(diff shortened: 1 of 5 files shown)


gabrielesvelto:
Sorry, I thought I wrote a reply here but I must have forgotten to post it. Having information on corrupt symbols is great. I haven't seen it happen in the past but it's perfectly possible that's because we didn't really have a way to detect them.

Gankra:
I don't think I've ever noticed *completely* corrupt symbols but I've definitely found many situations where a line is messed up (and we discard it) or entries clash in irreconcilable ways (forcing us to discard one at ~random). Although I assume if the symbols were completely corrupt I just wouldn't notice because my output would agree with socorro and I would be content with the result and not dig into why some random function didn't have cfi.

luser:
You can broadly assume that any of this arcane internal state that's exposed is the result of years of painful crash investigation by former colleagues. :slightly_smiling_face:  Many of them were a result of me personally being annoyed at having folks ping me to ask why symbols were missing for specific crashes, and deciding to do the work to surface that info in Socorro for self-service analysis.

PR #263, opened 2021-10-29
Add experimental support for recovering function args + design "unstable/optional features" infra/docs

Gankra:
EDIT: this PR has scope-creeped into me designing the needed infrastructure for properly talking about experimental/optional features of minidump-processor/minidump-stackwalk. It was originally just introducing this feature, but I wasn't comfortable just landing it until I had designed this infra.

Original comment:

----

I ended up hacking this together today to help a coworker debug a weird crash report.
Spent the rest of the day polishing it up to make it usable.


luser:
Fair enough! At one point I was exploring the idea of using this crate to provide a tool that could act as a gdb target using the gdb remote protocol, so that Linux/macOS minidumps could be post-mortem debugged similarly to the way Windows minidumps can with WinDbg/Visual Studio. The appeal of that approach was that you don't have to reinvent all the wheels that debuggers have, you just have to plumb the information from the minidump through to where gdb can get at it.

bholley:
+1 to all of the reasons @Gankra gave. The gdb target thing would be cool though — even cooler if you wasmed the gdb instance and ran that in an in-browser terminal emulator inside of socorro!

gabrielesvelto:
> The gdb target thing would be cool though — even cooler if you wasmed the gdb instance and ran that in an in-browser terminal emulator inside of socorro!

That escalated quickly :sweat_smile:

More seriously, having some kind of debugger-inside-Socorro would be an incredibly pleasant feature and would make crash inspection a much easier endeavor for those who don't do it every day. The biggest problem towards implementing something like that would be non-technical though: its access would have to be restricted because you'd need to be able to poke at the actual minidumps.

Gankra:
Yeah I'm not sure function arguments could be exposed to non-security-clearance folks since that may contain private information, in theory.

luser:
> Yeah I'm not sure function arguments could be exposed to non-security-clearance folks since that may contain private information, in theory.

Right, our general principle was that aside from the address that caused e.g. a segfault, any memory data that came directly from the minidump was assumed to potentially contain sensitive information.

We already had additional info available in Socorro behind a login when I left, so that doesn't seem like it would be a huge blocker? Especially if it was all implemented client-side, access to raw dumps is auth-protected so the code could just `fetch` the dump file and do whatever it wanted.

Gankra:
Ok so anyway: r? @luser 

(at least to just the concept of including this machinery in the project)

luser:
This feels like it'd fit better in the [minidump-tools](https://github.com/luser/rust-minidump/tree/master/minidump-tools) crate? That's where I've been putting ports of the weird little one-off tools I wrote to extract specific bits of information from minidumps over the years. Can this be refactored so it's a post-processing step after minidump processing, just iterating over the frames of each thread and attempting to recover the args if possible?

Gankra:
It already is, actually. I just arbitrarily decided to hook it up in the stackwalk module because that's like, Where Stuff Happens in my head (after all this stackwalking work).

bholley:
I think there is value in having this kind of information surfaced directly in the common tool, rather than users needing to dig for it. I believe socorro also includes the output of stackwalk (with appropriate auth), so this would get surfaced there for free.

luser:
> I think there is value in having this kind of information surfaced directly in the common tool, rather than users needing to dig for it. I believe socorro also includes the output of stackwalk (with appropriate auth), so this would get surfaced there for free.

I'm not opposed to this, I pushed for getting as much data out of minidumps into a self-service form in crash-stats as possible when working on crash reporting. I don't want to speak negatively about this cool work that Gankra did, but if you're going to expose data like this from the crash reporting system, you need to make sure that you have high confidence in the data so you don't present misleading evidence to people who are investigating crashes.

For example, there's a longstanding issue with crashes on x86-64 where if the faulting address is a non-canonical pointer, the GPF from the CPU doesn't include the address, so operating systems report a bad value in the userland signal/exception (IIRC it shows up as 0 on Linux/macOS, and -1 on Windows). This isn't even Breakpad or Socorro's fault, but it has caused massive confusion over the years as a red herring to engineers.

Accurately recovering function arguments from the stack requires more debug info than is available in Breakpad-format symbol files. The DWARF and PDB files are also on the symbol server, and doing this correctly, the same way debuggers do, is plausible, but it's a lot of work. The code here is great and useful because it surfaces good-enough info with like 5% of the effort required, but I would be wary of presenting it in the crash-stats UI as objective reality.

If that's what you really want then ultimately I'm not going to stop you because you're using this for work and this is just a hobby project for me nowadays.

luser:
That is to say, despite this repo being in my personal account namespace, at present I recognize @Gankra as the defacto maintainer since she's the one doing all the work. (Don't worry, you get all the power with none of the responsibility in this case! Take *that* Uncle Ben!)

bholley:
I appreciate the words of caution. How about we surface the info, but caveat it appropriately? Can we enumerate the list of complications that might make it wrong, and simply link to that info from the dump?

Gankra:
Rebased and update this feature to be a bit more principled:

* It's "hidden" behind a --recover-function-args flag in minidump-stackwalk
* Similarly it is now enabled in minidump-processor only if you set recover_function_args in ProcessorOptions (a new thing I added in the last evil_json PR).

Gankra:
Ok so the second commit ([Add better machinery for introducing new optional/experimental features.](https://github.com/luser/rust-minidump/pull/263/commits/72bd4b04e40b85bae8773affc4cf041357ef23d0)) introduces a new scheme I have been mulling over for introducing unstable or expensive stuff in a way that lets people get the new stuff without breaking.

With that introduced, I'm more comfortable finally landing this.

Gankra:
Having now properly formed a notion of what an unstable/stable feature is, I am finding myself mentally unblocked on a lot of gunk I've been meaning to garden. That last few commits:

* Add discussion of unstable features to the json-schema
* remove the sensitive/exploitability stubs completely, as they were always null and consumers should be assuming they're optional anyway.
* --raw-json is now --evil-json and properly marked as such to keep consistent with the rest of the code which refers to it as such. This is important because it is now *properly* marked as `[UNSTABLE]`, and consistent naming helps the documentation talk about unstable features. As a result, everything it uniquely outputs is now properly marked as `[UNSTABLE:evil_json]`.

Gankra:
Also: yes I know I am betraying my own CONTRIBUTING docs by not actually testing this feature at all but I would need to make a new test input that's on x86 AND has symbols and god I am so tired.

Gankra:
Want to land this so other things can build off of it.

luser:
> I would need to make a new test input that's on x86 AND has symbols and god I am so tired.

I'm pretty sure there are a few of those in the Breakpad repo, we could copy them over for testcases.

Issue #234, opened 2021-09-03
Make minidump_dump print out the memory info list


gabrielesvelto:
minidump_dump should print out the entries in the minidump memory info list like breakpad does. See [this](https://searchfox.org/mozilla-central/rev/55e8eba74b60b92d04b781f7928f54ef76b13fa9/toolkit/crashreporter/google-breakpad/src/processor/minidump.cc#4771-4788) and [this](https://searchfox.org/mozilla-central/rev/55e8eba74b60b92d04b781f7928f54ef76b13fa9/toolkit/crashreporter/google-breakpad/src/processor/minidump.cc#4584-4601)

Gankra:
Fixed by #264

PR #264, opened 2021-10-29
Implement MemoryInfoStream

### LINKED ISSUE #21: Support reading memory info list stream
There's a memory info list stream, defined as `MDRawMemoryInfoList`:
https://github.com/luser/rust-minidump/blob/e73f08c09f6177eae07f6f4f028367c61f42de1f/minidump-common/src/format.rs#L3510

It uses `MD_MEMORY_INFO_LIST_STREAM` as the stream id:
https://github.com/luser/rust-minidump/blob/e73f08c09f6177eae07f6f4f028367c61f42de1f/minidump-common/src/format.rs#L635

It has a different header than other list stream types, in that it defines the size of the header as the first field, and then the size of each list entry as well as the number of list entries, presumably for forwards-compatibility. Breakpad doesn't currently do anything very smart here, it just rejects the stream if the types aren't the sizes it expects:
https://chromium.googlesource.com/breakpad/breakpad/+/54fa71efbe50fb2b58096d871575b59e12edba6d/src/processor/minidump.cc#4541

We should support reading this stream. The `MinidumpMemoryInfoList` struct should provide methods to iterate over all list entries and look one up by address like other minidump lists.

Gankra:
It's dumped in mindump_dump, and parsed by minidump-processor (but not yet used).

Also did some gardening on MemoryStream since a lot of this code was copy pasted
and I wrote back any cleanups I made to the new code.

Fixes #21

### DIFF
--- a/minidump-processor/src/processor.rs
+++ b/minidump-processor/src/processor.rs
@@ -164,7 +164,10 @@ where
         // Just give an empty list, simplifies things.
         Err(_) => MinidumpUnloadedModuleList::new(),
     };
-    let memory_list = dump.get_stream::<MinidumpMemoryList>().ok();
+    let memory_list = dump.get_stream::<MinidumpMemoryList>().unwrap_or_default();
+    let _memory_info_list = dump
+        .get_stream::<MinidumpMemoryInfoList>()
+        .unwrap_or_default();
 
     // Get memory list
     let mut threads = vec![];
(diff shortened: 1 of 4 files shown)


Gankra:
Shoutouts to past me for doing the work to write all the generic ex_list code, made this nice and mindless to implement (once I remembered how any of this worked, been a while!).

Gankra:
Oh, also fixes #234 I guess!

gabrielesvelto:
> Oh, also fixes #234 I guess!

Yes, thank you! Being able to process the memory info list stream will be extremely useful to improve the stack walker output after we deploy it to Socorro and we move to the enhancement phase. Besides validating pointers to executable memory there's a couple more features I thought of. I filed bug [1728963](https://bugzilla.mozilla.org/show_bug.cgi?id=1728963) for doing memory measurements and bug [1738651](https://bugzilla.mozilla.org/show_bug.cgi?id=1738651) for detecting potential bit-flips in the crash.

Issue #256, opened 2021-10-22
Consider validating instruction pointers using data from the `MemoryInfoListStream`


gabrielesvelto:
As per title and https://github.com/luser/rust-minidump/pull/245#issuecomment-945812018

Currently this would only be possible on Windows. On Linux we could implement it by parsing `/proc/self/maps` which is included in the minidumps, or we could also populate the stream. I filed https://github.com/msirringhaus/minidump_writer_linux/issues/8 to experiment on doing that. On macOS we'd need to populate the stream and I filed bug [1737261](https://bugzilla.mozilla.org/show_bug.cgi?id=1737261) to implement it.

Gankra:
#264 implements this

As I beef up the processor I have been vaguely mulling this new architectural feature:

Introduce a struct in process_state.rs called something like ProcessEnv that variously contains things like:

* MemoryInfoList
* Modules
* /proc/self/maps
* OS/CPU kind

So that rather than feeding down a million random things to all the different analysis steps (stackwalk, exploitability, arg_recovery), there's just this general thing that you can get all your random global questions answered.

In particular this would be nice for abstracting an "address_is_executable" query which needs to check both MemoryInfo and `/proc/self/maps`, which is gross to do at every call site. (Although we can restrict stackwalking to just accessing one "region" of memory, heuristics like "is this a valid instruction pointer" require access to the entire MemoryInfoList.)

Gankra:
I think I mentioned this elsewhere but tossing it here to: currently "page is writeable/executable" is a fuzzy signal because it is considered best-practice for a JIT to use W^X for its generated code pages. Pages do in fact need to be updated, and they can need to be updated *while the code in that page is being semantically executed* (i.e. it's on the stack).

This means at the time of a crash, a totally valid return address could point to non-executable memory. We *really* want an explicit signal for "this page is a JIT page". I'm going to ask microsoft about the tools they have for this.

PR #268, opened 2021-11-01
Implement LinuxMaps stream

Gankra:
This is a dump of /proc/self/maps from the crashing process. Details on the format
included in the comments. Because this info does a similar thing to MinidumpMemoryInfoList
I have also introduced a UnifiedMemoryInfo type which you can use to wrap the two up
and not deal with the differences.

minidump_processor now reads both streams and sets up a UnifiedMemoryInfo, but it doesn't
yet do anything with it. Also did some cleanup on the previous patch that I noticed while
designing the unified API.

Part of #256

### DIFF
--- a/minidump/src/minidump.rs
+++ b/minidump/src/minidump.rs
@@ -99,24 +99,6 @@ pub enum Error {
     CodeViewReadFailure,
 }
 
-#[derive(Debug, Clone)]
-pub struct MinidumpMemoryInfoList<'a> {
-    /// The memory regions, in the order they were stored in the minidump.
-    regions: Vec<MinidumpMemoryInfo<'a>>,
-    /// Map from address range to index in regions. Use `MinidumpMemoryInfoList::memory_info_at_address`.
-    regions_by_addr: RangeMap<u64, usize>,
-}
-
-#[derive(Debug, Clone)]
-pub struct MinidumpMemoryInfo<'a> {
-    raw: md::MINIDUMP_MEMORY_INFO,
-    pub allocation_protection: md::MemoryProtection,
-    pub state: md::MemoryState,
-    pub protection: md::MemoryProtection,
-    pub ty: md::MemoryType,
-    _phantom: PhantomData<&'a u8>,
-}
-
 /// The fundamental unit of data in a `Minidump`.
 pub trait MinidumpStream<'a>: Sized {
     /// The stream type constant used in the `md::MDRawDirectory` entry.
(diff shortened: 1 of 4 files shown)


luser:
> This is a dump of /proc/self/mem from the crashing process

[pedantic]It's `/proc/self/maps`[/pedantic] (I'm 100% sure you know this I just want to make sure this typo doesn't wind up in the code to confuse anyone.)

Gankra:
Whoops good catch

minidump/src/minidump.rs
```diff
@@ -116,6 +111,101 @@ pub trait MinidumpStream<'a>: Sized {
...
+    /// This is the process's heap.
+    Heap,
+    /// This is the "Virtual Dynamically-linked Shared Object".
+    Vdso,
+    /// This is an anonymous mmap.
+    AnonymousMap,
+    /// Some other special kind that we don't know/care about.
+    UnknownSpecial(String),
+    /// This is a mapped file/device at the given path.
+    File(String),
```
luser, line 186: This (and other places where you have `String`) might want to actually be `Vec<u8>` for correctness' sake since they're effectively `OsString` client-side, right? We tend to try to avoid dropping data from minidumps unless it's completely busted, so it would be good to be cautious here.

Is there a useful abstraction in Rust for "some other system's `OsString`? if not, I'd support a very small abstraction like:

```rust
struct DumpOsStr<'a>(&'a [u8]);

impl<'a> Into<Cow<'a, str>> for DumpOsStr<'a> {
    fn into(self) -> Cow<'a, str> {
      String::from_utf8_lossy(self.0)
    }
}
```

minidump/src/minidump.rs
```diff
@@ -1649,6 +1722,404 @@ impl<'a> MinidumpMemoryInfo<'a> {
...
 
+impl<'a> MinidumpStream<'a> for MinidumpLinuxMaps<'a> {
+    const STREAM_TYPE: MINIDUMP_STREAM_TYPE = MINIDUMP_STREAM_TYPE::LinuxMaps;
+
+    fn read(
+        bytes: &'a [u8],
+        _all: &'a [u8],
+        _endian: scroll::Endian,
+    ) -> Result<MinidumpLinuxMaps<'a>, Error> {
+        let regions = str::from_utf8(bytes)
```
luser, line 1733: Per above I'm concerned about this making data silently unavailable if non-well-formed-utf-8 filenames are present in the contents. I know it's slightly more annoying, but using `bytes..split(|b| b == b' ')` here, then tweaking `from_line` to parse from bytes by splitting on spaces and stringifying everything but the filename would avoid that.

That being said, given that it's an edge case I fully support you filing an issue for someone to fix later, noting it here, and moving on in life.

luser (approved):
Now that you've done all this work to basically match Breakpad on feature parity, I'm interested in tackling the polish work in anticipation of releasing a 1.0 in the near future.

minidump/src/minidump.rs
```diff
@@ -116,6 +111,101 @@ pub trait MinidumpStream<'a>: Sized {
...
+    pub is_write: bool,
+    pub is_exec: bool,
+    pub is_shared: bool,
+    pub is_private: bool,
+
+    // Fields in the format we ignore (not yet useful)
+    // * offset
+    // * dev
+    // * inode
+    _phantom: PhantomData<&'a u8>,
```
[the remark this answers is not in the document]
Gankra: I vaguely wanted to allow for a future where this struct is holding pointers into the minidump. Mostly because I was copying the MinidumpMemory code and so including lifetimes was the default. I could rip it all out if you want. The only thing that might be worth pointing into is the "filename" strings (currently I read them into actual Strings because I was feeling lazy).

minidump/src/minidump.rs
```diff
@@ -116,6 +111,101 @@ pub trait MinidumpStream<'a>: Sized {
...
+}
+
+/// A broad classification of the mapped memory described by a [`MinidumpLinuxMapInfo`].
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum MinidumpLinuxMapKind {
+    /// This is the main thread's stack.
```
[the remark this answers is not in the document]
Gankra: The main thread is `[stack]` while other threads are `[stack:<tid>]`. I wasn't really sure whether this distinction was important, but since I needed to parse them separately I figured I might as well expose the difference. I'm not sure what I would fold them into, given I don't have a tid for the main thread (or is it just always 1? that sounds vaguely familiar...).

minidump/src/minidump.rs
Gankra: mm this is a good point, I will ponder on this

minidump/src/minidump.rs
Gankra: Ugh yeah you're right. Gross.

luser: You've done enough here, let's get this merged and we can deal with the fiddly shit later. It shouldn't impact the public API much, it's just implementation details.

minidump/src/minidump.rs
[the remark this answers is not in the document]
luser: Okay, that's totally reasonable, keep what you have.

minidump/src/minidump.rs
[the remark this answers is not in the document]
luser: Oh, good point! I have a thought: how about you swap the `String`s for `Cow<'a, str>`, just make them always `Cow::Owned` with what you have now, and then that will let you use borrowed strings later, or if we get motivated to handle the invalid utf-8 case it will allow for that as well?

minidump/src/minidump.rs
gabrielesvelto: Yeah, on Linux that's definitely a `Vec<u8>` because traditional Unix paths have no encoding, they're just raw sequences of bytes. Nobody is crazy enough to use non-printable paths on their machines but we have to keep in mind that we're parsing untrusted data so we can't assume those are always valid strings.

luser: It's at least a tiny bit better than most other things since this data comes directly from the kernel and Breakpad includes it wholesale into the minidump, but the filenames do come from userland so I'm sure it's trivially easy for it to wind up as WTF-8 in practice if some code decides to write out a temp file with part of the name interpolated from other data and then mmap's it.

I guess you could fairly easily dump the maps streams from a random sample of Firefox Linux crash dumps and determine what percentage, if any, fail to parse as valid UTF-8.

PR #269, opened 2021-11-01
Stub out implementations of the stringy linux streams

### LINKED ISSUE #207: Implement MINIDUMP_STREAM_TYPE::LinuxCpuInfo
This contains a dump of `/proc/cpuinfo`, in much the same way the implement `LinuxLsbRelease` contains a dump of `/etc/lsb-release`.

This at very least contains `system_info.cpu_microcode_version` in our target JSON schema.

Low priority, potentially a good starter bug.

Gankra:
Fixes #207

Adds implementations for:
* LinuxCpuInfo (/proc/cpuinfo)
* LinuxProcStatus (/proc/self/status)
* LinuxEnviron (/proc/self/environ)

The only new data we collect is LinuxCpuInfo.microcode_version, which gets fed through to
cpu_microcode_version in the json schema. All the code for the other streams is written,
so it should now be trivial to grab whatever key we decide is interesting with only a few
new lines.

LinuxAuxv, LinuxCmdLine and LinuxDsoDebug have been excluded from
this patchset as they all have more custom formats than the KEY SEPARATOR VALUE formats
of the implemented streams.

(This PR is built on top of the previous one to avoid merge conflicts, only the second patch is new.)

### DIFF
--- a/minidump-processor/src/process_state.rs
+++ b/minidump-processor/src/process_state.rs
@@ -596,7 +596,7 @@ Unknown streams encountered:
                 "cpu_count": sys.cpu_count,
                 // TODO: Issue #19
                 // optional
-                "cpu_microcode_version": null,
+                "cpu_microcode_version": sys.cpu_microcode_version,
             },
             "crash_info": {
                 "type": self.crash_reason.map(|reason| reason.to_string()),
--- a/minidump-processor/src/processor.rs
+++ b/minidump-processor/src/processor.rs
@@ -110,14 +110,20 @@ where
         .cpu_info()
         .map(|string| string.into_owned());
 
+    let linux_standard_base = dump.get_stream::<MinidumpLinuxLsbRelease>().ok();
+    let linux_cpu_info = dump.get_stream::<MinidumpLinuxCpuInfo>().ok();
+    let _linux_environ = dump.get_stream::<MinidumpLinuxEnviron>().ok();
+    let _linux_proc_status = dump.get_stream::<MinidumpLinuxProcStatus>().ok();
+
     let system_info = SystemInfo {
         os: dump_system_info.os,
         os_version: Some(os_version),
         cpu: dump_system_info.cpu,
         cpu_info,
+        cpu_microcode_version: linux_cpu_info.and_then(|info| info.microcode_version),
         cpu_count: dump_system_info.raw.number_of_processors as usize,
     };
-    let linux_standard_base = dump.get_stream::<MinidumpLinuxLsbRelease>().ok();
+
     let mac_crash_info = dump
         .get_stream::<MinidumpMacCrashInfo>()
         .ok()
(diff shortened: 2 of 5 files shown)


luser:
> LinuxAuxv, LinuxCmdLine and LinuxDsoDebug have been excluded from
> this patchset as they all have more custom formats than the KEY SEPARATOR VALUE formats
> of the implemented streams.

The first and last of those are not text-based at all, and the middle is strings separated by null terminators, right? Having an accessor to get the cmdline as a `Vec<&'a [u8]>` would be nice but probably not necessary for crash-stats purposes. I can't recall if the other two are actually used by Breakpad's crash processing at all, and I'm out of brainpower to look into it tonight.

luser:
Y'know, after reading your `/proc/self/maps` patch and now looking at this, I wonder if it would be worthwhile to figure out sharing parsing code with some existing crate that parses these for its own process? I know I've encountered at least one of those before.

The overhead of making that possible might outweigh the value of the shared code, certainly.

Gankra:
> 
> 
> > LinuxAuxv, LinuxCmdLine and LinuxDsoDebug have been excluded from
> > this patchset as they all have more custom formats than the KEY SEPARATOR VALUE formats
> > of the implemented streams.
> 
> The first and last of those are not text-based at all, and the middle is strings separated by null terminators, right? Having an accessor to get the cmdline as a `Vec<&'a [u8]>` would be nice but probably not necessary for crash-stats purposes. I can't recall if the other two are actually used by Breakpad's crash processing at all, and I'm out of brainpower to look into it tonight.

afaik this is all correct. Certainly none of these streams affect the current json schema minidump-stackwalk generates for socorro. I don't know if something like exploitability takes a peak or something (doubtful).

I certainly could implement the CmdLine parser, but that would be a bunch of *new* code that literally isn't for anything until we find a reason to do it. I only implemented these maybe-useless streams since it was just a copy-paste of the code for the actually useful ones. :)

Auxv and DsoDebug I *truly* have no idea what they are. Like, even looking it up I just get back something that is totally opaque to me. 🤷‍♀️

luser:
> Auxv and DsoDebug I _truly_ have no idea what they are. Like, even looking it up I just get back something that is totally opaque to me. woman_shrugging

I think `DsoDebug` might be a Breakpad-invented translation of some internal `ld.so` state? `Auxv` is just an array of crap that the kernel stuffs in memory because `ld.so` needs it. The contents are available in a live process via [`getauxval`](https://man7.org/linux/man-pages/man3/getauxval.3.html) (that man page documents what most of them are). In any event, you don't need them. In fact, IIRC the only reason Breakpad includes them is to make its `mindump2core` tool work?

Gankra:
Small fix from actually testing on real data:

Now properly parse the "microcode" value as 0xSOMEHEX instead of SOMEDECIMAL. Confirmed on one minidump that it does the same thing as socorro.

gabrielesvelto:
> I think `DsoDebug` might be a Breakpad-invented translation of some internal `ld.so` state?

It looks like it stores the ELF headers of the various loaded modules as they appear in the process' memory. It doesn't seem to serve any purpose outside of `minidump2core`.

Issue #273, opened 2021-11-02
Expose the raw iterator for key-value linux streams?


Gankra:
A bunch of linux streams are just a bunch of string key-value pairs. Right now we just pick out a few interesting values, stuff them in a struct, and call it a day. But it would probably be more useful to expose the raw key-value iterator so that users can scan the list and pull out other values on demand.

This requires a bit more thinking about my current unquoting scheme. Maybe expose versions with and without the unquoting stuff? Not sure if there are situations where `trim`ing the values would be inappropriate.

Also not sure if the current iterator impls would have weird lifetime problems. Certainly all those streams would need to have a `'mdmp` lifetime to be able to create the iterator on demand.

At that point I guess you could also expose the raw byte buffer too.

luser:
I'd probably hold off on overdesigning the API until you find yourself needing something more than what you have. There's already API surface to get raw stream contents as bytes, which is an OK fallback position for these simple types of streams.

Gankra:
Fixed by #275

PR #275, opened 2021-11-03
Linux cleanup

### LINKED ISSUE #272: Create a proper abstraction for linux os strings
In handling all the Linux* streams I absentmindedly parsed everything as utf8 since I know that's ~basically true. But things like paths (which are a lot of the values we process) are technically just "arbitrary bytes", so we should have a proper LinuxOsString abstraction for the output and do the parsing/matching on byte-strings.

Right now nothing *bad* will happen if this assumption is broken. We'll just fail to parse whatever data we're reading and come back with default/empty values for the stream. Just a quality of implementation thing, and probably not one that will come up often (since non-utf8 paths are hard to make in the first place).

Gankra:
This adds a new strings submodule with LinuxOsString and LinuxOsStr types that provide
most of the functionality we need. Ultimately though minidump_processor feeds things that
truly are LinuxOsStrings through String::from_utf8_lossy because we need to do *something*
to output them to json/cli.

This also moves all the linux streams to properly using Cow and lifetimes.

Also includes a random bugfix I noticed while refactoring.

fixes #272


Gankra:
New commits added:

* implemented my proposed new api for LinuxStreams (Fixes #273)
* pulled synth_minidump out to its own crate so other crates could depend on it as a dev_dependency and use it in their tests
* linux stream tests largely move to minidump_processor as only it actually parses them


I really didn't like the way we were handling these streams, the new refactor to have them be just key-value iterators seems way more principled and useful. This is a breaking change in that now minidump will not parse out any fields for LSB or CpuInfo, instead deferring the task to SystemInfo.

Gankra:
Was this too much effort for a minimally important set of APIs? Maybe.

Was this bothering me too much to leave it be? Absolutely.

minidump-processor/Cargo.toml
```diff
@@ -38,3 +38,4 @@ scroll = "0.10.2"
 
 [dev-dependencies]
 test-assembler = "0.1.5"
+synth-minidump = { path = "../synth-minidump" }
```
luser, line 41: This is a good idea and I'm not sure why I never got around to doing it!

luser:
I'm not going to approve this because that would imply that I reviewed it and I haven't, but it sounds good! Unless you explicitly request my review on things I work under the assumption that you will merge your own changes as needed.

minidump-processor/Cargo.toml
Gankra: I assumed you (reasonably) thought the subcrates could grab it as a pub symbol in test builds but sadly It Doesn't Work Like That/

Gankra:
Yeah I'm pretty happy with this. Always appreciate just a high level "this makes sense" though.

Issue #265, opened 2021-10-29
minidump_dump doesn't handle windows thread stack RVA bug


Gankra:
I noticed this problem and fixed it in minidump-processor, but forgot about minidump_dump (because I never use it and don't especially understand its purpose). This was annoying when I wanted dumps of stack memory and remembered minidump_dump is supposed to do that!

Unfortunately it can't because of this issue. It's annoying to fix at the level of stream parsing because you need to pull up the MemoryStream to do a proper by-address-lookup. The easy direct fix is to make MinidumpThreadList::print (optionally?) take a MemoryStream so that it can fix up the stack memory too.

Alternatively you could give MinidumpThreadList a `fix_stacks` method that takes the MemoryStream and fixes every thread's stack (and then minidump-processor and minidump-dump can both just invoke that).

https://github.com/luser/rust-minidump/blob/b75c795348c23e195b7eee953c323df993916e15/minidump/src/minidump.rs#L1579-L1587

https://github.com/luser/rust-minidump/blob/b75c795348c23e195b7eee953c323df993916e15/minidump-processor/src/processor.rs#L192-L199

gabrielesvelto:
minidump_dump is mostly used when you're dealing with incomplete, corrupted or odd minidumps and you just want to see the raw data in the various streams, no questions asked.

Gankra:
I'm not sure if that's for or against fixing this "bug".

Gankra:
Fixed by #289

Issue #152, opened 2021-05-04
use MISC_INFO_5 contents to output XSAVE contents in cpu context


Gankra:
In newer versions of windows the cpu context of each thread now includes a dynamic layout/sized section with Even More Registers based on XSAVE. The MISC_INFO_5 format (which we now parse) describes the high-level layout and kinds of sections. However interpreting the contents further requires rust-minidump having a native understanding of the format. I do not know if this is totally vendor/cpu-specific, but at very least Intel documents the layout of each of its XSAVE sections in *Volume 1, Chapter 13 of the "Intel 64 and IA-32 Architectures Software Developer’s Manual"*.

This will require some API design work, as it will naively introduce the notion of the MinidumpThread stream depending on the contents of the MiscInfo stream, which doesn't yet happen. It's possible this could be "handled" by minidump-processor, although that's kinda bad UX for anyone else using rust-minidump on its own?

I proposed one way of doing this in https://github.com/luser/rust-minidump/pull/144

Gankra:
#289 got everything piped and wired up to the right place, but for now we're punting on actually parsing out the extra sections, absent a reason to care about those values (and some API design work needs to be done to expose the information).

See https://github.com/luser/rust-minidump/blob/16a7f1b392ed998ad0d02054f27770f29e066a56/minidump/src/context.rs#L626-L628

Gankra:
Since this could change the way we expose contexts, it's a 1.0 blocker

Gankra:
Other useful things for this:

The printing code shows you how to enumerate/analyze the xstate entries mechanically:

https://github.com/luser/rust-minidump/blob/16a7f1b392ed998ad0d02054f27770f29e066a56/minidump/src/minidump.rs#L3164-L3176

And this comment details how to interpret the xstate entries *logically* (they're weird!):

https://github.com/luser/rust-minidump/blob/16a7f1b392ed998ad0d02054f27770f29e066a56/minidump-common/src/format.rs#L7515-L7560

([comment rendered here](https://docs.rs/minidump-common/0.9.4/minidump_common/format/struct.XSTATE_CONFIG_FEATURE_MSC_INFO.html#the-xstate-format-in-minidumps))

PR #278, opened 2021-11-04
WIP stream dependencies

Gankra:
Introduces the ability for a stream to statically depend on another, allowing stream parsing to be refined by the contents of other streams (e.g. using MinidumpMemoryList to repair thread stack pointers or MiscInfo to resolve CpuContexts).

### DIFF
--- a/minidump-processor/src/processor.rs
+++ b/minidump-processor/src/processor.rs
@@ -85,11 +85,17 @@ where
     let memory_info_list = dump.get_stream::<MinidumpMemoryInfoList>().ok();
     let linux_maps = dump.get_stream::<MinidumpLinuxMaps>().ok();
     let _memory_info = UnifiedMemoryInfoList::new(memory_info_list, linux_maps).unwrap_or_default();
+    let misc_info = dump.get_stream::<MinidumpMiscInfo>().ok();
 
     // Thread list is required for processing.
+    // Pack MinidumpThreadList deps up.
+    let thread_list_deps = (memory_list, misc_info);
     let thread_list = dump
-        .get_stream_with_dep::<MinidumpThreadList>(memory_list.as_ref())
+        .get_stream_with_dep::<MinidumpThreadList>(Some(&thread_list_deps))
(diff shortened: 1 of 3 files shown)


Gankra:
This came to me as I was falling asleep last night.

It's the potential fix to #265 and #152, need to mull on it.

Gankra:
Unfortunately this *precise* solution isn't sufficient because MinidumpThreadList is the problematic child for *both* of the motivating problems, so it actually needs to depend on two streams (MinidumpMiscInfo and MinidumpMemoryList). *sigh*

Gankra:
Version 2 solves the problem of multiple dependencies by using a callback to make everything more general. It makes things a bit more complicated but it's not the worst?

Multiple dependencies require a kinda gross packing/unpacking in processor because you can't properly give things lifetimes otherwise afaict.

luser:
I think this concept is sensible, I'm just trying to imagine an API that supports this in a better way. I guess one way to do this would be to split the `MinidumpStream` trait up into three parts, like:

- `MinidumpStreamBase`: bits common to all streams.
- `MinidumpStandaloneStream`: the current behavior of "read a stream with no dependencies".
- `MinidumpDependentStream`: something like what you have here, for streams that require other streams.

IDK, I'd probably have to play with it to see how it felt. OTOH I'm not *overly* concerned about `MinidumpStream` being quirky because all the implementations are in this crate. As long as the public API for "read a stream" isn't super awkward I think it'll be fine.

Breakpad just completely punts here and lets every stream object hang on to a `shared_ptr<Minidump>` or something like that. I guess if you wanted you could make `MinidumpStream` take `Rc<Minidump>`? That'd make multithreaded usage annoying, but maybe that's not a particularly important use case.

Gankra:
Ah yeah I hadn't thought about making dependent streams a seperate trait with a separate path. That would certainly make the deps explicit and prevent users from accidentally double parsing dependencies. But also it would be a pain when you just don't Care. Interesting...

Gankra:
Closing in favour of #289

PR #285, opened 2021-11-09
Use the thread names in the evil json

### LINKED ISSUE #284: Implement extracting thread names from evil_json
I thought we had implemented writing them to ThreadNamesStream in firefox's minidump writer but it seems like this isn't the case, at least for https://crash-stats.mozilla.org/report/index/422d8af0-1681-45f4-a290-cd8120211109

I forgot about doing this since we briefly agreed to not implement evil_json support, but I ended up adding it, so shouldn't be too hard to do.

Schema:

ThreadIdNameMapping: "5177: "IPC I/O Parent",5178:"Timer",...."

(yes it's a weird json object in a string, just like cert info)

Gankra:
Fixes #284

### DIFF
--- a/minidump-processor/src/evil.rs
+++ b/minidump-processor/src/evil.rs
@@ -1,0 +1,103 @@
+use log::error;
+use serde_json::map::Map;
+use serde_json::Value;
+use std::collections::HashMap;
+use std::fs::File;
+use std::io::BufReader;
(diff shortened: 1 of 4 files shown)


Gankra:
Introducing ProcessorOptions here as I anticipate needing/wanting it for other borderline features like #263

PR #286, opened 2021-11-09
Implement last_error_value

### LINKED ISSUE #156: Extract LastErrorValue from Thread TEBs
See my implementation of this in breakpad: https://phabricator.services.mozilla.com/D107939

Gankra:
Fixes #156

### DIFF
--- a/minidump/src/minidump.rs
+++ b/minidump/src/minidump.rs
@@ -2238,6 +2238,26 @@ impl<'a> MinidumpThread<'a> {
         writeln!(f)?;
         Ok(())
     }
+
+    /// Gets the last error code the thread recorded, just like win32's GetLastError.
+    ///
+    /// The value is heuristically converted into a CrashReason because that's our
+    /// general error code handling machinery, even though this may not actually be
+    /// the reason for the crash!
+    pub fn last_error(&self, cpu: Cpu, memory: &MinidumpMemoryList) -> Option<CrashReason> {
+        // Early hacky implementation: rather than implementing all the TEB layouts,
+        // just use the fact that we know the value we want is a 13-pointers offset
+        // from the start of the TEB.
+        let teb = self.raw.teb;
+        let pointer_width = cpu.pointer_width()?;
+        let offset = pointer_width.checked_mul(13)?;
+        let addr = teb.checked_add(offset)?;
+        let val: u32 = memory
+            .memory_at_address(addr)?
+            .get_memory_at_address(addr)?;
+
+        Some(CrashReason::from_windows_code(val))
+    }
 }
 
 impl<'a> MinidumpStream<'a> for MinidumpThreadList<'a> {
--- a/minidump/src/minidump.rs
+++ b/minidump/src/minidump.rs
@@ -2256,7 +2256,7 @@ impl<'a> MinidumpThread<'a> {
             .memory_at_address(addr)?
             .get_memory_at_address(addr)?;
 
-        Some(CrashReason::from_windows_code(val))
+        Some(CrashReason::from_windows_error(val))
     }
 }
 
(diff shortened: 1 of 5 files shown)


Gankra:
```
socc-pair --crash-id=3f131f66-733d-43f7-8005-c6f700211109 
```
seems happy with this, except I guess we don't have STATUS_TIMEOUT in our definitions:

```
-      did not match
+      last_error_value: STATUS_TIMEOUT
-      last_error_value: unknown 0x00000258
```

minidump/src/minidump.rs
```diff
@@ -2238,6 +2238,26 @@ impl<'a> MinidumpThread<'a> {
...
+        // from the start of the TEB.
+        let teb = self.raw.teb;
+        let pointer_width = cpu.pointer_width()?;
+        let offset = pointer_width.checked_mul(13)?;
+        let addr = teb.checked_add(offset)?;
+        let val: u32 = memory
+            .memory_at_address(addr)?
+            .get_memory_at_address(addr)?;
+
+        Some(CrashReason::from_windows_code(val))
```
gabrielesvelto, line 2259: We do things a little differently in the existing stackwalker. First we try to convert the value into an NTSTATUS code and if that fails we try one of the winerror.h codes. If neither works we just stick to the raw value. So this would call for a dedicated enum which can be either an `NtStatusWindows` value, a `WinErrorValue` or a raw `u32`.

Gankra: Yes I came to this implementation after considering that implementation and our CrashReason::from_exception code. Is this not strictly better? (And if not, why does that logic not propagate back to CrashReason::from_exception?).

For reference, the impl of from_windows_code in this patch is:

```
    pub fn from_windows_code(exception_code: u32) -> CrashReason {
        if let Some(err) = md::ExceptionCodeWindows::from_u32(exception_code) {
            CrashReason::WindowsGeneral(err)
        } else if let Some(err) = md::WinErrorWindows::from_u32(exception_code) {
            CrashReason::WindowsWinError(err)
        } else if let Some(err) = md::NtStatusWindows::from_u32(exception_code) {
            CrashReason::WindowsNtStatus(err)
        } else {
            CrashReason::WindowsUnknown(exception_code)
        }
    }
```

(Oh I just noticed that this inverts the order of the two conversions, not sure how big the overlap is...)

gabrielesvelto: Exception codes were strictly meant for minidumps so they shouldn't appear in `GetLastError()` and some of them overlap with NTSTATUS IIRC. Double-checking the headers though the fact that we're trying winerror.h codes first and then NTSTATUS makes sense, and it would make sense for `get_last_error()` too. The range in which winerror.h and NTSTATUS overlap is mostly made of actual "statuses" and not errors, so those are far more likely to come from winerror.h than vice-versa.

In short the only difference here would be not to try and convert to exception codes, the rest should be identical. Maybe something like:
```rust
pub fn from_windows_code(exception_code: u32) -> CrashReason {
    if let Some(err) = md::ExceptionCodeWindows::from_u32(exception_code) {
        CrashReason::WindowsGeneral(err)
    } else {
        from_windows_error(exception_code)
    }
}

pub fn from_windows_error(error_num: u32) -> CrashReason {
    if let Some(err) = md::WinErrorWindows::from_u32(error_num) {
        CrashReason::WindowsWinError(err)
    } else if let Some(err) = md::NtStatusWindows::from_u32(error_num) {
        CrashReason::WindowsNtStatus(err)
    } else {
        CrashReason::WindowsUnknown(error_num)
    }
}
```

gabrielesvelto:
> ```
> socc-pair --crash-id=3f131f66-733d-43f7-8005-c6f700211109 
> ```
> 
> seems happy with this, except I guess we don't have STATUS_TIMEOUT in our definitions:
> 
> ```diff
> -      did not match
> +      last_error_value: STATUS_TIMEOUT
> -      last_error_value: unknown 0x00000258
> ```

Ah yes, I think I only added the errors and warnings to the definitions, not the "harmless" statuses. I'll file an issue to add them and also update the definitions given Microsoft released two new SDKs.

Gankra:
nice, your PR filled in the STATUS_TIMEOUT as expected :)

Gankra:
New commit should address concerns.

PR #289, opened 2021-11-10
WIP Lazily parse stack memory and cpu contexts

Gankra:
This is an alternative version of #278, which instead of weird magic traits that will do double-parsing behind your back, we just defer parsing the context/stack_memory until the user requests them, at which point we can demand dependent streams to do the task properly. I realized this was a reasonable approach when I made a similar api for #286.

Using this info for contexts is only stubbed out, but stack repair is implemented.

This is basically no extra burden to a legitimate user since they're going to want to parse out MemoryList and SystemInfo, and MiscInfo is optional. It's slightly more annoying for tests because we need to mock SystemInfo. For now I used the sloppy garbage SystemInfo I made for Linux integration tests in `processor`. I'll deal with it when I implement actually using the SystemInfo (and presumably break the tests).

Biggest "problem" is that this breaks the encapsulation of `print`, but since those are basically only there for dump_syms and aren't actually part of a trait or anything, we could just make them arguments to those print methods.

This is also more forwards compatible than #278 because we can introduce new methods that take additional arguments if we find places where analysis needs to be refined more.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,13 +11,15 @@ It's fairly heavily modeled after the [Google Breakpad](https://chromium.googles
 Print the raw details of the exception stream from a minidump:
 
 ```rust
-use minidump::{Error, Minidump, MinidumpException, MinidumpStream};
+use minidump::{Error, Minidump, MinidumpMiscInfo, MinidumpSystemInfo, MinidumpException, MinidumpStream};
 use std::io::{self, Write};
 
 fn work() -> Result<(), Error> {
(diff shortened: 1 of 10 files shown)


Gankra:
@luser what do you think? Gonna sleep on it but I personally think this is the approach we should go with!

Gankra:
Implemented the new Context parsing using SystemInfo in the new patch.

Gankra:
Happy with this, time to cut a new release!

luser:
I skimmed it and the approach seems fine. :)

PR #301, opened 2021-11-18
Add a timeout to symbol downloads

### LINKED ISSUE #300: breakpad-symbols: Add a bandwidth/time limit to symbol downloading
Deploying to mozilla's staging servers, a couple minidumps timed out because the .sym file download from AWS was infinite(???). Navigating to the URL I can confirm that firefox also is completely baffled by these URLs. We need some way for the symbol downloader to give up on symbols files that act like this and proceed on without it.

Gankra:
This also introduces a --symbol-download-timeout-secs flag to minidump-stackwalk
so it's configurable. The default is 1000 seconds (which for a 300MB .sym file
requires a connection speed of 300KB/s).

Weirdly I think adding the timeout at all seems to unstick the download, but I'm
not sure. Either way you can tune the value to your needs.

Fixes #300

### DIFF
--- a/breakpad-symbols/src/lib.rs
+++ b/breakpad-symbols/src/lib.rs
@@ -42,6 +42,7 @@ use std::fmt;
 use std::fs;
 use std::io::{self, Read, Write};
 use std::path::{Path, PathBuf};
+use std::time::Duration;
 
 pub use minidump_common::traits::Module;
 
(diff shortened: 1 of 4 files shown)


Gankra:
Surprisingly this timeout not only unblocks mdsw, but it also successfully resolves the file?? Maybe we're getting a truncated version or it just gets loopy at the EOF. Either way we manage to parse the symbols which is neat.

gabrielesvelto:
> Surprisingly this timeout not only unblocks mdsw, but it also successfully resolves the file?? Maybe we're getting a truncated version or it just gets loopy at the EOF. Either way we manage to parse the symbols and run the cfi, which is neat.

I love it when things work and we don't know why :smirk:

Gankra:
New version, more robust, more confusing

Issue #303, opened 2021-11-18
remove dwarf_symbolizer.rs


Gankra:
https://github.com/luser/rust-minidump/blob/master/minidump-processor/src/dwarf_symbolizer.rs

I've kept it technically functioning for a while for the sake of not breaking anything, but at this point I feel comfortable enough with the code to say we should just rip it out and all of its aging dependencies. We can add it back in later when we have a proper design/plan for native symbols.

PR #306, opened 2021-11-19
garbage-collect, update, and unify some dependencies

Gankra:
This removes like, 40 crates from a full build of the workspace. Main source of duplication at this point is minidump-tools' use of rust-disasm, which pulls in a lot of Very Old versions of things. Most of the diff is just running `cargo update`.

Also implements #303 

(I started with that and then got more aggressive once I saw how little the Cargo.lock changed.)


luser:
Go ahead and rip off the band-aid and get rid of `minidump-tools`, it's worthwhile now that you're actively doing this work.

PR #307, opened 2021-11-19
remove derilect experiments, update and gc dependencies

### LINKED ISSUE #303: remove dwarf_symbolizer.rs

### LINKED ISSUE #305: Figure out minidump-tools
Up to this point I've completely avoided understanding what minidump-tools is. I just pop in there and fixup code when I change an API and run away. It has no README, it has no docs. All I know is it's a weird library with some micro-binaries wrapping it.

I think whatever it's trying to do is almost certainly better handled by improving minidump-stackwalk or minidump-dump. If we conclude otherwise, then it needs some work to be able to publish it properly. We would need separate top-level crates for each binary, and we need to do something about the unconditional dependency on https://github.com/luser/rust-disasm, which makes Cargo reject publishing it at all.

If we determine it doesn't do anything we care about, we should just delete it and remove the maintenance burden.

Gankra:
Wanted to see how much this affects the project. About ~120 crates removed from the workspace's crate graph!

Fixes #305 
Fixes #303


minidump-stackwalk/src/main.rs
```diff
@@ -313,7 +312,6 @@ native debuginfo formats. We recommend using a version of dump_syms to generate
                 symbols_paths,
```
luser, line 312: You can now just drop the use of `MultiSymbolProvider` entirely here, and just do like:

```rust
let mut provider = if !symbols_urls.is_empty() {
    Box::new(Symbolizer::new(http_symbol_supplier( /* … */ )))
else {
    Box::new(Symbolizer::new(simple_symbol_supplier( /* … */ )))
};
```

breakpad-symbols/Cargo.toml
```diff
@@ -18,7 +18,7 @@ minidump-common = { version = "0.9.3", path = "../minidump-common" }
 range-map = "0.1.5"
 nom = "~1.2.2"
 log = "0.4.1"
-reqwest = { version = "0.11.3", features = ["blocking", "gzip"] }
+reqwest = { version = "0.11.6", features = ["blocking", "gzip"] }
```
luser, line 21: Given your hassle with `openssl` I'm questioning whether having the HTTP dependency down here in `breakpad-symbols` is the best choice or not. It is nice to be able to have a `HTTPSymbolProvider` available, and crates that punt entirely and give you a generic "bring your own HTTP implementation" API are super awkward to work with, but `reqwest` also hides a pretty big dependency tree behind it.

It's probably not worth spending too much time on here. In the long run it would be better to just make `minidump-processor` interop with `symbolic`, since Sentry's set of crates have solved all of these problems.

luser: n.b.: you may have seen the discussions about merging this `breakpad-symbols` crate with Sentry's equivalent. I originally thought the best plan would be to throw away my crate and use theirs, but it sounds like they don't love theirs either. Now that you've put additional work into this one and it is at least as useful as the canonical Breakpad implementation, I think there's a pretty strong argument for blessing this one as canonical.

luser (approved):

minidump-stackwalk/src/main.rs
Gankra: While this is true, I like keeping the machinery here for future improvements (and to keep it working). The overhead has got to be completely trivial.

breakpad-symbols/Cargo.toml
Gankra: Interesting thoughts I dare not act on at this precise moment.

PR #312, opened 2021-11-23
Give better feedback on corrupt minidumps 

Gankra:
Our error types have Display strings but we weren't using them. For a couple strings
I've added a parenthetical hint, reflecting the common cause of these two errors
(from what I'm seeing in socorro).

I've also added some `warn`s for duplicate entries, to hopefully make it easier
to notice when this happens. In particular socorro gets many MissingThreadList
minidumps which are just full of 13 dummy UnusedStream entries. Looking at the
minidump in minidump_dump is really confusing, because it says there's 13 but
only prints 1. The logs now make it clearer, as you see several copies of:

```
[WARN] Minidump contains multiple streams of type 0 (UnusedStream) at indices 0 (0 bytes) and 1 (0 bytes) (using 1)
```

(It will still be confusing to look at in minidump_dump, we should really hook it
up with proper logging...)

### DIFF
--- a/minidump-stackwalk/src/main.rs
+++ b/minidump-stackwalk/src/main.rs
@@ -296,49 +296,52 @@ native debuginfo formats. We recommend using a version of dump_syms to generate 
 
     // Ok now let's do the thing!!!!
 
-    if let Ok(dump) = Minidump::read_path(minidump_path) {
-        let mut provider = MultiSymbolProvider::new();
+    match Minidump::read_path(minidump_path) {
+        Ok(dump) => {
+            let mut provider = MultiSymbolProvider::new();
 
-        if !symbols_urls.is_empty() {
-            provider.add(Box::new(Symbolizer::new(http_symbol_supplier(
-                symbols_paths,
-                symbols_urls,
-                symbols_cache,
-                symbols_tmp,
-                timeout,
-            ))));
-        } else if !symbols_paths.is_empty() {
-            provider.add(Box::new(Symbolizer::new(simple_symbol_supplier(
-                symbols_paths,
-            ))));
-        }
+            if !symbols_urls.is_empty() {
+                provider.add(Box::new(Symbolizer::new(http_symbol_supplier(
+                    symbols_paths,
+                    symbols_urls,
+                    symbols_cache,
+                    symbols_tmp,
+                    timeout,
+                ))));
+            } else if !symbols_paths.is_empty() {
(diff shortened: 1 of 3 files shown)


luser:
> In particular socorro gets many MissingThreadList
minidumps which are just full of 13 dummy UnusedStream entries.

Maybe you're already aware, but FWIW the root cause of this is that the minidump-writing code (for sure in Breakpad, but maybe also in Microsoft's implementation) pre-allocates the minidump stream directory to a fixed size, zeroes it out, and then fills in each entry while writing the stream contents. If it fails to write a stream it bails and leaves that directory entry as well as any remaining ones after it zeroed out which cause this issue.

luser:
Also: let's add "switch from `failure` to a crate that's more well-maintained" to the 1.0 checklist. I've had good experiences with `thiserror` but I know there's a bunch of experimentation in this area. Ultimately I don't think this should have too much effect on the public API, but being intentional about the error type we expose, the enumerated set of errors it can contain, and the traits it implements will be a good thing.

luser:
Yeah, it's a PITA. I don't think there's a lot of value in trying to produce errors that API consumers can differentiate in code to handle specific situations. For this crate either you can read a piece of data or you can't, and the specific reason you can't read it isn't going to give you a means to work around the problem (like you might with APIs in other domains via retrying or using some alternate means to do the same thing).

I do think there'd be value to both crash-stats and other users of the crates in ensuring that whatever error type is returned, it supports iterating through a chain of error causes, and that this crate inserts context into that chain at an appropriate level of detail. ("Appropriate" here is not a well-defined concept.) As a strawman example:

```
Error processing minidump: Failed to read thread list stream.
  Caused by: Failed to read thread descriptor for thread 4
  Caused by: Failed to read NNN bytes at offset 0x123 in minidump
```

That seems achievable, and also provides sufficient detail to indicate what exactly went wrong, and even a starting point to investigate fixing things if the issue is in the client-side minidump writing code.

While writing that example I had a follow-on idea: there are many situations that `minidump-processor` attempts to handle as gracefully as possible so that it can surface *some* data from a minidump even if normally-available bits are missing. What if it recorded the non-fatal errors produced during processing and made them available in `ProcessState` as something like `non_fatal_errors: Vec<Error>`? You could easily serialize that into the JSON and make it visible in crash-stats so that if someone encountered a crash that appeared to be missing information you'd have a record of *why*. Like "oh there's only one frame in the crashing stack because the memory region for that thread was unreadable".

PR #314, opened 2021-11-25
add a bunch of docs to minidump

Gankra:
I was gonna do more cleanup on individual types and methods but the top level docs (which seemed more important) completely wiped me out.


gabrielesvelto:
Thanks for doing this, this is a good place to put this documentation and also to cover the extensions both we and Google did to the format. I'll file some bugs about that as soon as I have some spare time.

PR #325, opened 2021-12-01
Document the Symbolizer interface and shim better

Gankra:
Also repairs the symbolic shim so that it compiles again.

### DIFF
--- a/minidump-processor/src/symbols.rs
+++ b/minidump-processor/src/symbols.rs
@@ -1,3 +1,98 @@
+//! This module defines the interface between minidump-processor and its [Symbolizer][].
+//!
+//! There can only be one [Symbolizer][], and this is configured by minidump-processor's Cargo
+//! feature flags. The currently defined Symbolizers are:
+//!
+//! * breakpad_symbols -- feature: breakpad-syms (currently the default)
+//! * symbolic -- feature: symbolic-syms (not yet implemented, but compiles)
+//!
+//!
+//! minidump-processor and the [Symbolizer][] communicate using a series of traits. The symbolizer
+//! must provide implementations of these traits:
+//!
+//! * [SymbolProvider][] - provides symbolication, cfi evaluation, and debug statistics
+//!     * Implemented by [Symbolizer][]
+//!     * This actually doesn't need to be a trait in the current design, it exists to allow
+//!       multiple symbolicators to be used together, via [MultiSymbolProvider][]. The other
+//!       SymbolProviders have been removed, but I figured it would be a waste to throw out
+//!       this minimally intrusive machinery.
+//!
+//! * [SymbolSupplier][] - maps a [Module][] to a [SymbolFile][]
+//!     * minidump-processor does not directly use this, it's just there so the Symbolizer can
+//!       generically handle different symbol fetching strategies (which minidump-processor
+//!       selects and configures).
+//!
+//!
+//!
+//! While minidump-processor provides implementations of these traits:
+//!
+//! * [FrameSymbolizer][] - callbacks that symbolication uses to return its results.
+//!     * Implemented by [StackFrame][crate::process_state::StackFrame]
+//!     * Implemented by DummyFrame (private, for a stack scanning heuristic)
+//! * [FrameWalker][] - callbacks that cfi eval uses to read callee state and write caller state.
+//!     * Implemented by CfiStackWalker (private)
+//!
+//!
+//!
+//! The symbolizer is responsible for providing the following concrete functions, which
+//! minidump-processor uses to select and configure the symbol fetching strategy:
+//!
+//! * [http_symbol_supplier][] - a [SymbolSupplier][] that can find symbols over HTTP (and cache).
+//! * [simple_symbol_supplier][] - a [SymbolSupplier][] that can find symbols on disk.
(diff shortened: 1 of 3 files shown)


minidump-processor/src/symbols.rs
```diff
@@ -1,3 +1,98 @@
...
+//! * [SymbolSupplier][] - maps a [Module][] to a [SymbolFile][]
+//!     * minidump-processor does not directly use this, it's just there so the Symbolizer can
+//!       generically handle different symbol fetching strategies (which minidump-processor
+//!       selects and configures).
+//!
+//!
+//!
+//! While minidump-processor provides implementations of these traits:
+//!
+//! * [FrameSymbolizer][] - callbacks that symbolication uses to return its results.
```
luser, line 29: I implemented this this way with an eye towards interop, not forcing callers of the symbolication APIs to choose between exposing `StackFrame` in their public API or writing code to transform a `StackFrame` into their own data type. I guess this isn't a huge improvement over that, but decoupling things slightly felt sensible. IDK if any of the Sentry crates have a better approach, or a trait we could put somewhere that'd be amenable to sharing?

minidump-processor/src/symbols.rs
```diff
@@ -1,3 +1,98 @@
...
+//!
+//! * [SymbolProvider][] - provides symbolication, cfi evaluation, and debug statistics
+//!     * Implemented by [Symbolizer][]
+//!     * This actually doesn't need to be a trait in the current design, it exists to allow
+//!       multiple symbolicators to be used together, via [MultiSymbolProvider][]. The other
+//!       SymbolProviders have been removed, but I figured it would be a waste to throw out
+//!       this minimally intrusive machinery.
+//!
+//! * [SymbolSupplier][] - maps a [Module][] to a [SymbolFile][]
+//!     * minidump-processor does not directly use this, it's just there so the Symbolizer can
```
luser, line 21: I think separating out "fetch the data necessary to symbolicate this module" from "use this symbol data to symbolicate this frame" is a useful distinction to maintain. @swatinem spelled out a strawman API proposal on Discord for what they might want from this crate which does maintain the distinction, I'll paste it here for context:

```rust
trait Module {
    type FrameIter: Iterator<Item = Frame>;

    /// Resolve the caller with CFI information
    fn walk_with_cfi(cpu_context: &CpuContext, stack_memory: Memory) -> CpuContext;

    /// Symbolicate Frames and Inlinees for an instruction address
    fn symbolicate_frame(cpu_context: CpuContext) -> FrameIter;
}

trait SymbolProvider {
    type M: Module;

    /// Resolves and loads a [`Module`] based on its [`ModuleId`]
    async fn resolve_module(module: &ModuleId) -> M;
}

struct Thread {
    initial_context: CpuContext,
    stacktrace: Vec<Frame>,
}

async fn minidump_process<P>(minidump: &Minidump, symbol_provider: P) -> Vec<Thread>
where
    P: SymbolProvider,
{}
```
I don't want to bikeshed but I do think that we should choose names that are not confusables for data structures within minidumps. :) The names already in the crate aren't great, I mostly cribbed them from Breakpad.

minidump-processor/src/symbols.rs
```diff
@@ -327,4 +444,31 @@ mod symbols_shim {
         /// If we tried to parse the symbols, but failed.
         pub corrupt_symbols: bool,
     }
+
+    // These suppliers are entriely private to the implementation, so do whatever you
+    // want with them.
+
+    struct HttpSymbolSupplier {}
```
luser, line 451: I think the lib part of `minidump-processor`—the `process_minidump` API—should stay generic over the traits, but I don't think it should necessarily be a goal to make `minidump-stackwalk` itself able to plug in alternate implementations of those traits, do you? Its commandline arguments are pretty tightly bound to the current implementation strategy, and anyone (such as Sentry) who's going to use this crate with alternate implementations is assuredly going to be using the crate as a library and not invoking `minidump-stackwalk` directly.

We can just document that the `minidump-stackwalk` binary here is tailored to Mozilla's needs, and so it intentionally only supports fetching Breakpad-format symbol files from Tecken and caching them on local disk. (Specific implementation details subject to change in the future if Mozilla's needs change, naturally.)

Gankra: Yeah this is just like this to allow someone to prototype a symbolic-based implementation with minimal refactoring. Changing the interface is totally fine. This just lets them not have to worry about it if they don't want to.

Gankra: So to be clear, in the long-term I would just remove the feature and make symbolic-syms the only impl.

luser: 100% on board with that plan.

minidump-processor/src/symbols.rs
Gankra: To me the primary purpose of these traits is to avoid a circular dependency between breakpad-symbols and minidump-processor. breakpad-symbols defines the traits so that it doesn't need to know about the concrete types like StackFrame at all.

Although it also is just genuinely being used to allow me to have DummyFrame in the stack scanner (although arguably we should just have a has_function_name method on SymbolProvider or whatever, but then I'd need to write that out like 5 times blech).

Issue #126, opened 2020-12-22
Writer for Linux


msirringhaus:
For [this Firefox issue](https://bugzilla.mozilla.org/show_bug.cgi?id=1620993) I started implementing a MD writer for Linux. It can be found here: https://github.com/msirringhaus/minidump_writer_linux
It's still very rough around far too many edges, but the main functionality seems to work ok already. Also, it\s not designed yet for other platforms.

I have not used this crates definitions in `minidump-common/format.rs` for two reasons:
1. It was said that adding more crates to the Firefox tree was difficult from a review POV. But as it turns out, this crate probably could be added with minor effort. Ideally, its dependencies would need to be aligned with the versions vendored in Firefox (updating them in Firefox involves much testing-work to verify nothing else breaks). 
2. The definitions in `minidump-common/` do not work for writing to `[u8]` yet. For this they would either need to be defined as `repr(C, packed)`, which may lead to undefined behavior (which is what I'm doing at  the moment with my defintions). Or `IntoCtx` from `scroll` would need to be implemented for all structs. As far as I can tell, there is no automated possibility yet, so this would have to be done either by hand or via some derive-macro, yet to be invented.

I'd like to discuss, if and how we could integrate those two repos. If you are interested, we might even think about merging them into one.

Gankra:
Ok now that I'm done slapping this crate into shape I can look at this older issue. Especially relevant since I'm hoping to start up on the macos version of this.

Issue 1 isn't a big deal -- rust-minidump is now [part of our toolchains](https://searchfox.org/mozilla-central/source/taskcluster/ci/fetch/toolchains.yml#444) so we're already trusting the code. This would make it part of the rust vendor, which should be fine. Although I would probably pull out format.rs into a more dedicated crate (minidump-sys?) -- minidump_common has some random stuff that doesn't really matter.

Issue 2 is trivially solved by deriving Pwrite on every type we currently implement Pread for (I tested the data roundtrips right, at least for SYSTEM_INFO). Yes that's confusing because it's actually implementing IntoCtx, but that's how their derives are named. I had to remove one `repr(packed)` from a type but since we read with Pread I don't think it was actually doing anything? *Maybe* making it easier for Rust to compile to a memcopy (doubtful).

PR #335, opened 2021-12-06
WIP rework minidump-common a bit

Gankra:
I'm thinking about what should be done for #126, which would introduce more crates that
care about minidump-common and make it more of a 'product'. Most importantly we need to
derive Pwrite so a minidump client can use those type definitions to write.

I pulled out the error types to separate modules to make it easier to maintain the core
format.rs file. It was getting a bit unwieldy. For now I pub use the error modules back
into format.rs to keep things working the same, but I think longterm they probably want
to be totally separate and maybe even possible to disable with a feature?

traits.rs also really has no business existing in this context. We could also hide it
being a feature or split minidump-common into two crates, format.rs and traits.rs.
Possibly the former would be justifiably called 'minidump-sys', although I think that's
technically an abuse of naming convention since we don't link with a native minidump lib.


Gankra:
Oh whoops I accidentally committed my change to the example to test writing works right.

Gankra:
cc @gabrielesvelto since errors are "your" component.

Gankra:
NB I am hoping to use this for https://bugzilla.mozilla.org/show_bug.cgi?id=1620991, but maybe it's a better use of time to just copy whatever https://github.com/msirringhaus/minidump_writer_linux does?

gabrielesvelto:
Breaking out the error/exception codes is a good idea, makes the core stuff more manageable. That's especially true considering that those files will change very, very rarely (like once every 1-2 years when Microsoft updates its public headers). This has another benefit for me (and possibly others too): probably due to a bug or limitation in the rust analyzer parsing format.rs would OOM it (on a machine with 32 GiB of memory). It started happening when I first introduced the list of Windows error, so having it in a separate place most certainly helps.

luser:
More generally, `format.rs` being one gigantic file dates back to the time where I was using `bindgen` to convert the struct definitions from Breakpad's headers into Rust as a starting point. They're now hand-authored, so reorganizing them is very reasonable.

Gankra:
(letting this bitrot until ~january to minimize major changes while we're deploying this to socorro)

Gankra:
this has bitrotted to all hell

PR #338, opened 2021-12-08
Make --human the default output of minidump-stackwalk 

### LINKED ISSUE #337: Swap the default output format for minidump-stackwalk
Copying a comment from Matrix:

----

there are 3 binaries which call themselves "minidump-stackwalk"

 1.  breakpad's
 2.  mozilla's
 3.  rust-minidump's

1 produces human-readable output, 2 produces json output, and 3 was designed to replace 2 but supports 1's output as well. But since 2 was the primary focus, the default mode is 2's mode, not 1's.

Socorro runs 2, mozcrash.py runs 1, so replacing 1 with 3 required a command line change
(there is also now a --cyborg mode to get both, but that's not downstreamed to mc yet)

----

When I [deployed minidump-stackwalk for the mozcrash.py usecase](https://bugzilla.mozilla.org/show_bug.cgi?id=1741205), this created an unfortunate desync for local developers -- mozcrash.py expects mdsw 3 which requires a --human flag, but local users who didn't run `./mach bootstrap` are still running mdsw 1.

@willkg has said changing socorro isn't an issue, so swapping the default to --human (which is admittedly probably a better default for most people) is probably a good idea to improve compatibility. Conveniently I did actually already add a "dummy" --json flag, so the code changes are fairly trivial.

This *will* however break old versions of socc-pair because it does not pass --json, but that's a pretty minor concern (and my own tool to break).

Also I just had the drive-by thought that there's probably a cleaner system in here where instead of --cyborg and --output-file you just pass --json=path --human=path, but then how exactly you specify something goes to stdout has some ergonomics/parsing ambiguity conflicts.

Gankra:
There are 3 different binaries which call themselves minidump-stackwalk

1. breakpad's minidump-stackwalk (human output)
2. mozilla's minidump-stackwalk (json output)
3. rust-minidump's minidump-stackwalk (both)

This implementation (mdsw 3) was primarily designed to replace mdsw 2, so
default to json output was "natural". However mdsw 1 is actually run locally
and in the wild more, so emulating their interface is probably the
better default, even if the json output is the "first-class" output.

This will be part of the 0.9.6 release, and is a fairly major breaking change,
but also fairly trivial to handle downstream -- just pass the --json flag.

Note that minidump-stackwalk 0.9.5 introduced --json even while it was the
default, so you can more easily handle this transition by always specifying
--json or --human instead of relying on the default.

fixes #337

### DIFF
--- a/minidump-stackwalk/src/main.rs
+++ b/minidump-stackwalk/src/main.rs
@@ -34,7 +34,7 @@ fn main() {
         .arg(
             Arg::with_name("json")
                 .long("json")
-                .long_help("Emit a machine-readable JSON report (the default).
+                .long_help("Emit a machine-readable JSON report.
 
 The schema for this output is officially documented here:
 https://github.com/luser/rust-minidump/blob/master/minidump-processor/json-schema.md\n\n\n")
@@ -42,10 +42,10 @@ https://github.com/luser/rust-minidump/blob/master/minidump-processor/json-schem
         .arg(
             Arg::with_name("human")
                 .long("human")
-                .long_help("Emit a human-readable report.
+                .long_help("Emit a human-readable report (the default).
 
 The human-readable report does not have a specified format, and may not have as \
-many details as the default JSON format. It is intended for quickly inspecting \
+many details as the JSON format. It is intended for quickly inspecting \
 a crash or debugging rust-minidump itself.\n\n\n")
         )
         .arg(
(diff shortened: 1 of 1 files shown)


luser:
I am quite sure there isn't anyone relying on the output of *this* mdsw tool currently, so you should feel free to change it to whatever is most convenient.

luser (approved):

PR #339, opened 2021-12-08
fixup new lints

Gankra:


Gankra:
ugghhh the clippy lint is contradictory between versions

PR #340, opened 2021-12-08
MORE DOCS MORE DOCS MORE DOCS

Gankra:


Gankra:
You can preview the new rendered CLI docs here: https://github.com/Gankra/rust-minidump/tree/cleanup/minidump-stackwalk#minidump-stackwalk-cli-manual

minidump-processor/json-schema.md
```diff
@@ -400,13 +429,26 @@ these notes will assume that a crash happened by default, since that's the impor
 
   // Linux Standard Base information (Linux-specific extended system_info)
   //
-  // TODO: properly document the semantics of these values.
+  // All of these are raw dumps of specific keys in `/etc/lsb-release`.
```
gabrielesvelto, line 432: `/etc/lsb-release` is non-standard (in spite of its name) and should be considered legacy but we still use it if we find it. The minidump writer we use in Firefox can also us [`/etc/os-release`](https://www.freedesktop.org/software/systemd/man/os-release.html) instead, but it picks it only if the former is missing. In the long run we'll prefer `/etc/os-release` because that's standardized, but we can document it only when we cross that particular bridge.

PR #343, opened 2021-12-09
you can never stop me from writing more docs

Gankra:
Although this is mostly copying/moving examples and docs from other locations, including socc-pair.
A lot of corrections were made to old information in READMEs though.


minidump-stackwalk/README.md
```diff
@@ -55,11 +63,105 @@ If you're trying to analyze firefox minidumps, you'll want to point minidump-sta
 > minidump-stackwalk --symbols-url=https://symbols.mozilla.org/ /path/to/minidump.dmp
 ```
 
+Alternatively, if you want to locally reprocess a crash report on https://crash-stats.mozilla.org (socorro), you may want to use [socc-pair](https://github.com/Gankra/socc-pair), which automates this process (and diffs the local result with the one that server had).
```
Gankra, line 66: I feel *a little* weird linking to socc-pair from here, the section is specifically mozilla-specific, so it seems reasonable to call out the tool that does specifically this job?

PR #349, opened 2021-12-09
Unloaded modules fixes

### LINKED ISSUE #334: RangeMap isn't the right abstraction for UnloadedModules
I either didn't care or forgot about it when I filled in the implementation by copy-pasting the Modules stream.

UnloadedModules can necessarily overlap by address, so they need to be treated more like MultiSets rather than Sets.

This suggests changes to `by_addr` and `module_at_address`. by_addr can be changed in the implementation, but module_at_address needs a signature change to express the fact that it's a multi-set. Maybe return an iterator?

### LINKED ISSUE #348: Pretend unloaded modules are loaded in final output
Found while looking into #168 -- we need to at least temporarily reproduce a hack where breakpad would pretend unloaded modules were loaded for the purposes of `frame.module` and `frame.module_offset`.

Gankra:
* `UnloadedModulesList::module_at_address` is now `modules_at_address`, and returns an Iterator
* `UnloadedModules::by_addr` now does not deduplicate or discard overlaps (as is correct)
* `process_state::StackFrame` now includes an `unloaded_modules` field that contains all
  the overlapping unloaded modules
* json output will now pretend the first overlapping unloaded module is loaded for the
  purposes of module name and module offset
  * This will probably be deprecated in favour of a more principled approach later,
    but it accurately reproduces current behaviour.
* human output will now list off all unloaded module overlaps if there is no loaded module match:

```
Thread 34  (crashed)
 0  0x1b86d150 (unloaded asepkcs.dll@0x0x6d150)
```

Fixes #334
Fixes #348


Gankra:
Flaws with this PR that I think are "fine" temporarily, but I can fix if I get literally any pushback to motivate me to do it:

* Arguably the modules_by_addr array should still dedupe entries with the exact same name+offset
* We now linearly scan unloaded_modules for intersections with ever single stack frame
* And then copy the entire unloaded module entry into a Vec in each frame
* Even for frames where we have the module, and therefore won't use the unloaded results
* Grouping the results by name and dedupping offsets is done in the --human printer, when the --json output (and users of the actual library) will want that too

Gankra:
Oh also I didn't write any tests. Arguably I'd want to use `insta` snapshotting, but I don't have a "clean" minidump I can check in that demonstrates this, and setting up a synth minidump for this sounds exhausting.

Gankra:
second commit:

* now only check unloaded_modules if we fail to find a module (this will eliminate like 99.99% of accesses)
* hoisted the --human logic out to the general processor
* used BTreeMap/BTreeSet to make the output more stable (things show up in a specific order between runs)
  * Important to keep the JSON hack deterministic, since it just picks "the first value".
  * also realized the inner Set was necessary to dedupe the offsets anyway, doh!
* still cloning strings unconditionally for Entry but eh whatever, stars have to really align for it to even be suboptimal

Gankra:
Added tests (and did some testing infra cleanup)

Gankra:
NB: I could use cyborg output for the unloaded module test to avoid creating+writing the same file twice, but `insta` really works best when you have one snapshot per test (otherwise any time you change output you would have to iteratively rerun+accept+rerun+accept+rerun the tests to apply all the snapshot changes).

Gankra:
If we decide that writing the minidump twice is an issue, we could tweak it so that we can set the generator to either create the file or read a cached version from testdata. This could result in some confusing situations for someone unknowingly editing a cached test though.

Gankra:
realized I wasn't checking that offset deduping was working. Adding it as another commit nicely demonstrates that it is :)

minidump-processor/src/process_state.rs
```diff
@@ -362,7 +375,27 @@ impl CallStack {
...
+                    let mut first = true;
+                    for offset in offsets {
+                        if first {
+                            write!(f, "0x{:#x}", offset)?;
+                        } else {
+                            // `|` is our separator for multiple entries
+                            write!(f, "|0x{:#x}", offset)?;
+                        }
+                        first = false;
+                    }
```
gabrielesvelto, line 396: nit: You could use `map()` and `join()` and then write!() the whole string in one go (I'm assuming that `map()` does things in the order you'd expect).

gabrielesvelto (approved):
With the change you did from Vec<> to BTreeMap<> this is fine!

minidump-processor/src/process_state.rs
Gankra: Ah yeah, that's the API I was trying to remember. I think I'm fine leaving it like this for now.

PR #356, opened 2021-12-13
Start top-level errors with an error name.

### LINKED ISSUE #354: use error codes in stderr for processing errors
When minidump-stackwalk processes [ca472b14-67b6-4e00-a272-52f7c0211208])http://localhost:8000/report/index/ca472b14-67b6-4e00-a272-52f7c0211208?refresh=cache), it doesn't produce any JSON output, so there's no "status" field. Turns out Socorro's crash signature generation looks at that field for error codes and adds them to the signature. Fun times.

Instead of putting that data in the "status" field, minidump-stackwalk chose to spit it out to stderr. I agree that makes more sense.

However, now signatures have changed for error cases. I can adjust Socorro to look at stderr for phrases that indicate the various error situations, but the indicator text is an english sentence and someone could change that down the line breaking the Socorro processor code.

For example, minidump-stackwalk spits this out:

```
[ERROR] Error processing dump: The thread list stream was not found
```

I could look for "thread list stream was not found", but if anyone changes any of those words, the match fails and we'd never know about it.

I'd like to add error codes. So minidump-stackwalk spits out whatever, but also spits out `ERROR_NO_THREAD_LIST`. Maybe something like this:

```
[ERROR] Error processing dump: ERROR_NO_THREAD_LIST: The thread list stream was not found
```

I think it's sufficient to add error codes to this block here:

https://github.com/luser/rust-minidump/blob/6ccdead2bffb0c9db60037b1ea6923c69afba023/minidump-processor/src/processor.rs#L29-L38

Does that seem ok? Sometimes having error codes makes documentation easier. Would it help here?

Gankra:
fixes #354

### DIFF
--- a/minidump-stackwalk/src/main.rs
+++ b/minidump-stackwalk/src/main.rs
@@ -295,7 +295,10 @@ fn main() {
                     .copied()
                     .unwrap_or("<cause unknown>")
             });
-        error!("A panic occurred at {}:{}: {}", filename, line, cause);
+        error!(
+            "Panic - A panic occurred at {}:{}: {}",
+            filename, line, cause
+        );
     }));
 
     let mut options = ProcessorOptions::default();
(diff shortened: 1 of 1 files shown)


Gankra:
Examples:

```
[ERROR] HeaderMismatch - Error reading dump: Header mismatch
[ERROR] MissingThreadList - Error processing dump: The thread list stream was not found
[ERROR] Panic - A panic occurred at minidump-stackwalk\src\main.rs:305: oh no a panic message!!
```

Gankra:
Oh good the unloaded modules testing that I thought was a bit sketchy is indeed racy on the filesystem.

willkg:
I like this a lot! It's better than what I was thinking of doing. r+

PR #357, opened 2021-12-13
Use reqwest's Bytes type instead of copying into a Vec

Gankra:
50% chance this elides an enormous copy and a bunch of reallocations, but no idea.
The async branch needs this anyway so no harm to do it now.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -87,6 +87,7 @@ checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
 name = "breakpad-symbols"
 version = "0.9.6"
 dependencies = [
+ "bytes",
  "failure",
  "log",
  "minidump-common",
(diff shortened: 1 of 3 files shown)


flub (approved):

Gankra:
I skimmed the code a bit more and it seems like this *might* be just as bad? Depending on if internally `hyper` is allocating many buffers, because it seems `bytes` is always one concatenated buffer.

All of the code downstream from this point can operate on arbitrary Read impls, so in theory we can just pass it down. However there's two problems:

* async reqwest doesn't provide Read, so unclear if that will port (I think you request a stream..?)
* Read mutates the source (so you can only Read something once), and we need to Read twice -- once to write to the cache, and once to parse. I see no obvious way to get a "shallow" clone of the Response to get two Reads (and perhaps this cannot be supported by the implementation at all?)

Gankra:
Ok I'm reasonably convinced this will at least be more likely to avoid lots of reallocations of enormous buffers.

Gankra:
image

😎

PR #361, opened 2021-12-15
Don't trust high bits of crash addresses

### LINKED ISSUE #360: crash address being read as 64-bit for 32-bit platforms
socc-pair --crash-id=[082a152f-05b4-4de9-943b-c30550211210](https://crash-stats.mozilla.org/report/index/082a152f-05b4-4de9-943b-c30550211210)

```diff
 : {
   crash_info: {
-    did not match
+    address: 0xe5e5e5e5
-    address: 0xffffffffe5e5e5e5
     crashing_thread: 25
     type: EXCEPTION_ACCESS_VIOLATION_READ
   }
```

(+ is breakpad, - is rust-minidump)

This crash report is from x86, so it's almost certain that rust-minidump is wrong here.

Gankra:
This is a minor breaking change, as get_crash_address now requires a Cpu.

Also adds some minimal Exception support to synth-minidump.

Fixes #360

### DIFF
--- a/breakpad-symbols/src/sym_file/parser.rs
+++ b/breakpad-symbols/src/sym_file/parser.rs
@@ -62,7 +62,7 @@ named!(module_line<&[u8], ()>,
     || {}
 ));
 
-// Matches an INFO LINE record.
+// Matches an INFO URL record.
 named!(
     info_url<&[u8], Info>,
     chain!(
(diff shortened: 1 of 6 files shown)


Gankra:
confirmed: this fixes the report in #360

PR #367, opened 2021-12-17
WIP: Add unloaded module offsets to frames in JSON output. 

### LINKED ISSUE #168: Expose full unloaded_modules results for json output
<details>
<summary>(issue rewritten, original message archived here)</summary>

As far as I can tell we only find the loaded modules. This is a problem in two places:

* The unwinder won't consider those addresses mapped and mess up its backtrace
* The symbolicator won't map symbols of unloaded modules
* ~~minidump-processor won't report those modules (part of mozilla's schema)~~

One issue with unloaded modules is that they can overlap with themselves or other loaded modules. Presumably loaded modules should be "preferred" when doing address lookups, but there isn't an obvious right answer when two unloaded modules overlap. I guess just pick one?

I haven't checked where in the minidump format this info comes from yet.
</details>

Right now when doing stackwalking and symbolication, we only use the *loaded* modules to map addresses. This is certainly a pretty reasonable default, since ideally threads *shouldn't* be executing unloaded DLLs, but maybe we're leaving valuable information on the table?

--------

The relevant places are:

* instruction_seems_valid_by_symbols

https://github.com/luser/rust-minidump/blob/10d3dfd2e3c61e7030860617c883782afde0c026/minidump-processor/src/stackwalker/mod.rs#L206-L216

* fill_source_line_info

https://github.com/luser/rust-minidump/blob/10d3dfd2e3c61e7030860617c883782afde0c026/minidump-processor/src/stackwalker/mod.rs#L139-L147

* get_caller_by_cfi (in all stackwalker backends)

https://github.com/luser/rust-minidump/blob/10d3dfd2e3c61e7030860617c883782afde0c026/minidump-processor/src/stackwalker/amd64.rs#L29-L49

-----------

That said, `unloaded_modules` is a bit weird. It's not a pristine and detailed memory mapping like `modules` is. It's a *fixed-size circular log of unload events*. This means:

* memory ranges may overlap (even with a loaded module!)
* modules may show up multiple times (and at different locations!)
* unload events may be missing due to size limits on the log (we'll only have the N newest events)
* **they don't have debugids, only module names** (so we can't query symbol servers for the right version!!)

The last point is pretty fatal to using this information for symbolication, so this information can only really potentially be used for the first level of analysis of `instruction_seems_valid_by_symbols`, and *maybe* oddball analyses like exploitability (#25). 

And of course, there's also the possibility that trying to factor in the messy unloaded_modules will make our analysis *worse* because most of the time that memory really will be just coincidentally overlapping, causing us to e.g. hallucinate more during scans.

Gankra:
Reverts the JSON hack introduced for #348 and properly handles the situation. I have intentionally avoided publishing a new version of minidump-stackwalk until this PR so that there's no published version that includes the hack.

That said, we may "prefer" the hybrid solution where "module" is still hacked but we emit unloaded modules anyway. In this case the user could "tell" the module is unloaded by the presence of unloaded_modules.

However this would create a hairy situation if we ever decide to uncondtionally emit unloaded_modules (because maybe someone finds it useful). Then you would have to check if "module" appears in unloaded_modules BUT this is imperfect because you don't know if "module" represents a loaded or unloaded version. To fully resolve this, you would need to check the top level "modules" list and validate the offset, which is a pain in the ass.

Fixes #168


Gankra:
This is a "2022" PR, so don't sweat it now.

Vacation starts tomorrow.

Gankra:
See also: https://bugzilla.mozilla.org/show_bug.cgi?id=1746630

Gankra:
Rebased. Now that I've had time to sit on this, I'm pretty happy with landing this version. It's a bit more work for someone to use, but it's the Most Correct way of expressing this information.

PR #391, opened 2022-01-12
remove old cruft

### LINKED ISSUE #351: delete the shell script?
I literally never noticed this file until just now:

https://github.com/luser/rust-minidump/blob/master/bindgen.sh

I don't think, it's actually used/useful anymore?

Gankra:
fixes #351

While it seems a bit weird to remove an example, I think it actually has been "unlinked" from the project for quite a while, since I demoted `minidump` from the "root" crate to "just another crate". So at this point it's better off deleted than still floating there.

Our READMEs have plenty of examples now, and many of them get run+tested in CI.


luser:
Agreed.

PR #404, opened 2022-01-26
fix: Guard memory reservations with bounds checks

### LINKED ISSUE #326: Do not make unbounded memory reserves based on untrusted inputs
This is apparently an issue Sentry ran into when testing this library out. We would blow up their allocator guards against enormous allocations. The specific case we discussed involved the count of streams in the directory, which was larger than the file size could even hold.

There are a few options here:

1. Do not reserve but otherwise go along with the count
   * Optionally: also iterate the list backwards so overflows will fast-fail
2. Have a (static?) `max` on the reserve size
3. Have a (static?) `max` on the counts (e.g. declare that we refuse to acknowledge the validity of a minidump with 1000+ streams)
4. Reserve, but check that the count is physically possible (in terms of the input buffer size and size of the entries)

I am personally inclined to take the braindead simple approach of Option 1.

Here's the offending cases:

```
> rg "with_capacity|reserve\(|reserve_exact\("
minidump-common\src\traits.rs
79:        let mut vec: Vec<(Range<u64>, V)> = Vec::with_capacity(input.len());

minidump\src\minidump.rs
1122:    let mut raw_entries = Vec::with_capacity(count);
1194:    let mut raw_entries = Vec::with_capacity(number_of_entries as usize);
1212:        let mut names = HashMap::with_capacity(raw_names.len());
1350:        let mut modules = Vec::with_capacity(raw_modules.len());
1443:        let mut modules = Vec::with_capacity(raw_modules.len());
1620:        let mut regions = Vec::with_capacity(descriptors.len());
2315:        let mut threads = Vec::with_capacity(raw_threads.len());
2316:        let mut thread_ids = HashMap::with_capacity(raw_threads.len());
3902:    let mut strings = Vec::with_capacity(count as usize);
4040:    let mut module_links = Vec::with_capacity(count as usize);
4218:        let mut streams = HashMap::with_capacity(header.stream_count as usize);
```

Those that key off `len()` are fine because they're all based on a collection we've already constructed.

Filtering that down we get:

```
minidump\src\minidump.rs
1122:    let mut raw_entries = Vec::with_capacity(count);
1194:    let mut raw_entries = Vec::with_capacity(number_of_entries as usize);
3902:    let mut strings = Vec::with_capacity(count as usize);
4040:    let mut module_links = Vec::with_capacity(count as usize);
4218:        let mut streams = HashMap::with_capacity(header.stream_count as usize);
```

### LINKED ISSUE #381: Malformed crashdump causes OOM by allocating far too much memory
On the base64 decoded data of `TURNUJOnAAAA/2ZmZFlmZmZmZkAKCmZwCrv///8K/wo=`, it OOMs when parsing through the `minidump::Minidump::read(data)` method, and when running `minidump-stackwalk` (version 0.9.6), I get

```
# minidump-stackwalk crash.txt          
memory allocation of 45097156624 bytes failed
zsh: abort (core dumped)  minidump-stackwalk crash.txt
```

Swatinem:
This should fix #381 and fix #326

### DIFF
--- a/minidump/src/minidump.rs
+++ b/minidump/src/minidump.rs
@@ -682,6 +682,29 @@ fn read_codeview(
     })
 }
 
+/// Checks that the buffer is large enough for the given number of items.
+///
+/// Essentially ensures that `buf.len() >= offset + (number_of_entries * size_of_entry)`.
+/// Returns `(number_of_entries, expected_size)` on success.
+fn ensure_count_in_bound(
+    buf: &[u8],
+    number_of_entries: usize,
+    size_of_entry: usize,
+    offset: usize,
+) -> Result<(usize, usize), Error> {
+    let expected_size = number_of_entries
+        .checked_mul(size_of_entry)
+        .and_then(|v| v.checked_add(offset))
+        .ok_or(Error::StreamReadFailure)?;
+    if buf.len() < expected_size {
+        return Err(Error::StreamSizeMismatch {
+            expected: expected_size,
+            actual: buf.len(),
+        });
+    }
+    Ok((number_of_entries, expected_size))
+}
+
 impl MinidumpModule {
     /// Create a `MinidumpModule` with some basic info.
     ///
@@ -1091,20 +1114,14 @@ where
     let u: u32 = bytes
         .gread_with(offset, endian)
         .or(Err(Error::StreamReadFailure))?;
-    let count = u as usize;
-    let counted_size = match count
-        .checked_mul(<T>::size_with(&endian))
-        .and_then(|v| v.checked_add(mem::size_of::<u32>()))
-    {
-        Some(s) => s,
-        None => return Err(Error::StreamReadFailure),
-    };
-    if bytes.len() < counted_size {
-        return Err(Error::StreamSizeMismatch {
-            expected: counted_size,
-            actual: bytes.len(),
-        });
-    }
+
+    let (count, counted_size) = ensure_count_in_bound(
+        bytes,
+        u as usize,
+        <T>::size_with(&endian),
+        mem::size_of::<u32>(),
+    )?;
+
     match bytes.len() - counted_size {
         0 => {}
         4 => {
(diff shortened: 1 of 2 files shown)


minidump/src/minidump.rs
```diff
@@ -3946,8 +3946,21 @@ fn read_string_list(
     let count: u32 = data
         .gread_with(&mut offset, endian)
         .or(Err(Error::StreamReadFailure))?;
+    let count = count as usize;
```
flub, line 3949: since you bother doing a `.checked_mul()` it seems weird to not use `TryInto` here instead of the cast

Swatinem: I am casting from `u32` to `usize`. I don’t know if Rust even supports any platform for which that is fallible.

minidump/src/minidump.rs
```diff
@@ -3946,8 +3946,21 @@ fn read_string_list(
...
+    // NOTE: this only counts the RVA size, not the raw variable-size string data,
+    // but that should be good enough for the bounds check
+    let counted_size = match count.checked_mul(<md::RVA>::size_with(&endian)) {
+        Some(s) => s,
+        None => return Err(Error::StreamReadFailure),
```
luser, line 3954: You might alternately write this as:

```rust
let counted_size = count.checked_mul(<md::RVA>::size_with(&endian)).ok_or(Error::StreamReadFailure)?;
```

luser (approved):
This looks good, and thanks for adding those tests! [I had floated the idea of wrapping this logic in a generic method previously](https://github.com/luser/rust-minidump/issues/326#issuecomment-984030156). With your changes here + the existing logic in `read_stream_list`, that makes 4 places using the same logic, so I do think it's justifiable to factor out a single implementation like that, but it's also not terribly complex logic so I wouldn't reject this PR if you disagree or don't want to do that work.

minidump/src/minidump.rs
Swatinem: Oh, absolutely!

I will try making this a more generic util along with that improvement.

Swatinem:
I extracted the code into its own helper. Please take another look.

luser (approved):
Looks good, thanks!

minidump/src/minidump.rs
Gankra: rust has community support for a few 16-bit microcontrollery things but realistically we don't need to worry about compat with that kind of platform.

Gankra:
rad

PR #405, opened 2022-01-26
Add fuzzer to minidump

5225225:
All this does is tries to decode a minidump file and extract every known stream from it.

This is probably not going to see *all* the surface area (seeing as i'm not doing anything with the streams), but it should find *enough* for now. The fuzzer can always be improved in the future to cover more stuff.

This is finding multiple bugs, that I'll add fixes for in future PRs (One per bug, to make it easier both for me and easier to keep track of them).

Then once that's all done and the shallow bugs are done, might be worth getting this crate to be fuzzed as part of oss-fuzz.

PR #406, opened 2022-01-26
Check for invalid record count in mac stream

5225225:
Found with #405 (But I rebased this to not include that commit so it doesn't matter which one of the two is merged first).


5225225:
Hmm... Do we want to be calling dumps with an incorrect number of records *invalid*? Or should we do `header.records.iter().take(header.record_count)`, perhaps?

Both is better than crashing, but up to whoever looks at this what I should do. Both work.

luser:
> Hmm... Do we want to be calling dumps with an incorrect number of records _invalid_? Or should we do `header.records.iter().take(header.record_count)`, perhaps?

It's tricky, for sure. This crate has historically erred on the side of "get as much data as possible out of malformed minidumps", because our experience dealing with Firefox minidumps from end-users showed that we'd get dumps that were broken in all sorts of fascinating ways and getting *something* out of them was better than nothing. (Especially since Firefox exposes crash reports to enthusiastic end-users via `about:crashes`. There's nothing more demoralizing than having your browser crash, going to look up your crash report, and finding that it has no useful information because the crash broke things in a very specific way.)

5225225:
Okay, I'll take the first header.record_count

Actually, why is there a record count anyways? Surely we already parsed some number of records and put them in `header.records`?

5225225:
Ahhh, `records` is a fixed length array of records, and the ones past the record count is uninit (Well, presumably Default::default() because it's not wrapped in MaybeUninit, but *semantically* uninit).

Makes sense, avoids allocation. So if the record count is over 20, everything gets returned.

minidump/src/minidump.rs
```diff
@@ -2864,7 +2864,10 @@ impl<'a> MinidumpStream<'a> for MinidumpMacCrashInfo {
         let strings_offset = header.record_start_size as usize;
         let mut prev_version = None;
         let mut infos = Vec::new();
-        for record_location in &header.records[..header.record_count as usize] {
+
+        let records = header.records.iter().take(header.record_count as usize);
```
Gankra, line 2868: Can you add a comment here noting this is a concern? Something to the effect of:

```
// We use `take` here to better handle a corrupt record_count that is larger than the maximum supported size
```

Gankra (approved):

minidump/src/minidump.rs
5225225: Yep, that's done!

5225225:
Rebased ontop of `master` to resolve the merge conflict. Other than that, left it untouched.

Gankra:
thank you so much!! https://twitter.com/Gankra_/status/1488254127633227780

5225225:
awww that's made my day <3

Yep, I've not found any security vulns (other than DoS, i mean actual UB) in rust code through fuzzing

But I find a *absolute ton* of logic bugs and unexpected panics

The spiciest issues I tend to see are infinite loops in parsers and stack overflows (which kill a whole process, and catch_unwind will not save you `>:)`)

Gankra:
I like to think I've been *reasonably* good at enumerating and testing corner cases (at least you're hitting assertions!) but yeah it's extremely hard to think of them all when you are literally writing "the thing that runs when a program has encountered an unrecoverable corner case" and at some point you get tired of coming up with contrived inputs that "surely will never happen in practice anyway, why am I even bothering".

Nice to have a machine look for them instead! 😸

PR #409, opened 2022-01-27
ref: Replace failure by thiserror, making Errors std compatible

### LINKED ISSUE #371: Errors do not implement std::error::Error
Several errors implement `failure::Fail` right now, which comes with the added benefit of having a backtrace on stable. How much do you rely on backtraces in your use cases at the moment? I would like to create a PR swapping this all out with `std::error::Error`.

@Gankra @luser

Swatinem:
fixes #371

This is technically a breaking change, as it removes stuff from existing Error types.

### DIFF
--- a/breakpad-symbols/src/lib.rs
+++ b/breakpad-symbols/src/lib.rs
@@ -209,8 +209,8 @@ pub enum SymbolError {
     /// (e.g. a bad STACK WIN line can be discarded without affecting anything
     /// else). But sometimes we can't make any sense of the symbol file, and
     /// you find yourself here.
-    #[error("parse error: {0}")]
-    ParseError(String),
+    #[error("parse error: {0} at line {1}")]
+    ParseError(&'static str, u64),
 }
 
 /// An error produced by fill_symbol.
@@ -236,7 +236,7 @@ impl PartialEq for SymbolError {
             (self, other),
             (SymbolError::NotFound, SymbolError::NotFound)
                 | (SymbolError::LoadError(_), SymbolError::LoadError(_))
-                | (SymbolError::ParseError(_), SymbolError::ParseError(_))
+                | (SymbolError::ParseError(..), SymbolError::ParseError(..))
         )
     }
 }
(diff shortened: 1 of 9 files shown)


breakpad-symbols/src/sym_file/mod.rs
```diff
@@ -62,9 +60,9 @@ impl SymbolFile {
         let mut fully_consumed = false;
         loop {
             // Read the data in, and tell the circular buffer about the new data
-            let size = input_reader.read(buf.space()).map_err(|e| {
-                SymbolError::LoadError(format_err!("couldn't read input stream {}", e))
-            })?;
+            let size = input_reader
+                .read(buf.space())
+                .map_err(SymbolError::LoadError)?;
```
luser, line 65: Doesn't the `#[from]` annotation you included in the declaration of this variant generate a `From<io::Error>` impl, and thus make this work without `map_err`?

breakpad-symbols/src/lib.rs
```diff
@@ -194,22 +192,25 @@ pub fn relative_symbol_path(module: &dyn Module, extension: &str) -> Option<Stri
...
     /// Symbol file was too corrupt to be parsed at all.
     ///
     /// Because symbol files are pretty modular, many corruptions/ambiguities
     /// can be either repaired or discarded at a fairly granular level
     /// (e.g. a bad STACK WIN line can be discarded without affecting anything
     /// else). But sometimes we can't make any sense of the symbol file, and
     /// you find yourself here.
-    ParseError(Error),
+    #[error("parse error: {0}")]
+    ParseError(String),
```
luser, line 213: Pretty minor, but given the errors this code is generating in practice, seems like this could be:

```
#[error("parse error: {0} at line {1}")]
ParseError(&'static str, u64)
```

minidump/src/minidump.rs
```diff
@@ -666,17 +662,17 @@ fn read_codeview(
     location: &md::MINIDUMP_LOCATION_DESCRIPTOR,
     data: &[u8],
     endian: scroll::Endian,
-) -> Result<CodeView, failure::Error> {
-    let bytes = location_slice(data, location)?;
+) -> Result<CodeView, ()> {
```
luser, line 665: It might be more straightforward to make this return `Option<CodeView>` instead of using `()` as the error type. You'd be trading all the `.or(Err(()))?` for `.ok()?`.

luser (approved):

breakpad-symbols/src/lib.rs
Swatinem: Absolutely! I wasn’t aware there is always access to a line number.

PR #411, opened 2022-01-28
ref: Replace deprecated crates

Swatinem:
- Removes `tempdir` in favor of `tempfile`.
- Replaces `mmap` by `mmap2`.
- Replaces `chrono` by `SystemTime`/`time`.

All these crates have rustsec advisories, mostly because they are deprecated or unmaintained.

`chrono` is still being pulled in by `simplelog` though :-(

This is technically a breaking change, because both `mmap` and `chrono` were part of the public API.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -94,7 +94,6 @@ dependencies = [
  "nom",
  "range-map",
  "reqwest",
- "tempdir",
  "tempfile",
 ]
 
@@ -137,8 +136,7 @@ dependencies = [
  "libc",
  "num-integer",
  "num-traits 0.2.14",
- "serde",
- "time",
+ "time 0.1.43",
  "winapi",
 ]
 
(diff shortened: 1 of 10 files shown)


flub (approved):

flub:
let's wait for some other people's opinion on the public api change though :)

gabrielesvelto:
We have no issues with this in Mozilla as we're only using the tools stand-alone. +1 from me for dropping deprecated dependencies

luser:
Yeah, do it. I don't think there are really any users of this crate outside of Mozilla and Sentry currently anyway. We should hash out a few things if we want to ship a 1.0 (I believe @Gankra has an issue open with some thoughts) but it's not critical.


> This is technically a breaking change, because both `mmap` and `chrono` were part of the public API.

Oh, interesting. I guess to future-proof this we could newtype the bits of third-party crates that leak into the public API like this? (It would be neat if there was a clippy lint for "something from another crate is part of your public API"!) It's nice that we can use `SystemTime` here now! (Maybe I always could and I just didn't realize? Or maybe the bits of API I needed didn't exist when I first wrote the code?)

I do like the simple `Minidump::read_path` API. I guess the even-simpler solution here would be to box up the `Mmap` object there to hide it in the type signature?

PR #416, opened 2022-01-31
Add thread_id to CallStacks in the ProcessState

jan-auer:
The `thread_id` is part of the raw minidump thread structure, but not the processed `CallStack`. This is an inconvenience when dealing with the process state.

If preferred, we could also add the raw thread structure as field to `CallStack`.

### DIFF
--- a/minidump-processor/src/stackwalker/mod.rs
+++ b/minidump-processor/src/stackwalker/mod.rs
@@ -198,6 +198,7 @@ where
     CallStack {
         frames,
         info,
+        thread_id: 0,
         thread_name: None,
         last_error_value: None,
     }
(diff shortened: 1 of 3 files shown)


minidump-processor/src/stackwalker/mod.rs
```diff
@@ -198,6 +198,7 @@ where
     CallStack {
         frames,
         info,
+        thread_id: 0,
```
jan-auer, line 201: This is just a placeholder. It's overwritten after `stackwalker::walk_stack` in the processor.

Gankra (approved):
This has always bothered me, thank you!!

PR #417, opened 2022-01-31
Process OS build into a dedicated field

jan-auer:
The processor joins the operating system version and build into a single field.
While this is a common way to represent the version, it is convenient to have
them separate.


jan-auer:
I'm considering to split the version completely into its parts, and then adding
convenience methods to join them back together. Does someone have a preference?

Gankra (approved):
Especially appreciate you improved the docs at the same time :)

PR #425, opened 2022-02-02
Avoid overflow in bounds check of MinidumpMemory

5225225:
The fuzzer found yet another overflow.

We *could* make `MinidumpMemory` check for `self.base_address + self.size` being in bounds on construction, but I guess it's better to do it when you actually go to *use* it, since directly asking for the base address/size still works.

Still, that leads to cases like this.


Gankra (approved):
nice

PR #426, opened 2022-02-02
Add stack walk fuzzer, as well as minidump_process fuzzer

5225225:
This is draft while I clean up this PR, but especially the process fuzzer is still finding bugs, even if the reproducers it gives you are hellishly big and hard to work with. Still useful!


minidump-processor/fuzz/fuzz_targets/walk_stack.rs
```diff
@@ -0,0 +1,67 @@
...
+use minidump::format::CONTEXT_X86;
+use minidump::{MinidumpContext, MinidumpContextValidity, MinidumpMemory, MinidumpRawContext};
+use minidump::{MinidumpModule, MinidumpModuleList};
+use minidump_processor::walk_stack;
+use minidump_processor::{string_symbol_supplier, CallStack, Symbolizer};
+use std::collections::HashMap;
+use test_assembler::Section;
+
+struct TestFixture {
+    pub raw: CONTEXT_X86,
```
luser, line 13: Hm. I was going to suggest you could get a little more coverage by making this generic over the various supported CPU contexts, but I guess that's annoying. We should add `From<CONTEXT_X86>` et. al. impls for [`MinidumpRawContext`](https://github.com/luser/rust-minidump/blob/a32419975ba6f190081a148f04a41fc42aa4eb26/minidump/src/context.rs#L22) so that you could make this like:

```rust
struct TestFixture<C: Into<MinidumpRawContext>> {
    pub raw: C,
    // …
}
```

Having said that, I don't know how easy it is to write fuzz targets that test generics like that.

minidump-processor/fuzz/fuzz_targets/walk_stack.rs
```diff
@@ -0,0 +1,67 @@
...
+        Some(walk_stack(
+            &Some(&context),
+            Some(&stack_memory),
+            &self.modules,
+            &symbolizer,
+        ))
+    }
+}
+
+fuzz_target!(|data: &[u8]| {
```
luser, line 59: Do you maybe want to make this take two parameters where one of them could fill in the CPU context?

luser:
I'm impressed, this is really cool!

5225225:
This *seems* to work fine, except for my vim rust analysis complaining about `md::CONTEXT_X86` not implementing `arbitrary`. Works fine when built, but I basically never use features so I'm not entirely sure what I'm doing here. `minidump` needs to enable the `arbitrary` feature on `minidump-common`, if it is enabled for itself.

luser:
Awesome!

> minidump needs to enable the arbitrary feature on minidump-common, if it is enabled for itself.

Per [the docs](https://doc.rust-lang.org/cargo/reference/features.html#dependency-features), what you need in the `minidump` `Cargo.toml` is:

```toml
[features]
arbitrary = ["minidump-common/arbitrary"]
```

If you wanted to name the feature something like `fuzzing` instead you could do:

```toml
[features]
fuzzing = ["arbitrary", "minidump-common/fuzzing"]
```

You'd then also need a similar section for `minidump-common`:


```toml
[features]
fuzzing = ["arbitrary"]
```

Gankra (approved):
tentatively approved if you can figure out the feature stuff

5225225:
ahhhh
okay

the issue i was having is that adding a dependency with `optional = true` implicitly makes a feature that enables that dependency (As in, has `arbitrary = ["dep:arbitrary"]`)

which worked fine, except when I went to enable `minidump-common`

I had this line, but it was then complaining about 

```toml
[features]
arbitrary = ["minidump-common/arbitrary"]
```

```
error: failed to parse manifest at `rust-minidump/minidump/Cargo.toml`

Caused by:
  optional dependency `arbitrary` is not included in any feature
  Make sure that `dep:arbitrary` is included in one of features in the [features] table.
```

And I just didn't *read* the error message.

```toml
[features]
arbitrary = ["minidump-common/arbitrary", "dep:arbitrary"]
```

is correct.

I understand features a bit more now, the array on the right is the set of things (either features in dependent crates, your own crate, or dependencies that need to be activated) to enable when the feature is set.

5225225:
Okay cool.

Just adding `dep:arbitrary` works on *nightly* (without any feature flags, it was recently stabilized) but not on stable. 

This seems to work fine though.

Gankra:
Thanks for sorting it out!

PR #427, opened 2022-02-02
CFI fuzzer, fix some overflows it found

5225225:


breakpad-symbols/src/sym_file/walker.rs
```diff
@@ -737,7 +737,7 @@ enum CfiReg<'a> {
     Other(&'a str),
 }
 
-fn eval_win_expr(expr: &str, info: &StackInfoWin, walker: &mut dyn FrameWalker) -> Option<()> {
```
5225225, line 740: I'll need to make sure this wasn't made public when it shouldn't have been, I needed it for the fuzzer to call.

breakpad-symbols/src/lib.rs
```diff
@@ -52,6 +52,13 @@ pub use crate::sym_file::{CfiRules, SymbolFile};
 
 mod sym_file;
 
+// Public re-exports for the purposes of the cfi_eval fuzzer.
+#[doc(hidden)]
```
5225225, line 56: Assuming doc hidden is good enough to make this not public API. I can add a feature flag as well that the fuzzing crate sets.

luser: It might be good to hide this behind `#[cfg(feature = "fuzz")]`, yeah.

breakpad-symbols/src/sym_file/walker.rs
5225225: it was

i put it behind the same cfg(feature = fuzz) as the re-exports were.

Gankra (approved):
This is a bit more intrusive than the other fuzzers, which is, less nice.

But in the grand scheme of things it's pretty unobtrusive, and hard to argue with results.

PR #431, opened 2022-02-07
Improve how Linux signals are printed out

gabrielesvelto:
This includes a few different changes:
* If the exception records' si_code is 0 just print the signal name and omit the hexadecimal value of si_code (this was Breakpad's behavior)
* For signals that have been delivered by users (si_code <= 0) print the corresponding flag (e.g. SI_USER, SI_TKILL, etc...)
* For signals that have been sent from "somewhere" within the kernel  (si_code == 0x80) print SI_KERNEL
* Add human-readable si_code values for SIGTRAP and SIGSYS

### DIFF
--- a/minidump-common/src/format.rs
+++ b/minidump-common/src/format.rs
@@ -6519,8 +6519,22 @@ pub enum ExceptionCodeLinux {
 
 // These values come from asm-generic/siginfo.h
 #[derive(Copy, Clone, PartialEq, Debug, Primitive)]
-pub enum ExceptionCodeLinuxSigillKind {
+#[repr(i32)]
+pub enum ExceptionCodeLinuxSicode {
     SI_USER = 0,
+    SI_KERNEL = 0x80,
+    SI_QUEUE = -1i32,
+    SI_TIMER = -2i32,
+    SI_MESGQ = -3i32,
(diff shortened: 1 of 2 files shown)


minidump-common/src/format.rs
```diff
@@ -6519,8 +6519,22 @@ pub enum ExceptionCodeLinux {
 
 // These values come from asm-generic/siginfo.h
 #[derive(Copy, Clone, PartialEq, Debug, Primitive)]
-pub enum ExceptionCodeLinuxSigillKind {
+#[repr(i32)]
```
flub, line 6522: do you need this if you'r not dealing with ffi?

flub (approved):

minidump-common/src/format.rs
gabrielesvelto: It seems that the representation defaults to u32 so I can't use signed constants for the values (or maybe I'm doing it wrong?).

flub: oh, interesting.  TIL

Gankra:
God I love giant files that contain everything but we *really* need to factor out the errorcode stuff some day. I really should have finished #335...

PR #432, opened 2022-02-08
Switch to debugid crate instead of strings

flub:
The Debug ID and Code ID were represented by strings in the breakpad
format.  This switches this to the DebugId and CodeId types from the
debugid crate as also used by symbolic.

The benefit is strong types and that DebugId becomes Copy.

As a result of this some tests which used invalid Debug IDs had to be changed to use valid ones.  Also the CodeId is now normalised resulting in snapshot changes.  I went with uppercase as it seemed more similar to the current display, though lowercase would not need an extra string allocation so might be better.


minidump/src/minidump.rs
```diff
@@ -926,15 +928,21 @@ impl Module for MinidumpModule {
...
+                    &raw.signature.data4,
+                )
+                .ok()?;
+                Some(DebugId::from_parts(uuid, raw.age))
             }
             Some(CodeView::Pdb20(ref raw)) => {
-                let id = format!("{:08X}{:x}", raw.signature, raw.age);
-                Some(Cow::Owned(id))
+                let uuid = Uuid::from_fields(raw.signature, 0, 0, &[0, 0, 0, 0]).ok()?;
+                Some(DebugId::from_parts(uuid, raw.age))
```
flub, line 945: This `Pdb20` branch does not have any test coverage.  I think this is the right way to convert it but I'm not sure.  The previous generated string was also not a valid breakpad debugid because it was too short.

flub:
> Argh, sorry about all the merge conflicts. Friggin' dependencies.

My fault for sorting the dependencies...

> I'm not terribly confident in reviewing this, I really don't understand the semantics of debugid/codeid stuff. Especially once Breakpad's extensions and bugginess get involved. @luser do you have the time to spare?

Most of my confidence comes from the fact we run symbolic(ator) with this in prod, receiving lots of minidumps from breakpad etc, so its breakpad compatibility should be fine.

loewenheim (approved):

breakpad-symbols/src/lib.rs
```diff
@@ -27,16 +29,16 @@
 //!     let symbolizer = Symbolizer::new(supplier);
 //!
 //!     // Simple function name lookup with debug file, debug id, address.
-//!     assert_eq!(symbolizer.get_symbol_at_address("test_app.pdb",
-//!         "5A9832E5287241C1838ED98914E9B7FF1",
-//!         0x1010)
+//! let debug_id = DebugId::from_str("5A9832E5287241C1838ED98914E9B7FF1").unwrap();
```
Swatinem, line 32: maybe rustfmt the doc code blocks as well, like so: https://github.com/getsentry/sentry-rust/blob/master/rustfmt.toml

But that will probably introduce a ton of unrelated changes, so maybe just correctly indent this one ;-)

Swatinem (approved):

minidump/src/minidump.rs
luser: This isn't correct, and Breakpad isn't the source of truth on this matter. However, this is also an obsolete format, I think the last compiler to produce it was Visual C++ 6.0. I just shared this link (which is in the crate docs somewhere but now 404s) today, it mentions what's going on there:
http://web.archive.org/web/20210227224734/https://www.debuginfo.com/articles/debuginfomatch.html

flub: From that link I think it only says the signature is a timestamp in seconds since the epoch.  The comment at https://github.com/getsentry/breakpad/blob/63c90f00bf64a1b4a8883e50bbe8217048ed2a85/src/common/windows/pdb_source_line_writer.cc#L1217-L1229 seems to suggest that in this case the GUID conversion happens with 0-filled.  Technically this is an invalid UUID though, since the UUID variant flag (1st 3 bits of byte 33) is set to the legacy "variant 0" but then technically it is impossible to create a valid UUID out of this.

At the end of the day we're doing something to get some kind of DebugId-shaped identifier for which te symbol provider would have to do the same transformation and I think this is the same as breakpad was suggesting it might do in this case and probably good enough until/if someone ever encounters such a debugid?  So I think after learning more about this this is still a reasonable solution here.

luser: That's my point—it's *not* a UUID at all! Here, I realized that an old build of Firefox would have PDB 2.0 debug info, so if you download [the Firefox 1.0 win32 build](https://ftp.mozilla.org/pub/firefox/releases/1.0/win32/en-US/Firefox%201.0.zip), you can see what `CV_INFO_PDB20` looks like. I poked at the `firefox.exe`from there in WinDBG:

```
0:000> !sym noisy
noisy mode - symbol prompts on
0:000> .reload -f
.
DBGHELP: C:\Windows\system32\firefox.exe - file not found
DBGENG:  C:\Users\ted\Desktop\firefox.exe - Mapped image memory
SYMSRV:  BYINDEX: 0x3
         C:\ProgramData\Dbg\sym
         firefox.pdb
         418E89C31
SYMSRV:  UNC: C:\ProgramData\Dbg\sym\firefox.pdb\418E89C31\firefox.pdb - path not found
SYMSRV:  UNC: C:\ProgramData\Dbg\sym\firefox.pdb\418E89C31\firefox.pd_ - path not found
SYMSRV:  UNC: C:\ProgramData\Dbg\sym\firefox.pdb\418E89C31\file.ptr - path not found
SYMSRV:  RESULT: 0x80070003
SYMSRV:  BYINDEX: 0x4
         C:\ProgramData\Dbg\sym*https://msdl.microsoft.com/download/symbols
         firefox.pdb
         418E89C31
SYMSRV:  UNC: C:\ProgramData\Dbg\sym\firefox.pdb\418E89C31\firefox.pdb - path not found
SYMSRV:  UNC: C:\ProgramData\Dbg\sym\firefox.pdb\418E89C31\firefox.pd_ - path not found
SYMSRV:  UNC: C:\ProgramData\Dbg\sym\firefox.pdb\418E89C31\file.ptr - path not found
SYMSRV:  HTTPGET: /download/symbols/firefox.pdb/418E89C31/firefox.pdb
SYMSRV:  HttpQueryInfo: 80190194 - HTTP_STATUS_NOT_FOUND
```

You can see in the output that the debug id is `418E89C31`—that's `signature == 0x418E89C3` and `age == 1`. The current implementation should produce the same result here, but your replacement will not.

I'm not really worried about supporting these in practice, as they're obsolete, but it doesn't make sense to take code that is correct and break it.

flub: I get that this is not a UUID, to look it up though we have to somehow create a DebugId from it however and filling the remainder of the UUID bytes with zeroes is exactly the same approach as breakpad does at https://github.com/getsentry/breakpad/blob/63c90f00bf64a1b4a8883e50bbe8217048ed2a85/src/common/windows/pdb_source_line_writer.cc#L1217-L1229 which makes me think this is reasonable.  It also is simply what we can declare is done in order to map this old format, but if I understand you don't like this solution.

That leaves only two other options:
- The debugid crate, in its current form, is not suitable to represent all Debug Information File identifiers so can not be adopted.
- The minidump crate simply does not care about Pdb20 anymore and will not return DIF identifiers for them.

While it is perfectly possible to make a debugid::DebugId that has a specific format for Pdb20 I do think it the solution here (and in breakpad IIUC) to simply map it to an invalid UUID is more practical given how rare this is.  We could also compromise and make this more explicit by moving this logic into the debugid crate by adding a `DebugId::from_pdb20_signature` constructor and describing what is being done.

flub: > A more thorough solution would be to extend the debugid crate to handle these obsolete formats, but that doesn't seem particularly worthwhile given that they're unlikely to be encountered in the wild nowadays.

This is the option I"m exploring in https://github.com/getsentry/rust-debugid/pull/10 though that is not the right implementation at all currently.  I have something that's actually identical memory-layout wise locally that i'll finish tomorrow.

luser: Thanks! I'm amenable to either path. I appreciate you doing the work here, I know this line of inquiry feels very nit-picky but I have run into these sorts of weird edge cases enough in my life to be wary of leaving them for someone else to hit.

flub:
@Gankra how happy are you with the casing changes in the code ids?  This affects your json output format and not sure if you correct for this at lookup time (this is what symbolicator does, it fixes the casing based on the symbol server rather than hoping the casing of the original code id was correct).

Otherwise the debugid crate has been updated to represent the PDB 2.0 format so once that's released I'll update this PR.

Gankra:
I'll confirm with the others that this won't break us, but it sounds like keeping things uppercase would be a good compat improvement!

Gankra:
Wait actually hold on I'm a bit confused. I vaguely remember reading some discussion of certain symbol servers being case-sensitive, and I thought the conclusion was "ah yes, everyone expects these to always be uppercase". But now I can't find that (did it happen elsewhere...?).

What is the precise argument for this, again?

Gankra:
Ok so the precise change is that debugids (used for getting symbols/pdbs) would stay the same (always upper case with the "breakpad" formatter..? or does it preserve case?)

But codeids (used for getting the actual binaries, something mozilla doesn't do afaict) would get newly lowercased? 

We've tested and microsoft's official tooling seems to be wholy case-insensitive. Our tooling *is* case-sensitive but if only codeids change then I'm not sure we really care?

luser:
> But codeids (used for getting the actual binaries, something mozilla doesn't do afaict) would get newly lowercased?

Correct. The Windows binaries are available on Mozilla's symbol server, and Microsoft's debuggers will fetch the matching binaries given a minidump, but nothing in the crash reporting machinery relies on being able to do this itself.

The only usage I've seen of this is heycam's chkimg tool, [which does use the code id to build a URL to fetch the matching PE file](https://github.com/heycam/chkimg/blob/8efa6d1c12118c90bd6e64e865001411cc13ae0d/src/main.rs#L397).

I don't think there's a right answer here. There is no "original" case, all the IDs in question are stored as integers (or GUIDs, which are just a series of integers) in minidumps, so the case in question is just a matter of how you format those integers as hex strings. I don't have any evidence to suggest I know better than the Sentry folks on this matter. :)

Gankra:
I think we can land this for now and let it bake a bit before cutting a release (this is gonna be a big release lmao fuck)

flub:
> Ok so the precise change is that debugids (used for getting symbols/pdbs) would stay the same (always upper case with the "breakpad" formatter..? or does it preserve case?)

(so non-overlapping timezones seems a challenge, apologies to not responding :/)

Yes, DebugIds maintain the same formatting, also their breakpad formatting (which is uppercase UUID, lowercase age).

The change is in CodeIds which are now all lowercase.  Before the case was mixed, but as pointed out already this is stored in integers etc in the minidump so don't really have a case.  We have encountered some case-sensitive symbol servers which is why our users can configure exact casing rules used by the symbol servers but the casing conversions are only done just before the symbol server requests.

PR #433, opened 2022-02-08
fix missing format arg

Gankra:


Gankra:
No idea why CI didn't complain on the original PR...

5225225:
unreachable!(string_literal) is not format_args

it's a string literal

yes that's cursed

it was fixed in edition 2021

Gankra:
It's still *weird* that apparently our `master` CI is different than our PR CI? Maybe toolchains *just* rolled over.

PR #435, opened 2022-02-08
Make fuzzers work on async

5225225:


Gankra (approved):
Nice yeah, that's pretty simple!

PR #436, opened 2022-02-08
Actually await in the process

5225225:
Also run fmt because CI doesn't check formatting inside fuzzing targets.


Gankra (approved):
pobody's nerfect

PR #437, opened 2022-02-08
Fix up links to MSDN docs

relaxolotl:
The links to various MSDN docs in the minidump parser are now broken, thanks to two changes:
1. The `/desktop` part of the path has been renamed to `/win32`
2. The leading underscore from the struct names have been stripped from the URL

Certain URLs will be correctly redirected if they're only affected by the first change, i.e. https://docs.microsoft.com/en-us/windows/win32/Memory/memory-protection-constants. I've opted to update those regardless just to keep everything up to date.

URLs affected by both changes lead to a 404, hence this PR. I've double checked the links to make sure they actually work as advertised, ie they go somewhere useful and the name of the struct matches. I didn't check to make sure that the contents of the structs match up, however.

### DIFF
--- a/minidump-common/src/format.rs
+++ b/minidump-common/src/format.rs
@@ -4,7 +4,7 @@
 //! some [Breakpad][breakpad] and [Crashpad][crashpad] extension types are defined here and should
 //! match the definitions from those projects.
 //!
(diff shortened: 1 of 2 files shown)


Gankra (approved):
oh awesome, thanks so much!

Issue #439, opened 2022-02-09
Subtract with overflow when scanning stack frames


5225225:
Stack trace:

```
Running: fuzz/artifacts/process/minimized-from-52af49762b2e358f53eaaa27e96cc12c1789f5b6
thread '<unnamed>' panicked at 'attempt to subtract with overflow', /home/jess/src/rust-minidump/minidump-processor/src/stackwalker/x86.rs:271:33
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
==3738815== ERROR: libFuzzer: deadly signal
    #0 0x55f7e4e4b2f1 in __sanitizer_print_stack_trace /rustc/llvm/src/llvm-project/compiler-rt/lib/asan/asan_stack.cpp:87:3
    #1 0x55f7e7014078 in fuzzer::PrintStackTrace() (/home/jess/.cache/cargo/target/x86_64-unknown-linux-gnu/release/process+0x365b078)
    #2 0x55f7e6fee155 in fuzzer::Fuzzer::CrashCallback() (/home/jess/.cache/cargo/target/x86_64-unknown-linux-gnu/release/process+0x3635155)
    #3 0x7f082ffbd86f  (/usr/lib/libpthread.so.0+0x1386f)
    #4 0x7f082fccdd21 in raise (/usr/lib/libc.so.6+0x3cd21)
    #5 0x7f082fcb7861 in abort (/usr/lib/libc.so.6+0x26861)
    #6 0x55f7e70a64e6 in std::sys::unix::abort_internal::h1f5318f76822dfc9 /rustc/88fb06a1f331926bccb448acdb52966fd1ec8a92/library/std/src/sys/unix/mod.rs:259:14
    #7 0x55f7e4dc45f5 in std::process::abort::hbf55446b688adba4 /rustc/88fb06a1f331926bccb448acdb52966fd1ec8a92/library/std/src/process.rs:1995:5
    #8 0x55f7e6fd54c5 in libfuzzer_sys::initialize::_$u7b$$u7b$closure$u7d$$u7d$::h7c5979fb626d916c (/home/jess/.cache/cargo/target/x86_64-unknown-linux-gnu/release/process+0x361c4c5)
    #9 0x55f7e709a05a in std::panicking::rust_panic_with_hook::h3c44292d2b9e7acd /rustc/88fb06a1f331926bccb448acdb52966fd1ec8a92/library/std/src/panicking.rs:702:17
[...]

PR #442, opened 2022-02-09
Make every possible overflow (that I can see) checked

5225225:
This is still in draft because I need to

1. actually verify that all of these are *reachable* (As in, panic on the current version)
2. write tests to make sure all of the changes *worked*

These bugs weren't found by fuzzing since some of them are a pretty tight margin, and is basically "guess until you're correct". Which explains why the fuzzer couldn't *really* find em.

Going to say Fixes: #439 since that is indeed one of the issues that will be fixed by this.


minidump-processor/src/stackwalker/amd64.rs
```diff
@@ -281,13 +281,13 @@ where
             // It's unclear why, perhaps the test is buggy, but for now we
             // preserve that behaviour.
             if let Some(last_bp) = last_bp {
-                let address_of_bp = address_of_ip - POINTER_WIDTH;
+                let address_of_bp = address_of_ip.checked_sub(POINTER_WIDTH)?;
                 // Can assume this resolves because we already walked over it when
                 // checking address_of_ip values.
-                let bp = stack_memory.get_memory_at_address(address_of_bp as u64)?;
+                let bp: u64 = stack_memory.get_memory_at_address(address_of_bp as u64)?;
```
Gankra, line 287: So keeping in mind that the address_of_bp calculation *isn't* necessarily already-walked over, this code needs to be completely reworked. 

I'm *very* dubious of entering this entire block when i==0. I'm always off-by-one with stackpointer stuff because everything goes backwards but it feels like we should never try to do this analysis in that case?

If we found the instruction pointer on *the first* iteration (or if we start doing misaligned searches, for any number of iterations that would cover less memory than an entire pointer), then that means the function pushed *nothing* to the stack (or less than a pointer). This by-defintion means there is no base-pointer to recover!

So I think the *correct* fix is to just this entire last_bp analaysis on `i > 0`? And then the comment above *would* be correct and the above subtraction would be immune to underflow.

minidump-processor/src/stackwalker/amd64.rs
```diff
@@ -281,13 +281,13 @@ where
...
                 if last_bp == address_of_bp
                     && bp > address_of_ip
-                    && bp - address_of_bp <= MAX_REASONABLE_GAP_BETWEEN_FRAMES
+                    && bp.checked_sub(address_of_bp)? <= MAX_REASONABLE_GAP_BETWEEN_FRAMES
```
Gankra, line 290: This is definitely hard to see but I believe this underflow cannot happen.

If we reach this point in this code then we know:

```
address_of_bp < address_of_ip < bp
              ^               ^
     by construction          |
                              |
                immediately preceding check
```
So `bp - address_of_bp` is always positive.

minidump-processor/src/stackwalker/x86.rs
```diff
@@ -268,9 +268,11 @@ where
...
             const MAX_REASONABLE_GAP_BETWEEN_FRAMES: Pointer = 128 * 1024;
 
-            let address_of_bp = address_of_ip - POINTER_WIDTH;
-            let bp = stack_memory.get_memory_at_address(address_of_bp as u64)?;
-            if bp > address_of_ip && bp - address_of_bp <= MAX_REASONABLE_GAP_BETWEEN_FRAMES {
+            let address_of_bp = address_of_ip.checked_sub(POINTER_WIDTH)?;
+            let bp: u32 = stack_memory.get_memory_at_address(address_of_bp as u64)?;
+            if bp > address_of_ip
+                && bp.checked_sub(address_of_bp)? <= MAX_REASONABLE_GAP_BETWEEN_FRAMES
+            {
```
Gankra, line 275: Similar arguments apply to this code but it's written differently because the breakpad tests are Weird. It might be worth trying out my proposed semantics for the x64 impl here too.

Gankra (changes requested):
Thanks for the audit!

TL;DR I think all of these guards are unnecessary *except* for the `bp` case, which should be refactored to avoid the issue in a different way.

minidump-processor/src/processor.rs
```diff
@@ -264,6 +264,7 @@ where
                 let mut offsets = BTreeMap::new();
                 for unloaded in unloaded_modules.modules_at_address(frame.instruction) {
                     let offset = frame.instruction - unloaded.raw.base_of_image;
+
```
5225225, line 267: remove this

minidump-processor/src/stackwalker/amd64.rs
```diff
@@ -281,24 +281,26 @@ where
...
-                {
-                    // Final sanity check that resulting bp is still inside stack memory.
-                    if stack_memory
-                        .get_memory_at_address::<Pointer>(bp as u64)
-                        .is_some()
+                if i > 0 {
+                    let address_of_bp = address_of_ip - POINTER_WIDTH;
+                    // Can assume this resolves because we already walked over it when
+                    // checking address_of_ip values.
+                    let bp: u64 = stack_memory.get_memory_at_address(address_of_bp as u64)?;
```
5225225, line 288: remove type annotation

5225225:
Okay, what I currently have is as much as I can really do. I did try writing tests for the x86 version, but just *couldn't* get it to fail, even when I undid the `if i > 0` change.

Tried both translating the amd64 test, and also tried modifying an existing x86 test, setting the stack pointer to 0 and trying to get it to underflow. No dice.

But I've kinda left this PR long enough (sorry!) so I'm going to just say it's ready for review now and call it good enough.

minidump-processor/src/stackwalker/x86.rs
```diff
@@ -268,24 +268,27 @@ where
             // smaller than 128 KB.
             const MAX_REASONABLE_GAP_BETWEEN_FRAMES: Pointer = 128 * 1024;
 
-            let address_of_bp = address_of_ip - POINTER_WIDTH;
+            let address_of_bp = address_of_ip.checked_sub(POINTER_WIDTH)?;
             let bp = stack_memory.get_memory_at_address(address_of_bp as u64)?;
```
Gankra, line 272: Is there a reason these two lines can't be inside the `if i > 0`? This would also eliminate the need for a checked_sub, right?

minidump-processor/src/stackwalker/x86.rs
```diff
@@ -268,24 +268,27 @@ where
...
-                if stack_memory
-                    .get_memory_at_address::<Pointer>(bp as u64)
-                    .is_some()
-                {
-                    caller_bp = Some(bp);
-                }
-            } else if let Some(last_bp) = last_bp {
-                if last_bp >= caller_sp {
+
+            if i > 0 {
```
Gankra, line 274: Please also add a comment to the effect of 

If we're on the first iteration of the scan, there can't possibly be a frame pointer, 
because the entire stack frame is taken up by the return pointer. And if we're
*not* on the first iteration, then the last iteration already loaded the location
we expect the frame pointer to be in, so we can unconditionally load it here.

minidump-processor/src/stackwalker/x86.rs
Gankra: This might be why you had trouble testing your fix on x86...?

Gankra (changes requested):
Looks good overall, issues in comments.

minidump-processor/src/stackwalker/amd64.rs
```diff
@@ -281,24 +281,26 @@ where
...
-                let bp = stack_memory.get_memory_at_address(address_of_bp as u64)?;
-                if last_bp == address_of_bp
-                    && bp > address_of_ip
-                    && bp - address_of_bp <= MAX_REASONABLE_GAP_BETWEEN_FRAMES
-                {
-                    // Final sanity check that resulting bp is still inside stack memory.
-                    if stack_memory
-                        .get_memory_at_address::<Pointer>(bp as u64)
-                        .is_some()
+                if i > 0 {
```
Gankra, line 284:
```suggestion
                // If we're on the first iteration of the scan, there can't possibly be a
                // frame pointer, because the entire stack frame is taken up by the return
                // pointer. And if we're not on the first iteration, then the last iteration
                // already loaded the location we expect the frame pointer to be in, so we can
                // unconditionally load it here.
                if i > 0 {
```

PR #443, opened 2022-02-10
Implement MinidumpMemory64List stream

wwylele:
I happened to need to read a full dump and found this missing in the library, so I implemented it.

To reuse the existing code, I abstracted some functions on `MinidumpMemoryList` into a generic type `MinidumpMemoryListGeneric`, and made both `MinidumpMemoryList` and the new `MinidumpMemory64List` type aliases of the generic type.

### DIFF
--- a/minidump/src/minidump.rs
+++ b/minidump/src/minidump.rs
@@ -12,6 +12,7 @@ use scroll::{self, Pread, BE, LE};
 use std::borrow::Cow;
 use std::collections::BTreeMap;
 use std::collections::HashMap;
+use std::convert::TryInto;
 use std::fmt;
 use std::fs::File;
 use std::io;
@@ -330,10 +331,11 @@ pub struct MinidumpSystemInfo {
 }
 
 /// A region of memory from the process that wrote the minidump.
+/// This is the underlying generic type for [MinidumpMemory] and [MinidumpMemory64].
 #[derive(Clone, Debug)]
-pub struct MinidumpMemory<'a> {
+pub struct MinidumpMemoryBase<'a, Descriptor> {
     /// The raw `MINIDUMP_MEMORY_DESCRIPTOR` from the minidump.
-    pub desc: md::MINIDUMP_MEMORY_DESCRIPTOR,
+    pub desc: Descriptor,
     /// The starting address of this range of memory.
     pub base_address: u64,
     /// The length of this range of memory.
@@ -341,6 +343,12 @@ pub struct MinidumpMemory<'a> {
     /// The contents of the memory.
     pub bytes: &'a [u8],
 }
+
+/// A region of memory from the process that wrote the minidump.
+pub type MinidumpMemory<'a> = MinidumpMemoryBase<'a, md::MINIDUMP_MEMORY_DESCRIPTOR>;
+
+/// A large region of memory from the process that wrote the minidump (usually a full dump).
+pub type MinidumpMemory64<'a> = MinidumpMemoryBase<'a, md::MINIDUMP_MEMORY_DESCRIPTOR64>;
 
 #[allow(clippy::large_enum_variant)]
 #[derive(Debug)]
@@ -496,13 +504,20 @@ pub struct MinidumpException<'a> {
 }
 
 /// A list of memory regions included in a minidump.
+/// This is the underlying generic type for [MinidumpMemoryList] and [MinidumpMemory64List].
 #[derive(Debug)]
-pub struct MinidumpMemoryList<'a> {
-    /// The memory regions, in the order they were stored in the minidump.
-    regions: Vec<MinidumpMemory<'a>>,
+pub struct MinidumpMemoryListBase<'a, Descriptor> {
+    /// The memory regions, in the order they were stored in the  minidump.
+    regions: Vec<MinidumpMemoryBase<'a, Descriptor>>,
     /// Map from address range to index in regions. Use `MinidumpMemoryList::memory_at_address`.
     regions_by_addr: RangeMap<u64, usize>,
 }
+
+/// A list of memory regions included in a minidump.
+pub type MinidumpMemoryList<'a> = MinidumpMemoryListBase<'a, md::MINIDUMP_MEMORY_DESCRIPTOR>;
+
+/// A list of large memory regions included in a minidump (usually a full dump).
+pub type MinidumpMemory64List<'a> = MinidumpMemoryListBase<'a, md::MINIDUMP_MEMORY_DESCRIPTOR64>;
 
 /// Information about an assertion that caused a crash.
 #[derive(Debug)]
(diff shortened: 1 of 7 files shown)


minidump/src/minidump.rs
```diff
@@ -494,14 +501,21 @@ pub struct MinidumpException<'a> {
 }
 
 /// A list of memory regions included in a minidump.
+/// This is the underlying generic type for [MinidumpMemoryList] and [MinidumpMemory64List].
 #[derive(Debug)]
-pub struct MinidumpMemoryList<'a> {
-    /// The memory regions, in the order they were stored in the minidump.
-    regions: Vec<MinidumpMemory<'a>>,
+pub struct MinidumpMemoryListGeneric<'a, Descriptor> {
```
luser, line 506: I don't want to bikeshed too much, but having `Generic` is pretty uncommon in Rust type names. I don't have a great alternative, but here are a few suggestions to think about:

- `MinidumpMemory<T>`: this would require renaming the existing `MinidumpMemory`, we could use `MinidumpMemory32`?
- `MinidumpMemoryBase<T>`
- Just `Memory<T>`?

minidump/src/minidump.rs
```diff
@@ -1648,6 +1715,68 @@ impl<'a> MinidumpStream<'a> for MinidumpMemoryList<'a> {
...
+        let mut offset = 0;
+        let u: u64 = bytes
+            .gread_with(&mut offset, endian)
+            .or(Err(Error::StreamReadFailure))?;
+
+        let mut rva: u64 = bytes
+            .gread_with(&mut offset, endian)
+            .or(Err(Error::StreamReadFailure))?;
+
+        let (count, counted_size) = ensure_count_in_bound(
```
luser, line 1735: :+1:

luser (changes requested):
Thanks for the PR! @Gankra and I were just talking about this, I have taken a crack at supporting this in both Breakpad's codebase and in this crate in the past but never quite made it work sensibly.

The only thing I want to sort out before merging this is the type names, which I've commented on below.

@5225225 I think all the arithmetic and bounds checking here is done safely, but since you've been auditing the codebase for here would you mind double-checking?  `impl<'a> MinidumpStream<'a> for MinidumpMemory64List<'a>` is the interesting bit.

It would be nice to have a test of some sort here so that we don't accidentally break this. I don't have a Windows dev machine handy currently, but if you do you could run a "hello world" program and use WinDBG's `.dump /f` command to write a full memory dump of it, then stick it in `testdata` and write a simple integration test for it in `test_minidump.rs`. (I won't block merging on this, but we have ample experience that shows that if you don't have tests for a part of the API it's either already broken or it winds up broken.)

minidump/src/minidump.rs
```diff
@@ -1648,6 +1715,68 @@ impl<'a> MinidumpStream<'a> for MinidumpMemoryList<'a> {
...
+            bytes,
+            u as usize,
```
5225225, line 1737: This cast is the only suspicious thing I found here.

Because if we're on a 32 bit platform (which i'm going to assume is reasonable, but a 16 bit platform is not, people are probably not parsing minidumps on their SNES and if they are I would like to meet them), then a valid dump could actually lose information here, and we'll think count is shorter than it actually is.

But that's fine, actually, because on any platform, if the file has more than *our* usize's count of elements, we couldn't have even got here, because we can fit the full dump in memory.

So it's Fine Actually but it's subtle as to *how* it's fine.

That being said, if we see a dump that's got an invalid count of elements, it will selectively pass/fail based on which platform we're on (I can craft a dump that only fails on 64 bit platforms), which seems non-ideal.

TLDR: I'd make this a checked try_into, returning an error if it's out of bounds. That won't introduce any failures on valid files, but would make invalid files fail regardless of the platform's pointer size.

Maybe this method should be changed to internally do the conversion, and take `TryInto<usize>` so you don't need to do it yourself.

Up to @luser though.

wwylele:
Addressed the comments and added a test. The attached full dump is relatively large (8MB)...

minidump/src/minidump.rs
```diff
@@ -2041,7 +2171,7 @@ impl<'a> MinidumpLinuxMapInfo<'a> {
   base_address          = {:#x}
   final_address         = {:#x}
   kind                  = {:#?}
-  permissions           = 
+  permissions           =
```
Gankra, line 2174:
```suggestion
  permissions           = 
```

This whitespace is intentional

Gankra:
Oh another place where we enumerated implemented streams: https://github.com/luser/rust-minidump/blob/932e4fdcba191b9cad2bdf13609a585ba72272f5/minidump/src/minidump.rs#L4537

Gankra:
I have been a bit slow to react to this because I have been pondering a Grand Vision For Memory in the background for a while, and this is forcing me to actually form it into something coherent instead of Cool Dream.

Basically, I had always been picturing that this API would end up being similar to how we unify MemoryInfo and LinuxMaps into UnifiedMemoryInfo (not concerned with naming here):

https://github.com/luser/rust-minidump/blob/932e4fdcba191b9cad2bdf13609a585ba72272f5/minidump/src/minidump.rs#L114-L123

I don't really love how I've done UnifiedMemoryInfo, so I have been considering introducing a new API to the `Minidump` type itself that would basically be `get_memory_info(&self) -> Option<UnifiedMemoryInfo>`.

Under the hood this would try to deserialize both known memory info streams for you, so as a user of rust-minidump you *really* don't need to care about which kind you have, unless you want to specifically drill into specific features of one of them.

I was *imagining* introducing MEMORY64 support to take a similar path: we have a UnifiedMemory type (again names not important) which is an enum(?) wrapping the two forms and you get it with `Minidump::get_memory` which deserializes both kinds and provides an abstract interface over both. I don't think it would even need to change the interface since they're both fine with talking about 64-bit addresses only.

Now with all that written out and said, it's concrete enough in my mind that I don't think your PR fundamentally changes the ability to introduce that? **Which is really good.** You don't actually try to introduce an abstraction over the two that anything uses. So e.g. minidump-processor won't actually learn to use MEMORY64 dumps with this PR (which is fine, limited scope is nice and easy to review!).

I am going to sleep on this and come back to it tomorrow.

minidump/src/minidump.rs
```diff
@@ -4965,7 +5095,7 @@ mod test {
         // Whitespace intentionally wonky to test robustness
         let input = b"
 
- a90206ca83eb2852-b90206ca83eb3852 r-xp  10bac9000 fd:05 1196511 /usr/lib64/libtdb1.so  
+ a90206ca83eb2852-b90206ca83eb3852 r-xp  10bac9000 fd:05 1196511 /usr/lib64/libtdb1.so
```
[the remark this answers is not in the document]
wwylele: I am thinking about turning this into something more explicit like `\x20`. My IDE has trailing space auto trimming on and I would guess many people do...

Gankra:
Apparently the primary mulling I did was "fuck I really need to document this procedure", so I wrote some WIP "contributing docs" with a section dedicated to adding a stream:

https://github.com/Gankra/rust-minidump/blob/contrib/CONTRIBUTING.md#adding-a-minidump-stream

wwylele:
I added more stuff according to the guideline. For the unit test with synthesized data I currently only added `test_memory64_list` (which is mostly just `test_memory_list` but with a second memory block). All other tests `test_memory_list_*` could be duplicated for memory64 but it is kind of silly. I think we could start introducing some abstraction for these unit tests (an enum? a trait?). I left these tests out, and we can add them in once we decide the direction of abstraction.

Gankra (approved):
Wow thanks so much for soldiering through everything!

I'm inclined to merge this, but will give anyone else a bit to chime in.

Gankra:
Also don't worry about the clippy failures, rust just updated and added some new lints we haven't fixed yet. Not your problem.

luser (approved):
This looks great to me, thanks for going the extra mile with the tests and everything! It's obviously good to ensure that the code works as expected, but also having the tests ensures that we don't accidentally break this feature you added that you presumably want to keep using. :)

Gankra:
Thanks a bunch for all the hard work!!!

PR #451, opened 2022-02-16
Merge minidump_dump into minidump-stackwalk --dump

### LINKED ISSUE #304: Merge minidump_dump into minidump-stackwalk
~~While I have little love for minidump_dump, it has its uses and should be something you can `cargo install` and whatnot. That means pulling it out to be its own top-level crate just like minidump-stackwalk.~~

Merge minidump_dump's code into minidump-stackwalk as another output format (--raw-dump):

minidump_dump is here:

https://github.com/luser/rust-minidump/blob/master/minidump/src/bin/minidump_dump.rs

A new --raw-dump flag should be added, based on `--json`: https://github.com/luser/rust-minidump/blob/2dc7ecc384f8ed123387cace30157c5288b6b5f3/minidump-stackwalk/src/main.rs#L33-L40

Add it to the arg group as well: https://github.com/luser/rust-minidump/blob/2dc7ecc384f8ed123387cace30157c5288b6b5f3/minidump-stackwalk/src/main.rs#L66-L68

Incorporate it into the logic that disabled the implicit --human flag: https://github.com/luser/rust-minidump/blob/2dc7ecc384f8ed123387cace30157c5288b6b5f3/minidump-stackwalk/src/main.rs#L347-L348

Copy minidump_dump's code into a new function and call it here if that's the right output format: https://github.com/luser/rust-minidump/blob/2dc7ecc384f8ed123387cace30157c5288b6b5f3/minidump-stackwalk/src/main.rs#L374

Be sure to hoist out the output stream logic so that --output-file works right: https://github.com/luser/rust-minidump/blob/2dc7ecc384f8ed123387cace30157c5288b6b5f3/minidump-stackwalk/src/main.rs#L397-L403

Extra credit: audit if there's any weird interactions with existing flags

Add an `insta` test for the new output, based on this one: https://github.com/luser/rust-minidump/blob/2dc7ecc384f8ed123387cace30157c5288b6b5f3/minidump-stackwalk/tests/test-minidump-stackwalk.rs#L40-L57

Save the new snapshot (I can do this for you if you want):

```
> cargo test (this will fail) 
> cargo install insta (install snapshot review tool)
> cargo insta review (accept)
> cargo test (this should pass now)
```

Then remove minidump_dump.

Gankra:
Fixes #304


Gankra:
With apologies to anyone making literally any change to rust-minidump which will now almost certainly show up as a failed test lol

Gankra:
(The original issue proposed the flag be called --raw-dump to avoid confusion over "this is how you pass the minidump" or whatever, but I think I like --dump better overall)

minidump-stackwalk/src/main.rs
```diff
@@ -373,6 +384,23 @@ async fn main() {
...
+                output_f = File::create(output_path).unwrap();
+                &mut output_f
+            } else {
+                stdout = std::io::stdout();
+                &mut stdout
+            };
+
+            // minidump_dump mode
+            if raw_dump {
+                return print_minidump_dump(&dump, &mut output).unwrap();
```
luser, line 401: Do you really want this to respect the `--cyborg` option? Feels like it ought to be separate. I'd also suggest pulling in the [`pager`](https://crates.io/crates/pager) crate ([using it is a 1-liner](https://github.com/luser/ppjson/blob/977264905900c96abd672263f8eaed41d592dbaa/src/main.rs#L48)) to get user-friendly output here instead of dumping multiple screens' worth of output to a terminal.

luser (approved):
Totally sensible plan.

PR #454, opened 2022-02-17
Fix PDB 2.0 DebugId format

flub:
The previous implementation did not correctly represent the DebugId of
(very old) PDB 2.0 PE files.  The new debugid release suppots these
correctly.

### DIFF
--- a/minidump/src/minidump.rs
+++ b/minidump/src/minidump.rs
@@ -942,10 +942,7 @@ impl Module for MinidumpModule {
                 .ok()?;
                 Some(DebugId::from_parts(uuid, raw.age))
             }
-            Some(CodeView::Pdb20(ref raw)) => {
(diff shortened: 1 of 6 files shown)


Gankra:
Could you add a test for this behaviour?

Gankra (approved):
approved but would prefer tests

minidump/src/minidump.rs
```diff
@@ -4852,6 +4849,48 @@ mod test {
         );
     }
 
+    #[test]
+    fn test_module_list_pdb20() {
+        let name = DumpString::new("single module", Endian::Little);
+        let cv_record = Section::with_endian(Endian::Little)
+            .D32(md::CvSignature::Pdb20 as u32) // cv_signature
+            .D32(0x0) // cv_offset
```
flub, line 4857: I have no idea what the cv_offset is supposed to contain here, but I guess it doesn't matter as that's not what I changed or am testing.

Gankra:
thanks!!

PR #457, opened 2022-02-18
Attempt to parse out build and version info from csd_version if it can't be found

relaxolotl:
In minidumps created by breakpad on Linux, the OS's version number can only be found in its MinidumpSystemInfo's csd_version field. See the minidump writer [here](https://chromium.googlesource.com/breakpad/breakpad/+/refs/heads/main/src/client/linux/minidump_writer/minidump_writer.cc#1281), which shows where it sources the contents of `csd_version` from. A quick search for`major_version`/`minor_version`/`build_version` in the same file also turns up nothing.

This adds in a helper to parse out the OS version number and build number if it turns out there's a chance that a Linux minidump is being examined. This is an adaptation of [symbolic's implementation](https://github.com/getsentry/symbolic/blob/6b3f424a60dcffa31fbde1deccde3ac6e1bbc0bd/symbolic-minidump/src/processor.rs#L856-L865) of the same functionality, but with string splitting instead of using regexes. The original `csd_version` field is left intact on the `MinidumpSystemInfo` in case the value is needed for any reason by any consumers. 

I've stolen the linux minidump used by symbolic to be used here for testing purposes. This isn't notably larger than the other minidumps already in this repository, so I'm hoping it's fine for me to include this. The tests targeting the linux minidump aren't particularly thorough, though: There's literally just the one to check that this patch works. 

Fun breadcrumbs: Hopefully to nobody's surprise, this is related to https://github.com/getsentry/symbolicator/issues/620.


minidump/src/minidump.rs
```diff
@@ -2626,6 +2630,51 @@ impl MinidumpSystemInfo {
...
     }
+
+    /// Strings identifying the version and build number of the operating
+    /// system.
+    ///
+    /// Tries to parse the version number from the build if it cannot be found
+    /// in the version string. If the stream already contains a valid version
+    /// number or parsing from the build string fails, this will return what's
+    /// directly stored in the stream.
+    pub fn os_parts(&self) -> (String, Option<String>) {
```
relaxolotl, line 2641: I'm not entirely happy with this name, but I couldn't come up with a better alternative so I went with the name used by symbolic. Suggestions are very welcome here.

flub, line 2641: nit reading just the doc comment and signature i have no idea what the parts in the return type tuple are.

flub: "Returns a tuple of the version and optional build number of the operating system" might help me.

minidump/src/minidump.rs
```diff
@@ -2626,6 +2630,51 @@ impl MinidumpSystemInfo {
...
+        }
+
+        // Try to parse the Linux build string. Breakpad and Crashpad run
+        // `uname -srvmo` to generate it. This roughtly resembles:
+        // "Linux [version] [build...] [arch] Linux/GNU"
+        let raw_build = self.csd_version().unwrap_or(Cow::Borrowed(""));
+        let mut parts = raw_build.split(' ');
+        // Skip past "Linux", grab uname -r portion
+        let version = parts.nth(1).unwrap_or("0.0.0");
+        // Start from the end and pull out everything until only the build is left
```
flub, line 2665: this comment threw me off a bit and only after looking at what `uname -v` looks like on several machines did it make sense to me.  I think it's easier to describe this as "trim the operating system (uname -o) and machine (uname -m). of the tail" or something along those lines.

minidump-processor/src/processor.rs
```diff
@@ -93,17 +93,7 @@ where
...
-        dump_system_info.raw.major_version,
-        dump_system_info.raw.minor_version,
-        dump_system_info.raw.build_number
-    );
-
-    let os_build = dump_system_info
-        .csd_version()
-        .map(|v| v.trim().to_owned())
-        .filter(|v| !v.is_empty());
+    let (os_version, os_build) = dump_system_info.os_parts();
```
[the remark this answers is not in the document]
flub: your approach and reasoning seems sensible to me.

flub (approved):

minidump-processor/src/processor.rs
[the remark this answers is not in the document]
luser: Without reading your entire patch: I think it's important that the `Minidump*` classes provide access to the exact data contained within the minidump, because there are lots of ways to interpret specific pieces of data. However, it's totally fine to also have helper methods to pick out specific useful bits that require interpretation, such as the `code_id` / `debug_id` bits on `MinidumpModule`.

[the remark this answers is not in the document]
relaxolotl: Thank you for the context! In that case, I'll stick with what's in the PR right now as it respects that and I'll add just a little bit extra to the helper's docstring to provide some guidance on what it does.

PR #468, opened 2022-02-19
[WIP] use newer version of test-assembler to get better errors

5225225:
This won't work in CI since I used relative paths (I guess I could have used a git dependency for the crate, but :shrug:)

The diff is.... smaller than I was expecting.


luser:
Nice! Let's get your test-assembler PR merged and we can publish a new release of that crate to use here.

Gankra:
Gonna close this as bitrotted, r i p to the future

PR #471, opened 2022-02-24
Return a nil code ID when a module doesn't have a recognizable CodeView signature

relaxolotl:
Sentry's currently in the process of testing out rust-minidump by comparing its results to breakpad's on various minidumps out in the wind. [It looks like modules in minidumps generated on Linux may omit CV records entirely if a build ID could not be found for them.](https://chromium.googlesource.com/breakpad/breakpad/+/8a0edac9abfec72e0a86aebd6a0a38761c7c8962/src/client/linux/minidump_writer/minidump_writer.cc#651) When that happens, the entire module is zeroed-out except for its name, its image's base memory address and its image's size. 

Currently rust-minidump returns a `CodeId` derived from the timestamp (which is empty) and the image size, which isn't exactly correct and can be a little misleading. This PR updates the code to match what [breakpad does](https://chromium.googlesource.com/breakpad/breakpad/+/8a0edac9abfec72e0a86aebd6a0a38761c7c8962/src/processor/minidump.cc#1971) by returning a nil `CodeId` in these situations instead.


minidump/src/minidump.rs
```diff
@@ -912,26 +912,31 @@ impl Module for MinidumpModule {
...
-                // TODO: Figure out a way to get access to `MinidumpSystemInfo`
-                // here instead of relying on `age`.
-                CodeId::new(format!("{:#}", raw.signature))
-            }
-            _ => {
-                // TODO: Breakpad stubs this out on non-Windows.
-                CodeId::new(format!(
-                    "{0:08X}{1:x}",
-                    self.raw.time_date_stamp, self.raw.size_of_image
-                ))
```
relaxolotl, line 933: Sorry for the massive diff here. I realized that the order of the match branches didn't... match the order used in other functions (e.g. `debug_identifier`, `debug_file`, etc) so I shuffled these around so they followed the same ordering as the rest of the file.

minidump/src/minidump.rs
```diff
@@ -912,26 +912,31 @@ impl Module for MinidumpModule {
...
-                // between these two platforms since macOS uses PDB70 instead of
-                // its own dedicated format.
-                // See the following linked comment in a Windows PDB writer:
-                // https://github.com/microsoft/microsoft-pdb/blob/082c5290e5aff028ae84e43affa8be717aa7af73/PDB/dbi/pdb.cpp#L602-L606
-                // TODO: Figure out a way to get access to `MinidumpSystemInfo`
-                // here instead of relying on `age`.
-                CodeId::new(format!("{:#}", raw.signature))
-            }
-            _ => {
-                // TODO: Breakpad stubs this out on non-Windows.
```
relaxolotl, line 929: I'm not too sure what exactly this meant originally but I _did_ remove this in the PR, since we have relatively well-defined behaviours for all three known `CodeView` signatures as well as those that are unknown. Would like to double check to make sure this is correctly addressed and therefore fine to remove.

minidump/src/minidump.rs
gabrielesvelto: You could split your change and the re-ordering of the matches in two separate patches. It helps with review and it makes the intention clearer when looking at the history.

flub:
The patch seems good to me, and this comment is probably more suited to be addressed in an other PR, if at all.  But I'm curious what other people think of the signature of `.code_identifier() -> CodeId` compared to `.debug_identifier() -> Option<DebugId>` which kind of makes more sense since a module could not have an identifier (they're fonts, or whatever other data as shown in this description).  To be fair I've always struggled a bit with the DebugId/CodeId types having "nil" versions of them which essentially bakes in a `None` value, but I understand where this came from as they started with having to deal with whatever breakpad pulled out of minidumps.

So would it make sense to make `.code_identifier()` return an option?

luser:
> So would it make sense to make `.code_identifier()` return an option?

I think so, yeah. For Windows minidumps the "code identifier" is always available because it's just those bits from the PE headers that are in the minidump module struct. For non-Windows platforms it has no real definition beyond the ELF BuildId stuff I shoved in there, and the recent Mach-O UUID changes here.

luser:
Returning `Option<CodeId>` seems strictly better than Breakpad's behavior of returning the literal string `"id"`, in any event!

Gankra:
I agree that an Option is probably better here than a sentinel value.

minidump/src/minidump.rs
relaxolotl: Thanks for the suggestion! I'm used to the aggressive squashing in sentry's repositories hence the single commit in the original branch; I've broken the reordering out so it's easier to follow what's going on based on the history.

This would mean that ideally this isn't squashed into the trunk if/once it is merged though. 🙏

relaxolotl:
what did i do to incite clippy's wrath on code that i didn't touch

Gankra:
new rust release, don't worry about it

relaxolotl:
@Gankra should I opening a PR to resolve them assuming nobody's already working on it? Most of the problems appear to be pretty simple to fix

relaxolotl:
Cleaned things up; Thanks for the quick feedback! The one of the checks will be failing for a little bit but should be fixed soon™️ in a different PR.

I mulled over the idea of splitting out the type change to `code_identifier`'s return value (`CodeId`->`Option<CodeId>`) into a different PR, but I settled on keeping it here because this is where it introduces the behaviour of returning some sort of nil value. Instead of having two separate conversations about the change and its impact, we might as well do it all in one go here. 

The type change is split out into a different commit if it turns out that opening a completely separate PR is preferred. The only cause for doing so that I could think of is if something/somebody needs to have the nil-returning behaviour but isn't ready to take on the type change yet. It was simple enough for symbolicator when I was testing it locally with this patch (these patches?), but this might not be the case for Mozilla or other users of this crate, so let me know if that's a blocker.

minidump-common/src/traits.rs
```diff
@@ -19,7 +19,7 @@ pub trait Module {
     /// builds of the same code module.  This may contain a uuid, timestamp,
     /// version number, or any combination of this or other information, in an
     /// implementation-defined format.
-    fn code_identifier(&self) -> CodeId;
+    fn code_identifier(&self) -> Option<CodeId>;
```
flub, line 22: Would it be good to document on the trait why this can be `None`?  Describe somehow that some minidump writers write non-object things like mmapped fonts etc into the module stream and that for those this can not provide a valid CodeId.

(while adding that it could be referred to in `.debug_identifier()` as well)

minidump/src/minidump.rs
```diff
@@ -1084,11 +1089,14 @@ impl Module for MinidumpUnloadedModule {
     fn code_file(&self) -> Cow<'_, str> {
         Cow::Borrowed(&self.name)
     }
-    fn code_identifier(&self) -> CodeId {
-        CodeId::new(format!(
+    fn code_identifier(&self) -> Option<CodeId> {
+        // TODO: This should be returning None if the unloaded module is coming
+        // from a non-Windows minidump. We'll need info about the operating
+        // system, ideally sourced from the SystemInfo to be able to do this.
```
flub, line 1095: IIRC when I looked at unloaded modules I've only seen them from windows' minidump writer.  But I guess that's fairly academic as eventually other writers might start doing this too.

However more generally I'm not sure if this comment is really true?  If say Linux was able to give us an unloaded module it could still have been a really shared object and thus have had a real CodeId I think.

relaxolotl: Could you clarify the bit about the comment being incorrect? I could see the argument for its inaccuracy in the sense that `MinidumpUnloadedModule::code_identifier` currently would only ever be invoked for Windows minidumps right now, so there wouldn't exist any code paths that would lead to this when parsing non-Windows minidumps. Is that what the second paragraph is trying to convey?

Were you pointing out its inaccuracy with the hopes that the comment would be removed, or simply corrected? Do you feel that commenting on this adds nothing and should be removed?

minidump-common/src/traits.rs
relaxolotl: I think some documentation on this may be useful, but I have mixed feelings about adding this note to the trait itself, as these platform and/or implementation-specific details may not be applicable to every implementation of the trait (i.e. would this be applicable to `SimpleModule`?) 

It does look like the modules that fall into the `None` case are coming from shared memory segments (It's a bunch of stuff sitting in System V shared memory) which sounds quite similar to mmapped fonts, so I'll leave a comment above the branch noting this. Hopefully this works as a good compromise 🤞

Gankra:
I landed the clippy fixes.

Are you ready to land?

relaxolotl:
Thanks for the heads up! I'd be more comfortable with at least one approval from somebody, but I think the PR is otherwise good to go. Most of what's under contention is related to code comments, which gives me some hope that the actual logic itself is fine.

Gankra:
I'll leave it up to flub, they seem to be the one reviewing this

relaxolotl:
I'll rebase for the time being to appease the clippy check, but in that case I reckon this'll be ready to land Monday-ish.

minidump/src/minidump.rs
flub: I don't understand the claim that this should always be None on non-Windows, to the extend that having such a comment seems misleading to me.  The 2nd paragraph is trying to explore if there is a reason why a hypothetical linux minidump writer might never be able to produce something that has a CodeId.  And why such a hypothetical writer may need us to have knowledge of the OS to be able to give the correct return value.

I was pointing this out in the hopes that either I am shown to be wrong and the comment could be tweaked to clarify this better, or I'm not wrong and I guess in that case maybe removing the comment rather than tweaking it makes most sense?

flub (approved):
Apologies, I didn't mean to hold this up over a comment :/

minidump/src/minidump.rs
relaxolotl: I see. Unless I've missed something, there simply isn't any way to extract a code ID from a non-Windows minidump. It's just not implemented. 

However, the current implementation still returns an incorrect value for non-Windows minidumps, [if you compare it to breakpad's.](https://chromium.googlesource.com/breakpad/breakpad/+/42d24756174947921d282ee87e2b8e48f6eed3bf/src/processor/minidump.cc#3730) Even without the comparison to breakpad, I think the simplicity of this implementation compared to the one for a loaded module raises at least one red flag about its correctness. Based on breakpad, `code_identifier` for unloaded modules should return `None` unless writers for other non-Windows OSes are changed to write unloaded modules to their minidumps. This makes the first sentence still correct, although it omits the entire story.

If you look at `MINIDUMP_UNLOADED_MODULE` it doesn't look like it stores a CodeView record, meaning that we can't use the strategy used in loaded modules to determine which operating system the unloaded module comes from purely based off of its own contents. I'm sure with some creativity we could use repurpose or augment one of its fields to get this information, but this means that the second sentence in the comment is still correct as well: we don't have enough information in just the unloaded modules stream to fix `code_identifier` so it matches breakpad.

Did this clarify things? I'm under the impression that the comments just need a little bit of rephrasing. It seems like it might be ambiguous about what it is referencing in the second sentence, for example: "why such a hypothetical writer may need us to have knowledge of the OS to be able to give the correct return value" suggests that it gives off the impression that it's talking about the minidump writer instead of the `code_identifier` impl. I do think that removing this TODO would be simply ignoring a spot where the code is (admittedly theoretically) incorrect.

flub: > I see. Unless I've missed something, there simply isn't any way to extract a code ID from a non-Windows minidump. It's just not implemented.

Ah, I thought that the CodeId is identical to the DebugId for non-windows as e.g. also described here: https://docs.rs/debugid/latest/debugid/struct.CodeId.html
So I guess that's where our confusion comes from.


But you're right, `MINIDUMP_UNLOADED_MODULE` as defined simply does not have sufficient info here.  I'd be a little surprised if a hypothetical writer on another platform tries to re-use this stream based on that I guess.  But who knows.  Anyway, doesn't do any harm pointing out the breakpad logic.

PR #483, opened 2022-03-05
This adds the `code_id` and `code_file` query items to the URL used when requesting a symbol file.

### LINKED ISSUE #479: missingsymbols, code_id, code_filename--oh my!
I was looking at [bug 1757890](https://bugzilla.mozilla.org/show_bug.cgi?id=1757890) and I'm pretty sure that the HttpSymbolSupplier isn't passing `code_id` and `code_filename` when it does symbol requests over HTTP from Tecken.

The docs for the download API are here:

https://tecken.readthedocs.io/en/latest/download.html#get---DEBUG_FILENAME---DEBUG_ID---SYMBOL_FILE-

The relevant bit in ye olde stackwalker is here:

https://github.com/mozilla-services/minidump-stackwalk/blob/634f9624f5583521e79039e9d29b498f971a4fb4/minidump-stackwalk/http_symbol_supplier.cc#L240-L252

I think this is the line in rust-minidump that's pertinent because it's building the url with the relevant path and I'm pretty sure I don't see querystring parameters added:

https://github.com/luser/rust-minidump/blob/b6f69022333493a5981c6d639f36509a9d99faf5/breakpad-symbols/src/lib.rs#L452

In order for @gabrielesvelto to be happy (which we all want), can we get the `code_id` and `code_filename` added as querystring params to symbol requests for Tecken?

(Alternatively, if I'm wrong, let me know and I can keep poking around. It's been a week, so this is entirely possible.)

gabrielesvelto:
These parameters are used by the symbol server to populate the list of missing modules and allow downstream scripts to find said modules.

This fixes #479

### DIFF
--- a/breakpad-symbols/src/lib.rs
+++ b/breakpad-symbols/src/lib.rs
@@ -212,6 +212,10 @@ pub enum SymbolError {
     /// In this case other symbol providers may still be able to find it!
     #[error("symbol file not found")]
     NotFound,
+    /// The module was lacking either the debug file or debug id, as such the
+    /// path of the symbol could not be generated.
+    #[error("the debug file or id were missing")]
+    MissingDebugFileOrId,
     /// Symbol file could not be loaded into memory.
     #[error("couldn't read input stream")]
     LoadError(#[from] std::io::Error),
@@ -435,7 +439,7 @@ fn commit_cache_file(mut temp: NamedTempFile, final_path: &Path, url: &Url) -> i
 async fn fetch_symbol_file(
     client: &Client,
     base_url: &Url,
-    rel_path: &str,
+    module: &(dyn Module + Sync),
     cache: &Path,
     tmp: &Path,
 ) -> Result<SymbolFile, SymbolError> {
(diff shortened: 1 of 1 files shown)


Gankra (approved):

Gankra:
Before landing this, just a non-blocking question: are these args also accepted by microsoft's symbol server API? I have been conservatively assuming tecken and ms symbol servers have fairly divergent APIs, but it seems that they actually just agree, at least in the path format. Does including these params change things at all?

willkg:
*old man: [with too much bravado]* This is how we've done things at Mozilla for years and my experience with building web services is that unknown querystring parameters are typically ignored!

*gankra:* ugh. that like doesn't answer my question at all.

*old man: [deflating like a popped blimp]* True.

---

I'd love to have better documentation on how the Microsoft Symbols Server works especially what the API looks like. Best docs I've found are these:

https://docs.microsoft.com/en-us/windows-hardware/drivers/debugger/microsoft-public-symbols

Are there better docs out there? Particularly ones that document the API?

luser:
> Before landing this, just a non-blocking question: are these args also accepted by microsoft's symbol server API?

No, this was purely a Tecken invention for our own convenience. Microsoft's symbol server API is "here is a big bucket of files, you can find what you need by constructing the paths properly". (Seriously, that's the extent of it.) Given that this code is written to fetch Breakpad-format symbol files, however, it's immaterial, as it's never going to be talking directly to a Microsoft symbol server.

PR #487, opened 2022-03-09
ref: Pass SystemInfo through to Unwind impls

Swatinem:
Using this, the Unwind implementations can apply heuristics
based on the platform-specific calling conventions.


loewenheim (approved):

Swatinem:
That sounds reasonable. I think we can only go as far as having OS+ISA there. There are different calling conventions even for one OS (well, 32bit Windows), but I doubt we can’t detect those?

Either way, I guess this works so far.

PR #489, opened 2022-03-09
fix: Disable FramePointer unwinding on Windows x64

Swatinem:
On Windows x64, the FP register is permitted to point to the middle of
a stack frame, and we should thus be conservative and not rely on it.

This is on top of #488.

Breakpad did land a similar all-or-nothing change recently: https://chromium.googlesource.com/breakpad/breakpad/+/88f5fc451ebdede7341cefe349dd63fed886f140%5E%21/#F0
Though I’m a bit unsure about this big hammer. The x64 FP heuristics do some validation of ip/sp/bp to make sure the recovered values look reasonable. It is possible that with this change we would discard valid results.


Swatinem:
Some more anecdotes about this:
I ran into it while trying to write a dead-simple FP unwinder:
https://swatinem.de/blog/unwind-2/#walking-the-stack
And thinking that there must be something wrong here, filed a Rust issue: https://github.com/rust-lang/rust/issues/82333
In which the folks reminded me that indeed on Windows x64, the FP can point into the middle of the stack frame.

Gankra (approved):
This is... deeply frustrating!

But I appreciate that you put in the work to do this robustly. I guess ship it?

Gankra:
Also the fuzzer needs a tweak to include system info.

Swatinem:
> Also the fuzzer needs a tweak to include system info.

Yes, I have a chain of 3 PRs that need rebasing and landing in order :-D
GitHub does not like to change the base branch when it is a remote :-(

Gankra:
Oh I see, I was fine with just landing this PR, but sure let's do it that way.

PR #492, opened 2022-03-09
Add pwrite derives

Jake-Shadle:
- Derive Default for MINIDUMP_DIRECTORY
- Minor cleanup while debugging something
- Derive Pwrite for many minidump structs
- Add amd64 context flags and derive a few more pwrite

This PR adds derives for many (only iterating with x86_64 atm) structs so that they can be written. The particular use case was for https://github.com/EmbarkStudios/minidump_writer_linux/commit/04d22cfd030bf1df2705cf250f13b4a3f0771656 which allowed me to get rid of a massive amount of duplicated type definitions that the code had. I noticed during the refactor that there were no bitflags for the context_flags values in CONTEXT_AMD64, so I added that as well, but I'm a bit iffy on the naming.


Gankra (approved):
Nice!

PR #501, opened 2022-03-14
cleanup CONTEXT_ARM64 definitions

Gankra:
* Remove the last lingering repr(packed)
* Fix the order of fpcr/fpsr in FLOATING_SAVE_AREA_ARM64
* Remove the errant claim that there is a 33rd iregs value for pc
* Pull sp out of iregs, x31 isn't a register
* Add CONTEXT_BLAHBLAH flags for ARM64
* Document/explain things a bit better


Gankra:
CC  @Jake-Shadle

minidump-common/src/format.rs
```diff
@@ -4,6 +4,26 @@
...
+//! # Type Layouts
+//!
+//! This library isn't a "proper" minidump-sys library because it doesn't use repr attributes
+//! to force Rust to layout these structs identically to how they're laid out in memory.
+//!
+//! The reasons for this are 3-fold:
+//!
+//! 1. It isn't necessary because we specify how to serialize/deserialize things with `scroll`
+//!    via `derive(Pread, Pwrite)` which uses the declared field order and not the in-memory
+//!    layout, and assumes everything is packed anyway (which as a rule, minidump types are).
```
gabrielesvelto, line 16: Note, Windows headers use `#pragma pack(4)` for all structures. So they are packed but only on 4 bytes boundaries. In practice I don't think it matters because I've never encountered holes but still, might be worth mentioning.

Gankra: good insight!

Gankra:
The more I think about it, the more I should just go all in on microsoft's definition here (and expand _OLD similarly). Nothing actually cares about these registers right now so the uniformity argument doesn't apply.

Gankra:
Yeah this is nicer. This also had the nice side-effect of making me notice that I had messed up the (already pretty rough) printing code, which is better now.

PR #509, opened 2022-03-18
pull all the platform-specific error code enums into a new module

Gankra:
CC @gabrielesvelto @Jake-Shadle


gabrielesvelto:
I like it. Especially the Windows ones which caused VSCode's rust analyzer to go OOM on machines with less than 32 GiB of memory whenever I tried to edit format.rs. There must be some corner-case in the code that causes big enums to make memory go **boom** so it's nice to have them somewhere else. I might also think of a way to auto-generate that file entirely from the Windows SDK.

Issue #105, opened 2020-10-13
Support extracting commandline arguments from Windows dumps


luser:
This isn't incredibly important, but I did write a proof of concept implementation of this a while ago that I stumbled upon, it's on the `dump-command-line` branch:
https://github.com/luser/rust-minidump/commit/4f6eae86dd16f0aaf0b6a8056989d76b691a9934
https://github.com/luser/rust-minidump/commits/dump-command-line

It requires traversing a series of undocumented NT kernel structures in the minidump memory. I can't recall if it even actually worked given the memory that's available in a minidump.

PR #510, opened 2022-03-18
WIP TEB support

Gankra:
(based on top of #509 to avoid merge conflicts)


Gankra:
This technically implements #105 but I'm not terribly happy with this approach.

The TEB and PEB are *huge* and mostly a bunch of reserved ranges (for now), so it's quite wasteful to completely deserialize them just to grab one value and then drop the whole deserialized result.

I think probably what we want to do is remove `last_error_value` from MinidumpThread and instead have MinidumpProcessor do this work more manually and extract whatever values it wants.

This is also intentionally restricted to x64 windows, so it doesn't yet replace last_error_value which *ostensibly* works on x86 windows as well (unvalidated, but we at least try it). I'm not entirely sure if the TEB and PEB actually have the same definitions on x86 and differ only in pointer-width. Or if microsoft is actually a psychic wizard and made this stuff always be 64-bit (doubtful).

this stuff is defined in winternl.h

See: 

* [TEB](https://docs.microsoft.com/en-us/windows/win32/api/winternl/ns-winternl-teb)
* [PEB](https://docs.microsoft.com/en-us/windows/win32/api/winternl/ns-winternl-peb) ([windows-sys](https://docs.rs/windows-sys/latest/windows_sys/Win32/System/Threading/struct.PEB.html))

Gankra:
Oh also yeah the naming/concepts I'm adding through the crates is kinda a mess here, just wanted to hack it up into an MVP.

--dump now *does* manage to restore a completely boring but still promising

command_line: "C:\Program Files\Firefox Nightly\firefox.exe"

minidump/src/minidump.rs
```diff
@@ -1703,6 +1730,15 @@ impl<'mdmp, Descriptor> MinidumpMemoryListBase<'mdmp, Descriptor> {
             .map(|&index| &self.regions[index])
     }
 
+    pub fn read_memory_at_address<T>(&self, addr: teb::Addr64<T>) -> Option<T>
```
luser, line 1733: Giving this method the same name as the one on `MinidumpMemoryBase` but with a slightly different argument feels odd to me. I like the typed pointer offset, though! We could use the same pattern for `RVA` / `MINIDUMP_LOCATION_DESCRIPTOR` if we wanted to be more thorough.

Gankra:
this has bitrotted too much and I don't love the impl

Issue #297, opened 2021-11-18
API stability cleanup


Gankra:
if we ever want to do a 1.0 release with stability guarantees we should:

* give every stream a 'mdmp lifetime
* return Cows everywhere we can
* remove get_ from methods (unidiomatic)
* mark lots of things `#[non_exhaustive]`
* replace lots of pub fields with accessors
* clean up type names (should we really prefix so much with Minidump?)

a lot of work, not sure it's worth it, but writing it down so it's written down.

luser:
> replace lots of pub fields with accessors

Is there a reason for this aside from hiding the implementation details?

> clean up type names (should we really prefix so much with Minidump?)

I straight up copied all the type names from Breakpad when I started. It definitely doesn't match the lib naming guidelines now in use!

Gankra:
> Is there a reason for this aside from hiding the implementation details?

There's already a few places where we lazily compute/parse a value, and making it a field kinda force commits us to a specific timing. I think it **is** fine to expose the `raw` values of each stream as a field since we kinda *have* to eagerly deserialize that value, and its layout is mandated by the format. 

Although even for those, anything that uses the `ex_stream` format is actually extensible so strictly speaking we should be doing some obfuscation/future-proofing similar to MiscInfo to support newer "revisions" of that stream.

luser:
Mentioning this here as well: https://github.com/luser/rust-minidump/pull/312#issuecomment-976752323

PR #513, opened 2022-03-22
feat: Add rudimentary MIPS support

Swatinem:
In particular, this adds support for reading the CPU type and fixes some issues related to big-endian, such as UTF16 parsing, and DebugId parsing.

The Module DebugId parsing is now done at parsing-time because it needs to be endian aware.

For sentry folks, this fixes:
* Unknown minidump arch 1: https://sentry.my.sentry.io/organizations/sentry/issues/297362
* Garbage module names for example: https://sentry.my.sentry.io/organizations/sentry/issues/297548/events/99e6508a508744dd8aa19a06c57e5d23/

I’m a bit unsure how to write a unit-test for all this with reasonable effort 🤷🏻‍♂️

### DIFF
--- a/minidump-processor/src/process_state.rs
+++ b/minidump-processor/src/process_state.rs
@@ -926,10 +926,10 @@ Unknown streams encountered:
     // Convert an integer to a hex string, with leading 0's for uniform width.
     fn json_hex(&self, val: u64) -> String {
         match self.system_info.cpu {
-            Cpu::X86 | Cpu::Ppc | Cpu::Sparc | Cpu::Arm => {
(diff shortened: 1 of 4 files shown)


Swatinem:
The second commit makes the `Cpu` enum `non_exhaustive`. I wasn’t aware before, so this makes the PR technically a breaking change.

Gankra:
Yeah that's ok, it fits with #297

Gankra (approved):
looks good. MIPS is... esoteric enough that I don't feel too bad about having good testing setup.

PR #520, opened 2022-03-24
fix: Make sure to accept SymbolFiles without trailing newlines

Swatinem:
TBH, the logic that splits things into files is a bit weird. Instead of having the parser reject anything that is not newline terminated is a bit weird. Ideally the `SymbolFile` would split things into lines and feed lines into the Parser. Since the `SymbolFile` known exactly when the input stream is exhausted.

Either way, the fix is a bit convoluted and suboptimal, but does what its supposed to, so meh.

### DIFF
--- a/breakpad-symbols/src/sym_file/mod.rs
+++ b/breakpad-symbols/src/sym_file/mod.rs
@@ -70,6 +70,26 @@ impl SymbolFile {
                 if fully_consumed {
                     return Ok(parser.finish());
                 } else {
+                    // The reader is exhausted, but the parser was not fully
+                    // consumed on the last iteration. Probably because it was
+                    // expecting a trailing newline. Lets give it one, and do
+                    // one more parse attempt.
+                    let space = buf.space();
+                    if !space.is_empty() {
+                        space[0] = b'\n';
+                        buf.fill(1);
+
+                        let input = buf.data();
+                        let consumed = parser.parse_more(input)?;
+                        if consumed > 0 {
+                            callback(&input[..consumed]);
+                        }
+
+                        if input.len() == consumed {
+                            return Ok(parser.finish());
+                        }
+                    }
+
                     return Err(SymbolError::ParseError(
                         "unexpected EOF during parsing of SymbolFile (or a line was too long?)",
                         parser.lines,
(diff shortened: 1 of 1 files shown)


Gankra:
yeah i was worried about this when i wrote the impl but relieved to see that dump_syms seemingly guaranteed it. i guess you found a case that doesn't?

loewenheim (approved):

breakpad-symbols/src/sym_file/mod.rs
```diff
@@ -128,6 +150,26 @@ impl SymbolFile {
...
+                    // one more parse attempt.
+                    let space = buf.space();
+                    if !space.is_empty() {
+                        space[0] = b'\n';
+                        buf.fill(1);
+
+                        let input = buf.data();
+                        let consumed = parser.parse_more(input)?;
+                        if consumed > 0 {
+                            callback(&input[..consumed]);
```
Gankra, line 165: hmm this part is mildly concerning. I think if the symbol file is missing the newline, we should preserve that fact in the cache, and just rely on the parser itself always handling it every time it's loaded?

breakpad-symbols/src/sym_file/mod.rs
```diff
@@ -374,24 +418,21 @@ mod test {
...
-1000 30 7 53
-",
+            b"MODULE Linux x86 ffff0000 bar\r\nFILE 53 bar.c\r\nPUBLIC 1234 10 some public\r\nFUNC 1000 30 10 another func\r\n1000 30 7 53\r\n",
+        );
+    }
+
+    #[test]
+    fn test_symbolfile_no_trailing_newline() {
+        test_symbolfile_from_bytes(
+            b"MODULE Linux x86 ffff0000 bar\nFILE 53 bar.c\nPUBLIC 1234 10 some public\nFUNC 1000 30 10 another func\n1000 30 7 53",
```
Gankra, line 435: the final line here is malformed, right? Can we have a test where the final line is actually good and make sure the parser properly read it?

Gankra (approved):
Genuinely a reasonable approach to the problem! Just a couple nits.

breakpad-symbols/src/sym_file/mod.rs
Swatinem: I think the last line is fine, its the same as in the tests above.

breakpad-symbols/src/sym_file/mod.rs
Swatinem: hm, I’m not sure I follow? I used a `- 1` here, to make sure it matches the original stream.

PR #527, opened 2022-04-04
Change memory and stack dumping to be more readable

### LINKED ISSUE #375: minidump_dump: Support dumping the stack as 32/64bit words
This would make it a lot easier to read.

i.e. something like
```patch
@@ -1518,22 +1519,22 @@ Memory
 ",
             self.desc.start_of_memory_range, self.desc.memory.data_size, self.desc.memory.rva,
         )?;
         self.print_contents(f)?;
         writeln!(f)
     }

     /// Write the contents of this `MinidumpMemory` to `f` as a hex string.
     pub fn print_contents<T: Write>(&self, f: &mut T) -> io::Result<()> {
         write!(f, "0x")?;
-        for byte in self.bytes.iter() {
-            write!(f, "{:02x}", byte)?;
+        for bytes in self.bytes.chunks_exact(8) {
+            writeln!(f, "{:016x}", u64::from_le_bytes(bytes.try_into().unwrap()))?;
         }
         writeln!(f)?;
         Ok(())
     }

     pub fn memory_range(&self) -> Option<Range<u64>> {
         if self.size == 0 {
             return None;
         }
         Some(Range::new(
```

marti4d:
Fixes #375


Gankra:
cc @jrmuizel, what do you think of this format?

Gankra:
possible notes:

* do we want to use more horizontal space here? 3 or 4 pointers per line..?
* what does this *mean* for endianess? do we show our endianess? the crashing platform's endianess..? is purely grouping the bytes as seen without messing with endianess?

jrmuizel:
This format is fine. I think one pointer per line is easier to scan than multiple pointers per line.

The idea here is to be able to easily read return addresses so it should use the crashing platform's endianess and pointer size

marti4d:
@Gankra @jrmuizel: Alrighty - The stack should now be dumping with target-system-sized words. I tested it with a 32-bit dump, and it seems to do everything correctly.

marti4d:
Also fixes #528

Gankra (approved):
hoooooly heck this is **incredible** work.

I genuinely have no comments, this is just GOOD.

PR #529, opened 2022-04-07
Make the parser much more robust to enormous symbols.

Gankra:
This teaches the parser to resize its buffer up to a reasonably maximum limit, and then if a line
slams into that maximum, it just enters a PANIC RECOVERY mode where it discards bytes until it
finds a newline or EOF, then resumes normal behaviour. This basically just ends up discarding horrible
enormous (80kb+) symbols instead of discarding the entire file.

Also this adds a bunch of "symbol:" TRACE logs to make figuring this stuff out easier in the future. In particular if we *do* end up discarding a bunch of your lines, that will be noted in the TRACE at least.


Gankra:
this also gives better diagnostics to the issues seen in #520

Gankra:
merging as a hotfix, will review/test later

PR #534, opened 2022-04-20
Add more derive impls

Jake-Shadle:
These are various changes that I needed when implementing https://github.com/rust-minidump/minidump_writer_linux/pull/21


Gankra:
The GUID From impl is weird... is it just like, a definitionally Big Endian format regardless of platform..?

Jake-Shadle:
Yes, see that's how it is in the [standard](https://datatracker.ietf.org/doc/html/rfc4122.html#section-4.1.2), FWIW, the `uuid` crate does the same thing but I didn't feel this merited adding a dependency (even though `uuid` is already being pulled in via `debugid`)

PR #548, opened 2022-05-02
Make breakpad-symbols HTTP retrieval optional

Jake-Shadle:
This makes it so that retrieving symbols via HTTP is optional, eg. https://github.com/rust-minidump/minidump-writer/pull/21#discussion_r861912364 only wants to process symbols files that are present on the local disk for testing purposes, making HTTP optional removes a large number of unnecessary dependencies to speed up CI builds.

This _also_ nukes OpenSSL in favor of using `rustls` due the many aggravations that come with OpenSSL, some of which have had previous [commits](https://github.com/rust-minidump/rust-minidump/commit/a142f208f9d2899be3ca168aa1de0bb43dcffef3) to workaround. `rustls` on the other hand doesn't have really, any aggravations, at least that we have found in 3+ years of usage, everything just works better, including cross compilation and other scenarios that openssl is horrendous at.

Nuking OpenSSL was a drive by change since I was already modifying some code to make HTTP optional, so I can back it out if people have problems...I just really hate it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -118,19 +118,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "baf1de4339761588bc0619e3cbc0120ee582ebb74b53b4efbf79117bd2da40fd"
 
 [[package]]
-name = "chrono"
-version = "0.4.19"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "670ad68c9088c2a963aaa298cb369688cf3f9465ce5e2d4ca10e6e0098a1ce73"
-dependencies = [
- "libc",
- "num-integer",
- "num-traits 0.2.14",
- "time 0.1.44",
- "winapi",
-]
-
-[[package]]
 name = "circular"
 version = "0.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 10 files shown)


gabrielesvelto:
Not only this change LGTM but I like the versatility it adds (we'll need it in Firefox too!) and I'm glad to leave OpenSSL behind. @Gankra @Swatinem if you have nothing against this I'll approve it right away (minus fixing the unused dependency in the fuzzer).

luser (approved):
I have no love for OpenSSL myself. Back when I wrote this code I just slapped `reqwest` in there and called it a day. I don't know if there's a state-of-the-art for "allow consumers to choose their TLS implementation" nowadays. I can imagine a project that is already using a specific TLS implementation that wants to integrate with these crates wanting such a thing, but until we hit that point it's not worth overdesigning for.

I'd want to touch base with the Sentry folks to make sure that this doesn't make their lives worse before merging this, at least.

Jake-Shadle:
The state of the art is currently "choose the feature", most of the HTTP client implementations like reqwest and others unfortunately default to openssl (well, technically "native tls") still, which maybe made sense 3+ years ago, but doesn't really make sense these days IMO.

Gankra:
Yeah this seems reasonable to me

Swatinem (approved):
Sentry has its own cache provider, we only use the `SymbolFile` parser directly.

PR #552, opened 2022-05-03
Upgrade debugid to 0.8, uuid to 1.0

jhpratt:


repi:
Would be great to get this in and use the new 1.0 stable `uuid` crate, not that often a crate reaches 1.0 🎉

Swatinem:
Superceded by #559 which also bumped the fuzzing lockfiles

PR #511, opened 2022-03-21
WIP Teach breakpad-symbols to fetch native symbols and run dump_syms

Gankra:
This is so dumb I can't believe it actually works.

Issue #429, opened 2022-02-04
Implement a simple crash address bitflip check


Gankra:
This is a much more constrained version of https://github.com/luser/rust-minidump/issues/204

Gabriele originally described it here: https://bugzilla.mozilla.org/show_bug.cgi?id=1738651

Essentially, the idea is to take the [crash address](https://github.com/luser/rust-minidump/blob/891fb8b1b0c80cf9cea4cff4476e9e75c515d176/minidump-processor/src/process_state.rs#L169) and apply an analysis which will result in us outputting an extra crash_might_be_a_bitflip flag (or whatever):

1. If the crash address points into mapped memory, assume there's no bitflip
2. otherwise, for every bit in the crash address:
3. flip the bit. If it is now null or points into mapped memory, assume it's a bitflip.

This would ideally use UnifiedMemoryInfo so that we're not relying on just the pages we actually have the memory info for.

This can be refined and tuned more based on experimentation and implementation experience.

Gankra:
(I think this would be good for beginners but I haven't thought about it too much. I should really finish and land #263 as an example for adding a new "analysis" to minidump-processor.)

Gankra:
Some potentially interesting refinements/issues:

* The crash address may *either* be an address that was trying to be executed or memory that was trying to be accessed. It's possible we want to treat the two different. 
  * MemoryInfo *does* have information about whether pages are executable, but this is a difficult signal to use in a program with a JIT because best-practice is to enforce W^X on JIT pages. This means functions on the stack may legitimately be "executing" non-executable pages at the time the minidump is generated. It's possible we can work with JIT at crashtime to resolve this but eugh.

* null pointer accesses are organically often "nearly null", which means perfectly normal execution can try to access "null with a bit flipped". i.e. if you load a field at offset 0x8 of a struct through a null pointer, we will actually load 0x8. Possibly this can be handled by conservatively assuming it's not a bitflip if the flipped bit is low (significantly less likely to be accessing exactly null+1MB or whatever).

[...]

PR #558, opened 2022-05-16
WIP use dump_syms as a library and handle native binaries

Gankra:
This is a better version of #511 

It is blocked on https://github.com/mozilla/dump_syms/issues/371

It also does some of the prep work for #429 by adding infra for fetching a binary. However it's currently relying on dump_syms to unCAB mozilla binaries implicitly, so that feature would need to duplicate the logic in dump_syms, probably:

https://github.com/mozilla/dump_syms/blob/c2743d59b5aa321ade698f84b90f86b3d35a1b06/src/utils.rs#L32-L33

(You may also want to change the name that the file is saved to, currently I undo the `exe` -> `ex_` mangling we do, but we might want to distinguish between whether the file is expected to be a cab or not...?)


Gankra:
New commit: I've substantially reworked breakpad-symbols to expose native file fetching APIs and implemented native unCABing of mozilla binaries (behind features). 

I've also significantly reworked breakpad-symbols to be closer to "properly" asyncable. Now both SymbolFiles and file_paths are managed by OnceCells, so multiple tasks could in principle want the same file/SymbolFile, and will naturally block on it. To make things "fully" async we need to:

* use `join` in a bunch of places in breakpad-symbols
* use `join` in minidump-processor for all the thread stackwalks
* use `tokio::fs` instead of std::fs where possible

Swatinem (approved):

PR #559, opened 2022-05-19
deps: Update debugid/uuid dependencies

Swatinem:
This is technically a breaking change, as these types are exposed in public API.


Gankra (approved):
lgtm

PR #576, opened 2022-06-18
Replace the horrible symbol-shims module with a new minidump-symbol-client interface crate

Gankra:
This removes all dependencies that minidump-processor has on breakpad-symbols and subsequently removes all the features from it that were just forwarding breakpad-symbols features. You can now sub in your implementation with no changes to minidump-processor (unless you want it to use it in tests).

minidump-stackwalk now inherits the feature flags and uses breakpad-symbols directly. However now that everything is completely factored out, changing it to a different impl would only take a couple lines!

Some major renames and changes:

* minidump_processor::symbols no longer exists, although it re-exports minidump-symbol-client
* breakpad::Symbolizer => BreakpadSymbolClient
* breakpad::SymbolSupplier => SymbolClientStrategy
* `minidump_processor::symbols::*_supplier` => `SymbolClient::*_client`
  * All clients now take a #[non_exhaustive] args struct for future-compat
  * The 'simple' client has been renamed to 'local'
  * There is now a 'no_client' to make that easier
  * MultiSymbolizer has been completely removed
  * To allow for abstraction but also disabling, BreakpadSymbolClient::http_client will simply panic as unimplemented if the http feature is disabled. This is only true for the trait impl, the inherent method properly disappears.
* FrameWalker is now FrameWalkerCallbacks
* FrameSymbolizer is now FrameSymbolizerCallbacks

This all build on top of my other PR to introduce native files and dump_syms because they're both major refactors that were stuck in my craw.


Gankra:
cc @Jake-Shadle and @Swatinem this is an *enormous* breaking change to a bunch of the crates but the fix is actually relatively minimal, and puts us in a much better place to replace impls. Basically all of the private detail stuff is now completely hidden, so a theoretical symbolic-symbol-client crate """just""" needs to provide one type that implements this one SymbolClient trait.

Gankra:
per the README, a minimal user of minidump-processor now looks like this:

```rust
use minidump::Minidump;
use minidump_processor::ProcessorOptions;
use serde_json::Value;
use breakpad_symbols::{BreakpadSymbolClient, HttpClientArgs};
 
#[tokio::main]
async fn main() -> Result<(), ()> {
    // Read the minidump
    let dump = Minidump::read_path("../testdata/test.dmp").map_err(|_| ())?;
 
    // Configure the symbol client
    let mut client_args = HttpClientArgs::default();
    client_args.symbol_urls = vec![String::from("https://symbols.totallyrealwebsite.org")];
    client_args.symbols_cache = std::env::temp_dir().join("minidump-cache");
 
    // Use ProcessorOptions for detailed processor configuration
    let options = ProcessorOptions::default();
 
    // Specify a symbol supplier (here we're using the most powerful one, the http supplier)
    let symbol_client = BreakpadSymbolClient::http_client(client_args);
 
    let state = minidump_processor::process_minidump_with_options(&dump, &symbol_client, options)
        .await
        .map_err(|_| ())?;
 
    // Write the JSON output to an arbitrary writer (here, a Vec).
    let mut json_output = Vec::new();
    state.print_json(&mut json_output, false).map_err(|_| ())?;
 
    // Now parse it (here parsed into an arbitrary JSON Object for max flexibility).
    let json: Value = serde_json::from_slice(&json_output).map_err(|_| ())?;
 
    // Now read whatever values you want out of it
    if let Some(Value::Number(pid)) = json.get("pid") {
        println!("pid: {}", pid);
    }
 
    Ok(())
}
```

Gankra:
Because everything used here is in minidump-symbol-client, you can make your client trivially swappable by just having this one line typedefing the impl you want:

```rust
type SymbolClientImpl = breakpad_symbols::BreakpadSymbolClient;

let client_args = minidump_processor::LocalClientArgs::default();
...
let symbol_client = SymbolClientImpl::local_client(client_args);
```

This pattern is sprinkled throughout the tests.

Gankra:
Some historical naming inconsistencies like "symbols_cache" vs "symbol_paths" has been inherited by the Args structs. I will change it if anyone cares and has an opinion on the correct naming scheme.

Also none of the "constructors" actually need to be part of the trait, but I wanted to put them there so that the full API surface was defined in the trait for anyone who wanted to swap it out for Real.

Also the no_client constructor could arguably be replace with a dummy SymbolClient impl, which also might be useful for people who dont want to use breakpad_symbols at all. However in the current design it is actually kind of useful to have breakpad_symbols provide it, because then things like stats get updated properly (which would be mildly tedious to fill in for any other dummy "no impl")

Swatinem (commented):
To be completely frank, this looks extremely over-engineered and not desirable from my (or symbolicators) perspective.

The re-export of the complete `minidump-symbol-client` from `minidump-processor` is a red flag for me to begin with. The trait is supposed to be an extension point for minidump-processor, but now it carries a bunch of static constructors that are not being used **from within `minidump-processor`**, which is a clear indication that they should not exist in the first place.
The minidump processing part does not need to concern itself with details which file path stuff is loaded from, or if it comes from http or the local file system.

The interfaces involved in all this are already extremely unwieldy as is. For example, I never understood why `Module`, `FrameSymbolizerCallbacks` or `FrameWalkerCallbacks` are traits at all. The usage of `&dyn` in itself is a bit of a red flag for me as well.

gabrielesvelto:
> The file path stuff is future-proofing for work that @marti4d is planning where we load up the native binaries into a disassembler for deeper analysis or for things like #429.

We don't need to load binaries to implement that, the memory area around the crashing instruction should always be included in minidumps.

Swatinem:
I took a brief look at #558 as well.

What concerns me is that it adds `get_file_path` as new required and undocumented trait method even though it seems to be completely unused in the current implementation?

I also took another look at the existing traits and their methods. Indeed most of them are just being used for mocking in tests. I guess its a matter of personal preference mostly, and I prefer to keep public interfaces simpler and cleaner even if it makes testing a bit more complicated.

As I previously learned, there is another implementation of `FrameSymbolizerCallbacks` just for the purpose of the "does this look like a valid instruction addr" heuristic. For that use-case, I would rather prefer a dedicated trait method (with a default implementation that just says "yeh, yeh, whatever") that makes the intentions clearer.

---

Also I may get way ahead of myself here and open another can of worms saying that the current way of symbolication by usage of `FrameSymbolizerCallbacks` does not really allow for resolving inlinees. I remember bringing up the idea of splitting stack walking and symbolication into two different passes (as we have that now in symbolicator) for various reasons, among them a clear separation of concerns (there is strong evidence to this, as you even have two different file formats [executable, debug files] for these different steps).

Gankra:
> We don't need to load binaries to implement that, the memory area around the crashing instruction should always be included in minidumps.

This is true if you're literally only concerned with the last instruction of the top frame (and only of the crashing thread?). Possible reasons to want the original binary:

* Figuring out the exact offset to apply to each return-address to get the last instruction of every frame.
* Checking if the snapshotted binary matches the original (bitflip check)
* Dissassembling "backwards" in x86 is ambiguous, as some instruction suffixes are themselves valid instructions. Having more context lets you start further back and dissassemble forward until you hit a "synchronization" point

> What concerns me is that it adds get_file_path as new required and undocumented trait method even though it seems to be completely unused in the current implementation?

Yeah it's basically designed to be the lowest common denominator for any more advanced queries. Basically saying that the SymbolClient's job is to do its best to get that file onto your disk somewhere, either by finding it locally, or downloading it, and hey here's the path to that file for whatever you want to do next. We could definitely push all the future logic "into" SymbolClient and drop the ability to ask for the files, but it feels a bit like it's tying the hands of anyone who wants to build on top of these tools to upstream what they want.

i.e. I've been working on cargo-based stuff a lot recently and it's *really* annoying that there isn't a proper API to just make cargo fetch the `.crate` for any package for your own analysis. Cargo *knows* how to do that, but you can only coerce it to fetch packages by putting them in a Cargo.toml/Cargo.lock, which has weird arbitrary limitations. Because of this I've become more inclined to give tools some "plumbing" APIs that are basically just "hey I know you know how to do this already internally, so do this for me and let me do whatever with it".

> Also I may get way ahead of myself here and open another can of worms saying that the current way of symbolication by usage of FrameSymbolizerCallbacks does not really allow for resolving inlinees.

Ah hmm, is the issue that a single $rip may technically be in multiple modules at once, because frames from other modules were inlined into it? Hmm... that's a good point. I guess the solution is to just hand all the module information to the SymbolClient (breakpad-symbols) upfront and doing pure absolute-address queries from minidump-processor, no explicit module lookup allowed?

Swatinem:
Having a rich set of utilities is great, I feel you. Maybe it makes sense to quickly brainstorm what either of us needs from the API and figure something out that fits everyone ;-)

---

Inlinees is not about multiple modules. Its rather that a single intruction addr refers to more than one source file/line. So if your `fill_name` only supports a single source location, that won’t support inlinees.

Gankra:
this has bitrotted to hell and sentry wants a different approach

Gankra:
Are you able/willing to prototype that yourself? I feel like you have a better understanding of what you want.

PR #582, opened 2022-06-27
feat: Implement MIPS unwinding with CFI and stack scanning

Swatinem:
MIPS is a bit annoying because it uses the same CONTEXT type for both 64-bit and 32-bit machines. I had to do some gross type manipulations to get a separate `CpuContext` for the 32-bit variant, since that needs to have a differently-sized `Register` type as that is used in various places.

The other annoyance is that memory access had LE hardcoded. I added a new method specifically for BE, as I didn’t want to introduce a new parameter to the existing function. Also the CfiStackWalker now has the responsibility to fetch register contents from memory depending on its endianness.


minidump-processor/src/stackwalker/mips.rs
```diff
@@ -0,0 +1,396 @@
...
+                Ok(mips32) => {
+                    frame =
+                        get_caller_by_cfi(mips32, callee, grand_callee, stack, modules, syms).await
+                }
+                Err(mips64) => {
+                    frame =
+                        get_caller_by_cfi(mips64, callee, grand_callee, stack, modules, syms).await
+                }
+            }
+        }
```
Gankra, line 283: Frame pointers are unreliable trash on mips?

minidump-processor/src/stackwalker/mips.rs
```diff
@@ -0,0 +1,396 @@
...
+    const MAX_STACK_SIZE: u32 = 1024;
+    const MIN_ARGS: u32 = 4;
+    const POINTER_WIDTH: u32 = 4;
+    trace!("unwind: trying scan");
+    // Stack scanning is just walking from the end of the frame until we encounter
+    // a value on the stack that looks like a pointer into some code (it's an address
+    // in a range covered by one of our modules). If we find such an instruction,
+    // we assume it's an pc value that was pushed by the CALL instruction that created
+    // the current frame. The next frame is then assumed to end just before that
+    // pc value.
```
Gankra, line 118: this comment is just totally wrong from what I can tell. `ra` is saved in the middle of the frame because all your locals come before it.

Gankra: also all references to CALL should be jal

Gankra (changes requested):
hmmm I don't love the endianess stuff that's going on here. Does the minidump header parsing code correctly identify we're big-endian? If so, I feel like it should be recording the endianess and storing that fact in the memory subsystem. Then memory_at_address should just use The Right Endianess.

Also the stack scanning code seems completely incorrect if the diagrams I saw were to be trusted. Apparently locals are pushed before `ra` so finding `ra` kinda tells you nothing about where the next stackframe starts. And if anything else saved should be at a lower address than `ra`?

The 32/64 context stuff is also super whack but idk

Gankra:
the diagram in question, with my own annotations:

Screen Shot 2022-06-30 at 9 18 17 AM

Gankra:
Does your customer minidump use the complex case of scanning, or does it succeed on only the fast heuristic?

Gankra:
I did some more research into what people do to unwind on mips and I am now Angry an understand your pain. jesus christ

minidump-processor/src/stackwalker/mod.rs
```diff
@@ -206,13 +222,10 @@ where
...
-                "unwinding {}",
-                frame
-                    .function_name
-                    .clone()
-                    .unwrap_or_else(|| format!("0x{:08x}", frame.instruction))
-            );
+            match frame.function_name.as_ref() {
+                Some(name) => trace!("unwinding {}", name),
+                None => trace!("unwinding 0x{:016x}", frame.instruction),
+            }
```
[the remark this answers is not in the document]
Swatinem: I was wondering, not sure how these trace macros internally work, if any code is called when you are not subscribed to this log level. Since this is doing string cloning and allocation, I tried to avoid that.

minidump-processor/src/stackwalker/mips.rs
```diff
@@ -0,0 +1,396 @@
...
+        frame.instruction = ip - 8;
+
+        Some(frame)
+    }
+}
+
+/// This is a hack to have a different [`CpuContext`] type/impl depending on the
+/// context flags of the inner [`MipsContext`]
+#[derive(Clone)]
+struct Mips32Context(MipsContext);
```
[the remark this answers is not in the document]
Swatinem: Yes :-( The breakpad C++ code is also littered with a ton of if conditions to make that happen :-(

minidump-processor/src/stackwalker/mips.rs
Swatinem: yes, I don’t think they are being used at all for that purpose.

minidump-processor/src/stackwalker/mips.rs
```diff
@@ -0,0 +1,396 @@
...
+    instruction: Pointer,
+    modules: &MinidumpModuleList,
+    symbol_provider: &P,
+) -> bool
+where
+    P: SymbolProvider + Sync,
+{
+    if instruction < 0x1000 {
+        return false;
+    }
```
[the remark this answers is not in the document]
Swatinem: not really, this is just in line with all the other heuristics.

Swatinem:
> Does the minidump header parsing code correctly identify we're big-endian? If so, I feel like it should be recording the endianess and storing that fact in the memory subsystem. Then memory_at_address should just use The Right Endianess.

Oh absolutely! I completely forgot about that one. This indeed is a much better solution.

I also updated the comments and cleaned up the unwinder logic a little.

Not sure what you mean by "fast heuristic"? Probably that we assume `$ra` is saved at the bottom of the stack? For all the cases I have looked at, the code seems to push/save the `$ra` as the first thing on its stack frame (at the "bottom" depending on your mental model in which direction the stack grows :-D). That is also what the CFI for non-leaf functions says for those functions that do have CFI. See https://github.com/rust-minidump/rust-minidump/issues/583#issue-1286096217 for an example.

It also seems like the customer is running into the same problem I describe in https://swatinem.de/blog/unwind-tables/, namely that only half of the code seems to be covered by unwind tables to begin with :-(
In general, it is so sad, and surprising, to see how many customer binaries are missing CFI for half of their functions.

❤️ thanks so much for the thorough review!

luser:
> It also seems like the customer is running into the same problem I describe in https://swatinem.de/blog/unwind-tables/, namely that only half of the code seems to be covered by unwind tables to begin with :-(
> In general, it is so sad, and surprising, to see how many customer binaries are missing CFI for half of their functions.

Oof. Are they not uploading debug symbols to Sentry? I would think that would be the proper way to address this!

Swatinem:
> Are they not uploading debug symbols to Sentry? I would think that would be the proper way to address this!

They are! The debug symbols they upload only contain CFI for half of whats in them.

Swatinem:
Its not `.sym` files. They have native ELF executables with the appropriate sections.
I suspect its the classic problem of C executables just being a bunch of statically linked thrown together that were all compiled with different compile flags, and some of those may have been build without unwind info. :-( I have seen this problem way more frequently than I would like :-(

minidump-processor/src/stackwalker/mips.rs
```diff
@@ -0,0 +1,388 @@
...
+    let mut count = MAX_STACK_SIZE / POINTER_WIDTH;
+    // In case of mips32 ABI the stack frame of a non-leaf function
+    // must have a minimum stack frame size for 4 arguments (4 words).
+    // Move stack pointer for 4 words to avoid reporting non-existing frames
+    // for all frames except the topmost one.
+    // There is no way of knowing if topmost frame belongs to a leaf or
+    // a non-leaf function.
+    if callee.trust != FrameTrust::Context {
+        last_sp += MIN_ARGS * POINTER_WIDTH;
+        count -= MIN_ARGS;
```
Gankra, line 130: these can overflow/underflow and crash debug builds

minidump-processor/src/stackwalker/mips.rs
```diff
@@ -0,0 +1,388 @@
...
+            // (that's what breakpad does!)
+
+            trace!(
+                "scan seems valid -- caller_pc: 0x{caller_pc:016x}, caller_sp: 0x{caller_sp:016x}"
+            );
+
+            let mut caller_ctx = MipsContext::default();
+            caller_ctx.set_register(
+                PROGRAM_COUNTER,
+                caller_pc as u64 - (2 * POINTER_WIDTH as u64),
```
Gankra, line 152: Wait why are you apply an adjustment to pc here? that's usually the job of "final cleanup" in the outer frame. we store the return-address pc in the context, and `frame.instruction` gets the "adjusted" address we think is the CALL instruction.

minidump-processor/src/stackwalker/mips.rs
```diff
@@ -0,0 +1,388 @@
...
+
+            // Don't do any more validation, and don't try to restore fp
+            // (that's what breakpad does!)
+
+            trace!(
+                "scan seems valid -- caller_pc: 0x{caller_pc:016x}, caller_sp: 0x{caller_sp:016x}"
+            );
+
+            let mut caller_ctx = MipsContext::default();
+            caller_ctx.set_register(PROGRAM_COUNTER, caller_pc - (2 * POINTER_WIDTH));
```
Gankra, line 211: ditto on this adjustment

Gankra (approved):
seems like a decent starting point

although you probably want to write some tests so this stuff doesn't horribly bitrot?

Swatinem:
> although you probably want to write some tests so this stuff doesn't horribly bitrot?

I started trying to convern parts of the minidump I had locally to a test, but gave up halfway through. Not sure I want to hand-craft some stack memory either.

The real world dump I have here produces a "reasonable" stack trace, meaning: It scans a bit and ends up in `libc.so!nanosleep` for which it has CFI, afterwards it again starts to scan which then ends up in `libc.so!start (pthread_create.c)` so thats totally good enough for me.

Kalki70:
Hi!

The strange thing for me is that before we could see the stack trace info. I remember seeing in Sentry specific files and line numbers where the crash occurred. So I guess compiler flags were OK. But about a month ago I needed to see it for a new crash, and it was not possible. After trying many things I then went back to a version of our software from 6 months ago, when it was working, expecting then to narrow down which commit broke everything, and I realized I also could not see stack traces using that old code. If I saw in Sentry the uploaded images, it correctly indicated it was uploaded 6 months ago, but again missing symbols. So I started contacting Sentry guys, in case there had been some change.

Besides, our CI is now broken,  from months ago, because sentry-cli fails to upload symbols, with some error : `error: No such device (os error 19).` and it aborts.  I then found out that this happens since version 1.7.2 of sentry-cli. So, our CI, which uses the latest version, is not even uploading symbols anymore. This is is related to a symbolic link in /etc (/etc/mtab -->/proc/mounts). If I delete that link new sentry-cli versions seem to work, and I can try that in my dev environment.
 
I should see some improment soon then, if I understood correctly.

Kalki70:
Hi, @Swatinem ,

Oh, my... I checked, and buidling with debug info was disabled. I don't know since when. I guess somebody changed it at some point.  I can't see a full stack trace, I guess untill you roll out a new version with your changes, but at least I can see the file and line number that caused the crash.

I wonder if you can check event 46cd14ae700a4f608a0ba92d493f48dc for my organization.

Best regards,

Luis

Swatinem:
> I wonder if you can check event 46cd14ae700a4f608a0ba92d493f48dc for my organization.

Looks like minidump attachment quote is full, so the event does not have a minidump I can download. :-( I’m working through the last couple of steps to deploy this change, then we will see.

> sentry-cli fails to upload symbols, with some error : error: No such device (os error 19). and it aborts.

Yes, we are now better logging and error and are considering to filter these out. In the meantime, give it more targeted directories with binaries in it `/bin`, `/usr`, etc. If you give it the whole root, it will try to recurse into all kinds of directories it won’t find useful stuff anyways.

Kalki70:
@Swatinem ,

I have the minidump for that crash. Is there a way to send it to you, your email address, or maybe you have access to the slack channel with our company?

When that message appears ("no such device"). Sentry-cli aborts and stops searching in other directories. Maybe it shoudl just ignore it? For now I can modify the link : /etc/mtab --> /proc/mounts to a relative one, so it does not check  /etc on the host :  /etc/mtab --> ../proc/mounts and then it succeeds.

Kalki70:
Wonderful! It works, thanks!

PR #587, opened 2022-06-28
Derive Ord on PublicSymbol automatically

mstange:
The derived implementation is a lexicographical order based on the struct field order.


mstange:
> Could you add a comment to the struct describing how the ordering works

I've updated the PR with an added comment on the address field, to say that this field needs to be first.

gabrielesvelto:
LGTM, thanks!

PR #590, opened 2022-06-28
Remove inclusive ranges

### LINKED ISSUE #586: proc maps ranges are not inclusive
Today I discovered the range-map code in this repo.
It uses a `Range` type which models an **inclusive** range.

I don't think the code in this repo is very well served by an inclusive range type. I see lots and lots of occurrences of the pattern "end_address - 1".

Here's a place where we *don't* subtract 1 from the end, and I think that's wrong:

https://github.com/rust-minidump/rust-minidump/blob/fd680d28262bc84bdca6c9e4f64f1d3d937f081b/minidump/src/minidump.rs#L2337-L2343

Searching the web for "proc maps" turns up examples such as this one:

```
7faa721ce000-7faa721cf000 r--p 00000000 08:30 3021                       /usr/lib/locale/C.UTF-8/LC_ADDRESS
7faa721cf000-7faa724b5000 r--p 00000000 08:30 3034                       /usr/lib/locale/locale-archive
7faa724b5000-7faa724da000 r--p 00000000 08:30 11971                      /usr/lib/x86_64-linux-gnu/libc-2.31.so
```

These ranges are consecutive; the end address of one range is equal to the start address of the next range. The ranges don't overlap.

So really I'm suggesting two things in this bug:

 - We should fix the proc maps code, and
 - Inclusive ranges of integers are basically never what you want.

How do people feel about switching to the `rangemap` crate (no dash between "range" and "map")? https://docs.rs/rangemap/

mstange:
Fixes #586.

This PR removes the dependency on range_map, which allows us to stop using inclusive ranges.

All of our ranges describe ranges of bytes in memory; for those, you basically always want half-open ranges of the form `start_address..end_address`.

To replace `range_map::RangeMap`, we use either sorted Vecs, or `rangemap::RangeMap`.
Note the missing underscore.

`rangemap::RangeMap` does not implement PartialEq / Eq, so some of the tests are changed from `assert_eq!` to a match statement.

It also does not support double-ended iterators.

For any RangeMaps which I converted to Vecs, I've changed the way we handle overlapping ranges, to do the following:

```
cur.lines.sort_by_key(|l| l.address);
cur.lines.dedup_by(|next, current| current.end_address() >= next.end_address());
```

This allows overlapping ranges if the next range sticks out from the end of its predecessor. And I've changed the tests to allow either value to be returned in the overlapping part.
Since overlapping ranges are usually a garbage input anyway, I'm hoping that this is an adequate way to deal with the situation. From what I gathered from the history of this code, the way we dealt with overlapping ranges was mostly aimed at avoiding panics from range_map, and not with the goal of any particular output.

There are two case where we care about the specific way we process overlapping ranges is this case, which I've preserved. The first one is about STACK WIN records:

https://github.com/rust-minidump/rust-minidump/blob/e0f8b5a7b4b6d9eebdf297f59c7aaa49c93f8494/breakpad-symbols/src/sym_file/parser.rs#L443-L474

And the second is about duplicated modules, and is handled in this breakpad code: https://chromium.googlesource.com/breakpad/breakpad/+/54fa71efbe50fb2b58096d871575b59e12edba6d/src/processor/minidump.cc#2784
I'm pretty sure I preserved that one too.

Overall, it's not the prettiest change, especially because I'm adding multiple manual calls to binary_search. But at least it's getting rid of a lot of occurrences of the pattern `end_address - 1`.

### DIFF
--- a/breakpad-symbols/src/lib.rs
+++ b/breakpad-symbols/src/lib.rs
@@ -298,17 +298,17 @@ pub enum FileError {
 #[derive(Debug)]
 pub struct FillSymbolError {
     // We don't want to yield a full SymbolError for fill_symbol
-    // as this would involve cloning bulky Error strings every time
-    // someone requested symbols for a missing module.
-    //
-    // As it turns out there's currently no reason to care about *why*
-    // fill_symbol, so for now this is just a dummy type until we have
-    // something to put here.
-    //
-    // The only reason fill_symbol *can* produce an Err is so that
-    // the caller can distinguish between "we had symbols, but this address
-    // didn't map to a function name" and "we had no symbols for that module"
-    // (this is used as a heuristic for stack scanning).
+// as this would involve cloning bulky Error strings every time
+// someone requested symbols for a missing module.
+//
+// As it turns out there's currently no reason to care about *why*
+// fill_symbol, so for now this is just a dummy type until we have
+// something to put here.
+//
+// The only reason fill_symbol *can* produce an Err is so that
+// the caller can distinguish between "we had symbols, but this address
+// didn't map to a function name" and "we had no symbols for that module"
+// (this is used as a heuristic for stack scanning).
 }
 
 impl PartialEq for SymbolError {
@@ -895,10 +895,10 @@ mod test {
 
         let supplier = SimpleSymbolSupplier::new(paths.clone());
         let bad = SimpleModule::default();
-        assert_eq!(
-            supplier.locate_symbols(&bad).await,
-            Err(SymbolError::NotFound)
-        );
+        match supplier.locate_symbols(&bad).await {
+            Err(SymbolError::NotFound) => {}
+            _ => panic!(),
+        }
 
         // Try loading symbols for each of two modules in each of the two
         // search paths.
@@ -920,10 +920,10 @@ mod test {
         {
             let m = SimpleModule::new(file, id);
             // No symbols present yet.
-            assert_eq!(
-                supplier.locate_symbols(&m).await,
-                Err(SymbolError::NotFound)
-            );
+            match supplier.locate_symbols(&m).await {
+                Err(SymbolError::NotFound) => {}
+                _ => panic!(),
+            }
             write_good_symbol_file(&path.join(sym));
             // Should load OK now that it exists.
             assert!(
@@ -937,10 +937,10 @@ mod test {
         let debug_id = DebugId::from_str("ffff0000-0000-0000-0000-abcd12345678-a").unwrap();
         let mal = SimpleModule::new("baz.pdb", debug_id);
         let sym = "baz.pdb/FFFF0000000000000000ABCD12345678a/baz.sym";
-        assert_eq!(
-            supplier.locate_symbols(&mal).await,
-            Err(SymbolError::NotFound)
-        );
+        match supplier.locate_symbols(&mal).await {
+            Err(SymbolError::NotFound) => {}
+            _ => panic!(),
+        }
         write_bad_symbol_file(&paths[0].join(sym));
         let res = supplier.locate_symbols(&mal).await;
         assert!(
--- a/breakpad-symbols/src/sym_file/parser.rs
+++ b/breakpad-symbols/src/sym_file/parser.rs
@@ -3,18 +3,29 @@
 
 use nom::IResult::*;
 use nom::*;
-use range_map::{Range, RangeMap};
 use tracing::warn;
 
 use std::collections::HashMap;
 use std::fmt::Debug;
+use std::ops::Range;
 use std::str;
 use std::str::FromStr;
 
-use minidump_common::traits::IntoRangeMapSafe;
-
 use crate::sym_file::types::*;
 use crate::SymbolError;
+
+trait RangeIntersects {
+    fn intersects(&self, other: &Self) -> bool;
+}
+
+impl<T: Ord> RangeIntersects for Range<T> {
+    fn intersects(&self, other: &Self) -> bool {
+        self.start < self.end
+            && other.start < other.end
+            && self.start <= other.end
+            && other.start < self.end
+    }
+}
 
 #[derive(Debug)]
 enum Line {
@@ -158,7 +169,7 @@ chain!(
             size,
             parameter_size,
             name: name.to_string(),
-            lines: RangeMap::new(),
+            lines: Vec::new(),
         }
     }
     ));
@@ -302,7 +313,7 @@ pub struct SymbolParser {
     // When building a RangeMap when need to sort an array of this
     // format anyway, so we might as well construct it directly and
     // save a giant allocation+copy.
-    functions: Vec<(Range<u64>, Function)>,
+    functions: Vec<Function>,
     cfi_stack_info: Vec<(Range<u64>, StackInfoCfi)>,
     win_stack_framedata_info: Vec<(Range<u64>, StackInfoWin)>,
     win_stack_fpo_info: Vec<(Range<u64>, StackInfoWin)>,
@@ -510,21 +521,18 @@ impl SymbolParser {
             Line::Function(mut cur, lines) => {
                 cur.lines = lines
                     .into_iter()
-                    .map(|l| {
+                    .filter(|l| {
                         // Line data from PDB files often has a zero-size line entry, so just
                         // filter those out.
-                        if l.size > 0 {
-                            if let Some(end) = l.address.checked_add(l.size as u64 - 1) {
-                                return (Some(Range::new(l.address, end)), l);
-                            }
-                        }
-
-                        (None, l)
+                        l.size > 0
                     })
-                    .into_rangemap_safe();
+                    .collect();
+                cur.lines.sort_by_key(|l| l.address);
+                cur.lines
+                    .dedup_by(|next, current| current.end_address() >= next.end_address());
 
-                if let Some(range) = cur.memory_range() {
-                    self.functions.push((range, cur));
+                if cur.size != 0 {
+                    self.functions.push(cur);
                 }
             }
             Line::StackCfi(mut cur) => {
@@ -550,14 +558,17 @@ impl SymbolParser {
 
         // Now sort everything and bundle it up in its final format.
         self.publics.sort();
+        self.functions.sort_by_key(|f| f.address);
+        self.functions
+            .dedup_by(|next, current| current.end_address() >= next.end_address());
 
         SymbolFile {
             files: self.files,
             publics: self.publics,
-            functions: into_rangemap_safe(self.functions),
-            cfi_stack_info: into_rangemap_safe(self.cfi_stack_info),
-            win_stack_framedata_info: into_rangemap_safe(self.win_stack_framedata_info),
-            win_stack_fpo_info: into_rangemap_safe(self.win_stack_fpo_info),
+            functions: self.functions,
+            cfi_stack_info: self.cfi_stack_info.into_iter().collect(),
+            win_stack_framedata_info: self.win_stack_framedata_info.into_iter().collect(),
+            win_stack_fpo_info: self.win_stack_fpo_info.into_iter().collect(),
             // Will get filled in by the caller
             url: self.url,
             ambiguities_repaired: 0,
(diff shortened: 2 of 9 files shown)


breakpad-symbols/src/lib.rs
```diff
@@ -895,10 +895,10 @@ mod test {
...
         let supplier = SimpleSymbolSupplier::new(paths.clone());
         let bad = SimpleModule::default();
-        assert_eq!(
-            supplier.locate_symbols(&bad).await,
-            Err(SymbolError::NotFound)
-        );
+        match supplier.locate_symbols(&bad).await {
+            Err(SymbolError::NotFound) => {}
+            _ => panic!(),
+        }
```
Gankra, line 901: these might be clearer as `assert!(matches!(expr, Err(SymbolError::NotFound))`

breakpad-symbols/src/sym_file/parser.rs
```diff
@@ -3,19 +3,31 @@
...
+trait RangeIntersects {
+    fn intersects(&self, other: &Self) -> bool;
+}
+
+impl<T: Ord> RangeIntersects for Range<T> {
+    fn intersects(&self, other: &Self) -> bool {
+        self.start < self.end
+            && other.start < other.end
+            && self.start <= other.end
+            && other.start < self.end
```
Gankra, line 26: I think this merits a bit of a comment. The first two checks just seem to be "this isn't an empty (or inverted) range" but I'm a bit confused by why one of the comparisons is <=...

breakpad-symbols/src/sym_file/parser.rs
```diff
@@ -510,21 +521,18 @@ impl SymbolParser {
...
                         // Line data from PDB files often has a zero-size line entry, so just
                         // filter those out.
-                        if l.size > 0 {
-                            if let Some(end) = l.address.checked_add(l.size as u64 - 1) {
-                                return (Some(Range::new(l.address, end)), l);
-                            }
-                        }
-
-                        (None, l)
+                        l.size > 0
```
Gankra, line 527: I was like "why the fuck did I write this like this" and tried to follow the blame and I think this is just ancient code that survived a million refactors.

breakpad-symbols/src/sym_file/parser.rs
```diff
@@ -1111,39 +1078,40 @@ FUNC 1001 10 10 some func overlap contained
...
-        .ranges_values()
-        .map(|&(_, ref f)| f)
-        .collect::<Vec<_>>();
+    assert_eq!(sym.functions.len(), 2);
+    assert!(sym
+        .function_for_address(0x1000)
+        .unwrap()
+        .name
+        .starts_with("some func"));
+    assert_eq!(sym.function_for_address(0x1030).unwrap().address, 0x1001);
```
Gankra, line 1087: ! what's up with this

breakpad-symbols/src/sym_file/parser.rs
```diff
@@ -1111,39 +1078,40 @@ FUNC 1001 10 10 some func overlap contained
...
+                    size: 0x10,
+                    file: 0,
+                    line: 42,
+                },
+                SourceLine {
+                    address: 0x1001,
+                    size: 0x10,
+                    file: 0,
+                    line: 44,
+                },
```
Gankra, line 1108: ditto

breakpad-symbols/src/sym_file/parser.rs
```diff
@@ -1286,7 +1254,7 @@ STACK WIN 4 8d93e 4 4 0 0 10 0 0 1 prog string
 fn address_size_overflow() {
     let bytes = b"FUNC 1 2 3 x\nffffffffffffffff 2 0 0\n";
     let sym = parse_symbol_bytes(bytes.as_slice()).unwrap();
-    let fun = sym.functions.get(1).unwrap();
-    assert!(fun.lines.is_empty());
+    let fun = sym.function_for_address(1).unwrap();
+    assert_eq!(fun.line_for_address(0x1234), None);
```
Gankra, line 1258: ? why the change?

breakpad-symbols/src/sym_file/types.rs
```diff
@@ -33,6 +34,12 @@ pub struct SourceLine {
     pub line: u32,
 }
 
+impl SourceLine {
+    pub fn end_address(&self) -> u64 {
+        self.address.saturating_add(self.size.into())
+    }
+}
```
Gankra, line 41: hmm saturating always feels like a trap, why do you think this is better than discarding the entry?

breakpad-symbols/src/sym_file/types.rs
```diff
@@ -45,18 +52,34 @@ pub struct Function {
...
         }
-        Some(Range::new(
-            self.address,
-            self.address.checked_add(self.size as u64)? - 1,
-        ))
+        Some(self.address..self.end_address())
+    }
+
+    pub fn line_for_address(&self, address: u64) -> Option<&SourceLine> {
+        match self.lines.binary_search_by_key(&address, |l| l.address) {
```
Gankra, line 71: I would prefer if you could transplant the comment you deleted, binary-search is hell and always merits comments.

(edit: oh wait you added some to the other instance, this probably still merits them though)

breakpad-symbols/src/sym_file/types.rs
```diff
@@ -96,10 +119,7 @@ impl StackInfoCfi {
         if self.size == 0 {
             return None;
         }
-        Some(Range::new(
-            self.init.address,
-            self.init.address.checked_add(self.size as u64)? - 1,
-        ))
+        Some(self.init.address..self.init.address.checked_add(self.size as u64)?)
```
Gankra, line 122: this one stays checked while all the others saturate?

breakpad-symbols/src/sym_file/types.rs
```diff
@@ -149,22 +169,19 @@ impl StackInfoWin {
         if self.size == 0 {
             return None;
         }
-        Some(Range::new(
-            self.address,
-            self.address.checked_add(self.size as u64)? - 1,
-        ))
+        Some(self.address..self.address.checked_add(self.size as u64)?)
```
Gankra, line 172: Ah another case where you check still

breakpad-symbols/src/sym_file/types.rs
```diff
@@ -222,3 +239,27 @@ pub struct SymbolFile {
     /// still be correct.
     pub cfi_eval_corruptions: u64,
 }
+
+impl SymbolFile {
+    /// Returns the closest function to the address for which `function.address <= address`.
+    pub fn find_nearest_function(&self, address: u64) -> Option<&Function> {
```
Gankra, line 245: Feels like this wants a better name. `find_nearest_function_before`?

Gankra (changes requested):
Reviewed everything except minidump.rs, taking a break, this is a huge diff @-@

breakpad-symbols/src/sym_file/parser.rs
mstange: Yeah this is probably the most major change in behavior. We now allow overlapping functions if they aren't fully subsumed. And for the addresses in the overlapping parts, we don't care which function is returned. In the first test, all of the candidates begin with "some func". And the second one is testing that the function which starts at 0x1001 and sticks out the end from the previous one actually covers the byte at 0x1030.

breakpad-symbols/src/lib.rs
mstange: Oh I didn't realize `matches!` was a thing! All of its friends (e.g. assert_matches) seemed to be unstable so I didn't look further.

breakpad-symbols/src/sym_file/parser.rs
mstange: You caught a bug, the <= is wrong. I was following [our rectangle intersection code](https://searchfox.org/mozilla-central/rev/0069a0cdc3cd136ea7411521219df6665f260be7/gfx/2d/BaseRect.h#105-107) and still screwed it up.

breakpad-symbols/src/sym_file/types.rs
mstange: I don't have strong opinions on this. I used saturating simply because then I don't have to deal with None.
We may produce garbage in the bad cases but at least we won't panic.

breakpad-symbols/src/sym_file/types.rs
mstange: ok, will make sure there are some comments at every instance of this pattern

breakpad-symbols/src/sym_file/types.rs
mstange: Hmm, true, it's inconsistent that I kept all the methods which return ranges to use checked, and then used saturate in `end_address()`. The methods which return a range want to return None if the range is empty or inverted, so using checked is easier there.

breakpad-symbols/src/sym_file/types.rs
mstange: Sure, sounds good.

breakpad-symbols/src/sym_file/types.rs
Gankra: My default calculus here is that one Bad input can spoil the batch if we keep it around, but discarding a sketchy input (as in discarding the line) is unlikely to even be noticed, since we have millions of lines and you're gonna query like 20 of them. The places where I go out of my way to preserve sketchy inputs are places where it was an actual problem in production.

mstange: Filtering them out sounds good to me. But we use the same type for unfiltered and filtered line records. Should we return Option from end_address or not?

minidump/src/minidump.rs
```diff
@@ -1397,19 +1384,23 @@ impl MinidumpModuleList {
...
-            .enumerate()
-            .map(|(i, module)| (module.memory_range(), i))
-            .into_rangemap_safe();
+        let mut sorted_modules: Vec<usize> = (0..modules.len()).collect();
+        sorted_modules.sort_by_key(|i| modules[*i].base_address());
+        sorted_modules.dedup_by(|ni, ci| {
+            let cur = &modules[*ci];
+            let next = &modules[*ni];
+            let cur_end_address = cur.base_address().saturating_add(cur.size());
+            let next_end_address = next.base_address().saturating_add(next.size());
```
Gankra, line 1398: More saturating that should probably be checked. This one's harder to remove though, so maybe it's fine?

minidump/src/minidump.rs
```diff
@@ -1426,9 +1417,22 @@ impl MinidumpModuleList {
...
+            .sorted_modules
+            .binary_search_by_key(&address, |i| self.modules[*i].base_address())
+        {
+            Ok(i) => &self.modules[self.sorted_modules[i]],
+            Err(0) => return None,
+            Err(i) => &self.modules[self.sorted_modules[i - 1]],
+        };
+        let end_address = nearest_module
+            .base_address()
+            .saturating_add(nearest_module.size());
```
Gankra, line 1430: in theory this one should be validated on first digest so it's fine for it to panic..?

minidump/src/minidump.rs
```diff
@@ -1426,9 +1417,22 @@ impl MinidumpModuleList {
 
     /// Return a `MinidumpModule` whose address range covers `address`.
     pub fn module_at_address(&self, address: u64) -> Option<&MinidumpModule> {
-        self.modules_by_addr
-            .get(address)
-            .map(|&index| &self.modules[index])
+        let nearest_module = match self
+            .sorted_modules
+            .binary_search_by_key(&address, |i| self.modules[*i].base_address())
```
Gankra, line 1422: wait do we really get weird overlapping modules? gross that this can't use rangemap :(

minidump/src/minidump.rs
```diff
@@ -1721,22 +1720,26 @@ impl<'mdmp, Descriptor> MinidumpMemoryListBase<'mdmp, Descriptor> {
...
-            .enumerate()
-            .map(|(i, region)| (region.memory_range(), i))
-            .into_rangemap_safe();
+        let mut sorted_regions: Vec<usize> = (0..regions.len()).collect();
+        sorted_regions.sort_by_key(|i| regions[*i].base_address);
+        sorted_regions.dedup_by(|ni, ci| {
+            let cur = &regions[*ci];
+            let next = &regions[*ni];
+            let cur_end_address = cur.base_address.saturating_add(cur.size);
+            let next_end_address = next.base_address.saturating_add(next.size);
```
Gankra, line 1737: more saturating that's annoying to remove

minidump/src/minidump.rs
```diff
@@ -1745,9 +1748,22 @@ impl<'mdmp, Descriptor> MinidumpMemoryListBase<'mdmp, Descriptor> {
...
+            Err(0) => return None,
+            Err(i) => &self.regions[self.sorted_regions[i - 1]],
+        };
+        let end_address = nearest_region
+            .base_address
+            .saturating_add(nearest_region.size);
+        if address < end_address {
+            Some(nearest_region)
+        } else {
+            None
```
Gankra, line 1765: geez this pattern gets repeated a lot :(

Gankra (changes requested):
Ok review of minidump.rs... not a huge fan of how much manual binary searching is going on here.

minidump/src/minidump.rs
Gankra: (the binary searching)

Gankra:
we should add a counter for "times people have tried to remove range-map and given up" D:

PR #599, opened 2022-07-01
Add support for parsing INLINE_ORIGIN and INLINE.

mstange:
This parses the new format but produces the same output as before.
So with this patch alone, we will not get any inline information in the
output.
But this patch allows us to switch to the new sym format without
breaking anything (hopefully).

In `fill_symbol` we now call `get_outermost_source_location`, which has
to look at the inline information, because just by looking at the lines
you would not get the same output as before with the new sym files.
That's because, when inline information is present in sym files, the
line records now describe the lines *at the deepest level of inlining*.
In sym files without inline information, the line records describe the
lines at the outermost level.

### DIFF
--- a/breakpad-symbols/src/sym_file/mod.rs
+++ b/breakpad-symbols/src/sym_file/mod.rs
@@ -362,11 +362,11 @@ impl SymbolFile {
                 parameter_size,
             );
             // See if there's source line info as well.
-            func.lines.get(addr).map(|line| {
-                self.files.get(&line.file).map(|file| {
-                    frame.set_source_file(file, line.line, line.address + module.base_address());
-                })
-            });
+            if let Some((file_id, line, line_address)) = func.get_outermost_sourceloc(addr) {
+                if let Some(file) = self.files.get(&file_id) {
+                    frame.set_source_file(file, line, line_address + module.base_address());
+                }
+            }
         } else if let Some(public) = self.find_nearest_public(addr) {
             // We couldn't find a valid FUNC record, but we could find a PUBLIC record.
             // Unfortauntely, PUBLIC records don't have end-points, so this could be
--- a/breakpad-symbols/src/sym_file/types.rs
+++ b/breakpad-symbols/src/sym_file/types.rs
@@ -33,6 +33,25 @@ pub struct SourceLine {
     pub line: u32,
 }
 
+/// A single range which is covered by an inlined function call.
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
+pub struct Inlinee {
+    /// The depth of the inline call.
+    pub depth: u32,
+    /// The start address relative to the module's load address.
+    pub address: u64,
+    /// The size of this range of instructions in bytes.
+    pub size: u32,
+    /// The source file which contains the function call.
+    ///
+    /// This is an index into `SymbolFile::files`.
+    pub call_file: u32,
+    /// The line number in `call_file` for the function call.
+    pub call_line: u32,
+    /// The function name, as an index into `SymbolFile::inline_origins`.
+    pub origin_id: u32,
+}
+
 /// A source-language function.
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct Function {
@@ -46,6 +65,13 @@ pub struct Function {
     pub name: String,
     /// Source line information for this function.
     pub lines: RangeMap<u64, SourceLine>,
+    /// Inlinee information for this function, sorted by (depth, address).
+    ///
+    /// Essentially this can be considered as "one vec per depth", just with
+    /// all those vecs concatenated into one.
+    ///
+    /// Inlinees form a nested structure, you can think of them like a flame graph.
+    pub inlinees: Vec<Inlinee>,
 }
 
 impl Function {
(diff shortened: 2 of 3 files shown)


mstange:
I'm leaving the PR in draft state just because I want to write more tests. But it's already ready for review.

breakpad-symbols/src/sym_file/mod.rs
```diff
@@ -362,11 +362,11 @@ impl SymbolFile {
...
             );
             // See if there's source line info as well.
-            func.lines.get(addr).map(|line| {
-                self.files.get(&line.file).map(|file| {
-                    frame.set_source_file(file, line.line, line.address + module.base_address());
-                })
-            });
+            if let Some((file_id, line, line_address)) = func.get_outermost_source_location(addr) {
+                if let Some(file) = self.files.get(&file_id) {
+                    frame.set_source_file(file, line, line_address + module.base_address());
```
Gankra, line 367: hmm just noticed this should probably be a checked_add (pre-existing issue)

breakpad-symbols/src/sym_file/types.rs
```diff
@@ -46,6 +65,8 @@ pub struct Function {
     pub name: String,
     /// Source line information for this function.
     pub lines: RangeMap<u64, SourceLine>,
+    /// Inlinee information for this function, sorted by (depth, address).
+    pub inlinees: Vec<Inlinee>,
```
Gankra, line 69: I guess it probably doesn't matter because sorting is sorting, but it feels deeply bizarre to sort by depth more significantly than address? Surely address is the "significant" thing and depth is the "less significant" thing? Does this do something magical?

breakpad-symbols/src/sym_file/types.rs
```diff
@@ -58,6 +79,56 @@ impl Function {
...
+        Some((line.file, line.line, line.address))
+    }
+
+    /// Returns `(file_id, line, address)` of the line record that covers the
+    /// given address. Line records describe locations at the innermost level
+    /// of inline calls at that address.
+    pub fn get_innermost_source_location(&self, addr: u64) -> Option<(u32, u32, u64)> {
+        let line = self.lines.get(addr)?;
+        Some((line.file, line.line, line.address))
+    }
```
Gankra, line 100: Is there a specific purpose for finding the innermost?

Gankra (approved):
Seems solid. I'm kinda punting on checking your work on the inlinee format itself, though. Just checking if the code seems coherent for breakpad-symbols.

breakpad-symbols/src/sym_file/types.rs
mstange: Haha yes, this definitely deserves a comment. It threw me off too when I implemented the binary search over inlines in the addr2line crate.

Here's an example where sorting by `(address, depth)` would take you off-track:

```
inlinees: [
  { address: 0x0, size: 0x10, depth: 0 },
  { address: 0x0, size: 0x2, depth: 1 },
  { address: 0x6, size: 0x2, depth: 1 },
]

Look up { address: 0x7, depth: 0 }
```

In that example, if you sort by `(address, depth)`, your `Err(i) => pick item at i - 1` case will make you end up at an inlinee with depth 1.

breakpad-symbols/src/sym_file/types.rs
mstange: Not at the moment, but there will be once we build the inline stack. Usually the way it works is that you build the inline stack from "outside to inside" by getting inlines at depth k for k in 0.. until there is no inline at that depth, and then you get the "innermost source location" for the final frame's location.
We now have a public method for getting the inline, so it felt natural to add another public method to get the innermost line as well. But I can leave it out until we have a use for it.

breakpad-symbols/src/sym_file/types.rs
Gankra: Ohh it's like a flame graph, the lower frames can actually span a wider range. Interesting. Ok so really this is logically a `Vec<Vec<Inlinee>>`, where each element of the outer vec is "the different depths" and the inner vec is "all the frames at this depth". You're really only trying to binary search the inner vec, but you've opted to flatten everything so the binary search can still accidentally peek into the next/previous depth, but you have guards in place to catch that.

breakpad-symbols/src/sym_file/types.rs
Gankra: No it's fine if it has an eventual purpose. Is the key point here that `get_inlinee_at_depth` can never actually yield the value of `innermost_source_location`, so you need to carve that our a final step?

breakpad-symbols/src/sym_file/types.rs
mstange: That's exactly right! Those are great ways to conceptualize it which I hadn't thought of before. I had treated inline information as "a nested structure covering address ranges", but the flame graph comparison works really well, thanks for that idea.

I've tweaked the comment to say that this can be thought of as "one vec per depth" (but concatenated into one), and to mention the flame graph conceptual model.

breakpad-symbols/src/sym_file/types.rs
mstange: That's right.

PR #601, opened 2022-07-01
[proof of concept] concurrent stackwalking

Gankra:
This is a proof-of-concept for:

1. making breakpad-symbols more properly async by reworking the streaming parser to be more stateful and not in charge of the http Response and only* using tokio::fs.
2. making minidump-processor `join` all the stackwalks

Commit 1 is the bulk of the work, Commit 2 is just a quick and dirty hack.

The net result is that all stackwalking is now concurrent, but not parallel. hot stackwalks take about the same amount of time, because file i/o is ultimately blocking, while cold (http) stackwalks on firefox take about half the time from being able to download things in parallel.

Subsequent comments will include some notes/numbers.

### DIFF
--- a/breakpad-symbols/src/http.rs
+++ b/breakpad-symbols/src/http.rs
@@ -25,6 +25,10 @@ type FileKey = (ModuleKey, FileKind);
 ///
 /// See [`crate::breakpad_sym_lookup`] for details on how paths are searched.
 pub struct HttpSymbolSupplier {
+    inner: Arc<HttpSymbolSupplierInner>,
+}
+
+pub struct HttpSymbolSupplierInner {
     /// File paths that are known to be in the cache
     #[allow(clippy::type_complexity)]
     cached_file_paths: Mutex<HashMap<FileKey, CachedOperation<(PathBuf, Option<Url>), FileError>>>,
@@ -76,74 +80,15 @@ impl HttpSymbolSupplier {
         local_paths.push(cache.clone());
         let local = SimpleSymbolSupplier::new(local_paths);
         let cached_file_paths = Mutex::default();
-        HttpSymbolSupplier {
+        let inner = Arc::new(HttpSymbolSupplierInner {
             client,
             cached_file_paths,
             urls,
             local,
             cache,
             tmp,
-        }
-    }
-
-    #[tracing::instrument(level = "trace", skip(self, module), fields(module = crate::basename(&*module.code_file())))]
-    pub async fn locate_file_internal(
-        &self,
-        module: &(dyn Module + Sync),
-        file_kind: FileKind,
-    ) -> Result<(PathBuf, Option<Url>), FileError> {
-        let k = file_key(module, file_kind);
-        let file_once = self
-            .cached_file_paths
-            .lock()
-            .unwrap()
-            .entry(k)
-            .or_default()
-            .clone();
-        file_once
-            .get_or_init(|| async {
-                // First look for the file in the cache
-                if let Ok(path) = self.local.locate_file(module, file_kind).await {
-                    return Ok((path, None));
-                }
-
-                // Then try to download the file
-                // FIXME: if we try to parallelize this with `join` then if we have multiple hits
-                // we'll end up downloading all of them at once and having them race to write into
-                // the cache... is that ok? Maybe? Since only one will ever win the swap, and it's
-                // unlikely to get multiple hits... this might actually be ok!
-                if let Some(lookup) = lookup(module, file_kind) {
-                    for url in &self.urls {
-                        let fetch =
-                            fetch_lookup(&self.client, url, &lookup, &self.cache, &self.tmp).await;
-
-                        if let Ok((path, url)) = fetch {
-                            return Ok((path, url));
-                        }
-                    }
-
-                    // If we're allowed to look for mozilla's special CAB paths, do that
-                    if cfg!(feature = "mozilla_cab_symbols") {
-                        for url in &self.urls {
-                            let fetch = fetch_cab_lookup(
-                                &self.client,
-                                url,
-                                &lookup,
-                                &self.cache,
-                                &self.tmp,
-                            )
-                            .await;
-
-                            if let Ok((path, url)) = fetch {
-                                return Ok((path, url));
-                            }
-                        }
-                    }
-                }
-                Err(FileError::NotFound)
-            })
-            .await
-            .clone()
+        });
+        Self { inner }
     }
 }
 
@@ -224,7 +169,7 @@ async fn commit_cache_file(mut download: PartialDownload, url: &Url) -> io::Resu
 async fn fetch_symbol_file(
     client: &Client,
     base_url: &Url,
-    module: &(dyn Module + Sync),
+    module: Arc<SimpleModule>,
     cache: &Path,
     _tmp: &Path,
 ) -> Result<SymbolFile, SymbolError> {
@@ -239,7 +184,7 @@ async fn fetch_symbol_file(
     // give up on caching but let the parse+download continue.
 
     // First try to GET the file from a server
-    let sym_lookup = breakpad_sym_lookup(module).ok_or(SymbolError::MissingDebugFileOrId)?;
+    let sym_lookup = breakpad_sym_lookup(&*module).ok_or(SymbolError::MissingDebugFileOrId)?;
     let mut url = base_url
         .join(&sym_lookup.server_rel)
         .map_err(|_| SymbolError::NotFound)?;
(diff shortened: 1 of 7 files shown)


breakpad-symbols/src/http.rs
```diff
@@ -142,46 +151,70 @@ fn file_key(module: &(dyn Module + Sync), file_kind: FileKind) -> FileKey {
...
         )
     })?;
-    fs::create_dir_all(&base)?;
-
-    NamedTempFile::new_in(tmp_path)
+    fs::create_dir_all(&base).await?;
+
+    // Try to create a new file in the final destination with `.part` added to the extension.
+    // Use an attempt counter to vaguely allow multiple processes to race on this, without them
+    // having to cooperate in any way. If there's only one instance this will just work reasonably.
```
Gankra, line 166: So historically we've used this pattern of creating a tempfile somewhere else and swapping it in. This is problematic for a few reasons:

* The tempfile crate is not integrated with tokio::async
* Moving files between directories has the potential to run into "different file system" problems
* There is no signal between separate processes that a file is already being downloaded, so they will race to download and swap in the same file

This new approach is more similar to how e.g. Firefox downloads a file, in that it's created where it wants to go with a special suffix and then renamed when done. Now *ideally* separate processes could use this to detect an ongoing download and block/collaborate on this, but that has some complexities I am not in the mood to sort out right now. In the ideal version of this design we move to an architecture where you just spawn one instance on a server and it knows how to use all the system's parallelism!

(I think you could maybe use os file locking in some clever way, idk)

So for now if a .part is detected, this implementation just repeatedly tries different names until it gives up. This isn't really shippable.

breakpad-symbols/src/http.rs
```diff
@@ -142,46 +151,70 @@ fn file_key(module: &(dyn Module + Sync), file_kind: FileKind) -> FileKey {
...
 
     // INFO URL allows us to properly report the url we retrieved a symbol file
     // from, even when the file is loaded from our on-disk cache.
     let cache_metadata = format!("INFO URL {}\n", url);
-    temp.write_all(cache_metadata.as_bytes())?;
-
-    // TODO: don't do this
-    if final_path.exists() {
-        fs::remove_file(final_path)?;
-    }
```
Gankra, line 180: I genuinely forget why this exists. I have a feeling I was corrupting files and then all the caching machinery assumed the files weren't corrupt, so I just wanted existing files to get blown away..? But that doesn't really make sense because at this point we already discovered the file wasn't there... This is a super weird path.

breakpad-symbols/src/http.rs
```diff
@@ -375,12 +416,12 @@ pub fn unpack_cabinet_file(
...
-    std::io::copy(&mut reader, &mut temp)?;
-
+    let mut temp = create_cache_file(&final_cache_path).await?;
+    let mut std_file = temp.file.into_std().await;
+    std::io::copy(&mut reader, &mut std_file)?;
+    temp.file = fs::File::from_std(std_file);
     // And swap it into the cache
-    temp.persist_noclobber(&final_cache_path)
-        .map_err(|e| std::io::Error::new(std::io::ErrorKind::Other, e))?;
+    fs::rename(&temp.temp_path, &temp.dest_path).await?;
```
Gankra, line 424: This code sucks and isn't properly async because `cab` returns its own Reader that isn't async. It's in a disabled-by-default-even-in-minidump-stackwalk path though, so it doesn't affect my current numbers.

breakpad-symbols/src/sym_file/mod.rs
```diff
@@ -69,267 +69,182 @@ pub mod walker;
...
+pub struct StreamingSymbolParser {
+    buf: circular::Buffer,
+    parser: SymbolParser,
+
+    fully_consumed: bool,
+    tried_to_grow: bool,
+    in_panic_recovery: bool,
+    just_finished_recovering: bool,
+    total_consumed: u64,
+}
```
Gankra, line 81: All of this stuff is just the existing parse_async code but stateful and with an `add_input` and `finish` method so that an external source can handle getting bytes and drive the streaming parser, without it actually being async. This means all the sync and async paths are now unified and all the callback crap is now gone.

breakpad-symbols/src/sym_file/mod.rs
```diff
@@ -69,267 +69,182 @@ pub mod walker;
...
+        let mut buf = vec![0; 1024 * 8];
+        let mut parser = StreamingSymbolParser::new();
+        loop {
+            let n = file.read(&mut buf[..]).await?;
+            if n == 0 {
+                break;
+            }
+            parser.add_input(&buf[..n])?;
+        }
+        parser.finish()
```
Gankra, line 247: Here's a minimal example of driving the new streaming parser with a random async source.

minidump-processor/src/processor.rs
```diff
@@ -356,93 +356,110 @@ where
...
-            }
-        }
+    let requesting_thread = std::sync::Mutex::new(None);
+
+    let threads;
+    {
+        let dump_system_info = &dump_system_info;
+        let misc_info = &misc_info;
+        let exception_context = &exception_context;
+        let memory_list = &memory_list;
```
Gankra, line 366: All of this sucks. We need to capture a few pieces of state by-value, so we have to explicitly reference-ify all this state. This does however indicate the crapton of state that the stackwalker borrows, which gets in the way of actually spawning this loop as separate tasks (thread::scoped, on the other hand, would work fine...)

minidump-processor/src/processor.rs
```diff
@@ -356,93 +356,110 @@ where
...
+        let options = &options;
+        let requesting_thread = &requesting_thread;
+        threads = futures_util::future::join_all(thread_list.threads.iter().enumerate().map(
+            |(i, thread)| async move {
+                let id = thread.raw.thread_id;
+
+                // If this is the thread that wrote the dump, skip processing it.
+                if dump_thread_id.is_some() && dump_thread_id.unwrap() == id {
+                    return CallStack::with_info(id, CallStackInfo::DumpThreadSkipped);
+                }
```
Gankra, line 381: The loop just gets transformed into a join_all that returns CallStacks

minidump-processor/src/processor.rs
```diff
@@ -356,93 +356,110 @@ where
...
+                let thread_context = thread.context(&dump_system_info, misc_info.as_ref());
+                // If this thread requested the dump then try to use the exception
+                // context if it exists. (prefer the exception stream's thread id over
+                // the breakpad info stream's thread id.)
+                let context = if crashing_thread_id
+                    .or(requesting_thread_id)
+                    .map(|id| id == thread.raw.thread_id)
+                    .unwrap_or(false)
+                {
+                    *requesting_thread.lock().unwrap() = Some(i);
```
Gankra, line 392: This is a goofy hack to just modify the loop as little as possible

Gankra:
Symbolizer is ultimately pretty self-contained when it comes to acquiring a SymbolFile for a module, and that's where all the hard i/o happens, so I expect the next lowest-hanging fruit is to `spawn` actual parallel tasks somewhere around the `locate_symbols` boundary. This would require wrapping a few pieces of state in an Arc, which should actually be minimally intrusive, since both HttpSymbolSupplier and SimpleSymbolSupplier already own all their state and work with shared references. The only mutable state *any* of them has is the cached_file_paths in http for downloading native binaries, which is already setup for parallelism (wrapped in a Mutex).

Gankra:
I'm not sure if/when I'll have more time to work on this, I just had an epiphany on how to disentangle the parser and wanted to put this up.

Gankra:
possibly of interest to @Swatinem and @mstange who are thinking about perf/parsing. The new approach doesn't actually touch the "core" parser at all, so it's still completely swappable.

Gankra:
new commit: very naively wraps the SymbolSuppliers and Module in Arcs and spawns a task when starting those queries. In principle this makes Symbol loading/downloading/parsing truly parallel, although obviously the subtask for each thread is still totally blocked on this. I also still don't join on potential parallel paths within the symbol parser (i.e. trying every url in parallel).

# Benchmark Summary

* Hot (local): 1573ms -> 1462ms
* Cold (http): 3433ms -> 3270ms
* memory: 248MB -> 267MB

Some modest perf wins in exchange for some modest memory losses. Iterating past this point merits some profiling.

<details>

<summary> benchmark results </summary>

# New Hot (local)

```
benchmark results (ms):
  1472, 1450, 1451, 1474, 1466, 
average runtime: 00m:01s:462ms (1462ms)
median runtime: 00m:01s:466ms (1466ms)
min runtime: 00m:01s:450ms (1450ms)
max runtime: 00m:01s:474ms (1474ms)
```

# New Cold (http)

```
benchmark results (ms):
  3395, 3271, 3203, 3261, 3222, 
average runtime: 00m:03s:270ms (3270ms)
median runtime: 00m:03s:261ms (3261ms)
min runtime: 00m:03s:203ms (3203ms)
max runtime: 00m:03s:395ms (3395ms)
```



</details>

minidump-processor/src/processor.rs
```diff
@@ -356,93 +356,110 @@ where
...
+                    arg_recovery::fill_arguments(&mut stack, stack_memory.as_deref());
+                }
 
-        threads.push(stack);
+                stack
+            },
+        ))
+        .await
+        .into_iter()
+        .collect::<Vec<_>>();
```
Swatinem, line 461: `join_all` already returns a `Vec`.

minidump-processor/src/processor.rs
Swatinem: That is indeed one of the main disadvantages of async task spawning :-(

minidump-processor/src/processor.rs
```diff
@@ -356,93 +356,110 @@ where
...
+        let memory_list = &memory_list;
+        let thread_names = &thread_names;
+        let evil = &evil;
+        let modules = &modules;
+        let system_info = &system_info;
+        let unloaded_modules = &unloaded_modules;
+        let options = &options;
+        let requesting_thread = &requesting_thread;
+        threads = futures_util::future::join_all(thread_list.threads.iter().enumerate().map(
+            |(i, thread)| async move {
```
Swatinem, line 375: to combat the excessive rightward drift, you can also create a binding for the closure:
`let process_thread = |(i, thread)| async move { … };`
I think you would need some type annotations in that case.

Swatinem (commented):
this looks very promising! I haven’t reviewed the parser/http stuff in detail though.

Gankra:
we landed the most important part of this in #623

PR #602, opened 2022-07-01
Emit inlines in the minidump JSON

mstange:
This adds an `inlines` field to the JSON for every stack frame which has at least one inlined call at the stack frame address. Each element in the `inlines` array has the properties `function`, `file`, and `line`, with the latter two being optional. The inlines are ordered from inside to outside.

This format matches what I originally proposed it in [this bugzilla comment](https://bugzilla.mozilla.org/show_bug.cgi?id=1636194#c6) for the [Tecken / symbolication API](https://tecken.readthedocs.io/en/latest/symbolication.html). That's a different API, but its response JSON format was influenced by the minidump JSON format.

The implementation of the symbolication API in [profiler-get-symbols](https://github.com/mstange/profiler-get-symbols) is already using that format - it is used when profiling local Firefox builds or when using [profiler-symbol-server](https://github.com/mstange/profiler-symbol-server/) to resymbolicate profiles.

[Here is an example minidump JSON](https://gist.github.com/mstange/55648469006e25fd24285af9cea0e03a) where I manually replaced the XUL.sym file in the symbol cache with a XUL.sym containing inlines, which I generated from the corresponding XUL.dSYM with a local dump_syms which includes the work from https://github.com/mozilla/dump_syms/pull/392. (Note that example minidump JSON contains file paths of the form `/builds/worker/workspace/...`; this is only because that's what's in the dSYM. Ignore that difference.)

This PR is WIP because it's missing tests and documentation.

### DIFF
--- a/breakpad-symbols/src/lib.rs
+++ b/breakpad-symbols/src/lib.rs
@@ -452,6 +452,9 @@ pub trait FrameSymbolizer {
     fn set_function(&mut self, name: &str, base: u64, parameter_size: u32);
     /// Set the source file and (1-based) line number this frame represents.
     fn set_source_file(&mut self, file: &str, line: u32, base: u64);
+    /// Add an inline frame. This method can be called multiple times, in the
+    /// order "outside to inside".
+    fn add_inline_frame(&mut self, _name: &str, _file: Option<&str>, _line: Option<u32>) {}
 }
 
 pub trait FrameWalker {
--- a/breakpad-symbols/src/sym_file/mod.rs
+++ b/breakpad-symbols/src/sym_file/mod.rs
@@ -6,6 +6,7 @@ pub use crate::sym_file::types::*;
 pub use parser::SymbolParser;
 use std::fs::File;
 use std::io::Read;
+use std::ops::Deref;
 use std::path::Path;
 use tracing::trace;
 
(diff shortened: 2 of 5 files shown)


Gankra:
High level thoughts:

* "inside" vs "outside" means absolutely nothing to me. can we have better names? or somewhere that describes what that means? My guess is "outside" is like what would be the direct callee if it wasn't inlined, and inside the deepest grand-grand-grand...-callee?
* I don't understand why the code and json want to be reversed on the order?
* Please update https://github.com/rust-minidump/rust-minidump/blob/main/minidump-processor/json-schema.md

mstange:
> High level thoughts:
> 
>     * "inside" vs "outside" means absolutely nothing to me. can we have better names? or somewhere that describes what that means? My guess is "outside" is like what would be the direct callee if it wasn't inlined, and inside the deepest grand-grand-grand...-callee?

You got it, inside = callee-most, outermost = the function you'd see if you were ignoring all inline information.
I'm happy to change the terminology.

>     * I don't understand why the code and json want to be reversed on the order?

I think the JSON sould have the callee-to-caller order because that matches the order of "real" frames in the stack.

The order of calls to `add_inline_frame` is the reverse of that because that was easier to implement - otherwise you'd have to buffer things up and then reverse a Vec. Which would definitely be fine. And that's what I would suggest if symbols were returned in a return value, rather than via method calls. But since the current API is using method calls, I thought I might as well take advantage of it.

mstange:
Another option would be to "flatten" inline frames into "real" frames in the JSON. So there would be inline frames and "outer" / "real" frames sitting next to each other in the same list.

This would have the advantage that the JSON format doesn't change and that consumers of this format don't need to do any work to get inline information.

For inline frames, we could give the following values:

 - trust: "inline"
 - address: the same address as the corresponding real frame
 - module: the same module as the corresponding real frame

So, if the real stack is 0x10/XUL/context, 0x20/XUL/cfi, 0x30/XUL/cfi, and each of these addresses expands to a function with two extra inline functions, you would get:

0x10/Grandcallee10/XUL/inline, 0x10/Callee10/XUL/inline, 0x10/Caller10/XUL/context, 0x20/Grandcallee20/XUL/inline, 0x20/Callee20/XUL/inline, 0x20/Caller20/XUL/cfi, 0x30/Grandcallee30/XUL/inline, 0x30/Callee30/XUL/inline, 0x30/Caller30/XUL/cfi

This format is not a perfect fit for reality because the duplicated address / module is a bit silly, but on the other hand I can't think of any harm this would cause.

However, the flattening approach would not work for the human readable output which has per-frame register values. For the output with register values, we should only print the register values of each "real" frame once, and group the inlines per "real" frame.

Gankra:
obsoleted by #635

PR #609, opened 2022-07-07
bump the apple ptr-auth real bits to 47

Gankra:
The proper value for a system can be found with

```
% sysctl machdep.virtual_address_size
```

Which on the machine mstange tested yields 47. We were seeing 40 being too low in practice
for JIT code pages. Ideally this is information that the minidump_writer should stuff somewhere
so that we can precisely do this. Apparently crashpad already does this somewhere but I lost
track of that message.


Swatinem (approved):
I have looked at other pieces of code that do PAC stripping, and 47 looks like a good value here.

mstange:
> Apparently crashpad already does this somewhere but I lost track of that message.

https://github.com/rust-minidump/rust-minidump/issues/607#issuecomment-1176834572

Issue #204, opened 2021-06-11
Try to detect hardware-level memory corruption?


Gankra:
Tom Ritter mentioned this to me, I figured I'd write it down here in case someone feels inspired. 

Basically sometimes we get crash reports that we can't make sense of and decide are probably hardware failures in the RAM or loading it from disk or something. It would be nice if minidump-processor could actually do some analysis and suggest that possibility, similar to the exploitability heuristic in breakpad.

Some old bugzilla entries suggesting methodologies (potentially impossible for rust-minidump or require support from the minidump client, but hey, inspiration!):

* [Check if the binary around the crash matches what we ship](https://bugzilla.mozilla.org/show_bug.cgi?id=1274628)
* [Mark crashes that happen at invalid instructions](https://bugzilla.mozilla.org/show_bug.cgi?id=1274428)
* [Annotate whether the system has ECC ram](https://bugzilla.mozilla.org/show_bug.cgi?id=1489297)
* [Example where this analysis was done by hand](https://bugzilla.mozilla.org/show_bug.cgi?id=1475772)
* [Annotate JIT pages](https://bugzilla.mozilla.org/show_bug.cgi?id=1251395) (maybe already in MemoryInfo?)
* [Run memtest from the crashreporter](https://bugzilla.mozilla.org/show_bug.cgi?id=995652)

gabrielesvelto:
I ran a GSoC about this a couple of years ago (search for "helpful crash reporter" in [this page](https://wiki.mozilla.org/Community:SummerOfCode19)). There's useful stuff that can be done on the client side and that could weed out a lot of useless reports but it's always been "nice to have". That GSoC in particular went nowhere because the student disappeared midway through the work.

Note that most of this stuff would either have to happen in the code that writes the minidump (e.g. memtest'ing the crashed process memory) or afterwards in the minidump-analyzer. The latter is currently based on Breakpad but would be switched over to using this crate after we're done with Socorro.

luser:
> [Check if the binary around the crash matches what we ship](https://bugzilla.mozilla.org/show_bug.cgi?id=1274628)

This should be doable given that the symbol server also hosts the matching binaries. As it turns out, heycam wrote this in Rust a while back so you could in fact just reuse that code (or at least use it as a starting point): https://github.com/heycam/chkimg

gabrielesvelto:
I filed a couple more bugs which somewhat overlap the existing ones:
[...]

PR #611, opened 2022-07-07
WIP add support for using symbolserver binaries for bitflip checks

Gankra:
this is an experiment I don't plan to merge any time soon


gabrielesvelto:
DLLs in memory contains various changes that happen at load time, so I don't think you can compare them directly. See bug [1738774](https://bugzilla.mozilla.org/show_bug.cgi?id=1738774) for more details.

luser:
I mentioned this over in #204 but heycam wrote a Rust implementation of this a while ago that handles relocations etc: https://github.com/heycam/chkimg

Gankra:
I'm probably not going to find the time to dig into this deeper any time sooner, I underestimated the complexity

PR #614, opened 2022-07-08
ref: Move SymbolSupplier out of the processor

Swatinem:
The SymbolSupplier was so far re-exported from breakpad-symbols along
with some constructor functions.

This is on top of #613 and just moves some things out of `minidump-processor`.


Gankra:
gonna close as bitrotted, alas this dragon is eternal

PR #623, opened 2022-07-19
ref: Process threads concurrently

Swatinem:
This makes the inner loop that processes and stackwalks all minidump threads
concurrent using join_all concurrency. This could possibly reduce the time to process
a minidump if there is IO involved in fetching symbols for different modules concurrently.

I extracted this code from #601 with minimal modifications, so credit goes to @Gankra ;-)

The PR also adds docs for the `SymbolProvider` to clarify what its purpose and requirements are.


loewenheim (commented):
Just some style suggestions.

PR #628, opened 2022-07-22
Make minidump-processing optionally interactive

Gankra:
Add make minidump-stackwalk use some of this interactivity to provide feedback during non-json analysis (on the assumption that there's a human in the loop).

This is WIP because I need to implement proper progress bars.

### DIFF
--- a/minidump-processor/src/processor.rs
+++ b/minidump-processor/src/processor.rs
@@ -5,16 +5,17 @@ use std::borrow::Cow;
 use std::collections::{BTreeMap, BTreeSet};
 use std::ops::Deref;
 use std::path::Path;
+use std::sync::{Arc, Mutex};
 use std::time::{Duration, SystemTime};
 
 use minidump::{self, *};
 
-use crate::arg_recovery;
 use crate::evil;
 use crate::process_state::{CallStack, CallStackInfo, LinuxStandardBase, ProcessState};
 use crate::stackwalker;
 use crate::symbols::*;
 use crate::system_info::SystemInfo;
+use crate::{arg_recovery, FrameTrust, StackFrame};
 
 /// Configuration of the processor's exact behaviour.
 ///
@@ -63,6 +64,22 @@ pub struct ProcessorOptions<'a> {
     /// Currently this only work for x86, and assumes everything is either cdecl or thiscall
     /// (inferred from whether the symbol name looks like a static function or a method).
     pub recover_function_args: bool,
+
+    /// A reporter for the number of threads that have been processed
+    pub thread_stat_reporter: Option<Arc<Mutex<u64>>>,
+    /// A reporter for the number of stackframes that have been processed
+    pub frame_stat_reporter: Option<Arc<Mutex<u64>>>,
+    /// A reporter for streaming frames as they are walked
+    pub frame_reporter: Option<Arc<Mutex<Vec<WalkedFrame>>>>,
+    /// A reporter that yields the ProcessState before any stack analysis
+    pub partial_state_reporter: Option<Arc<Mutex<Option<ProcessState>>>>,
+}
+
+#[derive(Debug, Clone)]
+pub struct WalkedFrame {
+    pub thread_idx: usize,
+    pub frame_idx: usize,
+    pub frame: StackFrame,
 }
 
 impl ProcessorOptions<'_> {
@@ -83,6 +100,10 @@ impl ProcessorOptions<'_> {
         ProcessorOptions {
             evil_json: None,
             recover_function_args: false,
+            thread_stat_reporter: None,
+            frame_stat_reporter: None,
+            frame_reporter: None,
+            partial_state_reporter: None,
         }
     }
 
@@ -101,6 +122,10 @@ impl ProcessorOptions<'_> {
         ProcessorOptions {
             evil_json: None,
             recover_function_args: false,
+            thread_stat_reporter: None,
+            frame_stat_reporter: None,
+            frame_reporter: None,
+            partial_state_reporter: None,
         }
     }
 
@@ -116,6 +141,10 @@ impl ProcessorOptions<'_> {
         ProcessorOptions {
             evil_json: None,
             recover_function_args: true,
+            thread_stat_reporter: None,
+            frame_stat_reporter: None,
+            frame_reporter: None,
+            partial_state_reporter: None,
         }
     }
 
@@ -145,7 +174,7 @@ impl Default for ProcessorOptions<'_> {
 }
 
 /// An error encountered during minidump processing.
-#[derive(Debug, thiserror::Error)]
+#[derive(Clone, Debug, thiserror::Error)]
 pub enum ProcessError {
     #[error("Failed to read minidump")]
     MinidumpReadError(#[from] minidump::Error),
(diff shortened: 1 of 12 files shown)


minidump-stackwalk/src/main.rs
```diff
@@ -397,8 +402,53 @@ async fn main() {
...
+            };
+
+            // Do one final sync stat update
+            if needed_stats.load(std::sync::atomic::Ordering::Relaxed) {
+                let symbol_stats = provider.pending_stats();
+                let (t_done, t_pending) = *thread_stat_reporter.lock().unwrap();
+                let frames_walked = *frame_stat_reporter.lock().unwrap();
+
+                update_status(&symbol_stats, t_done, t_pending, frames_walked);
+            }
```
Gankra, line 449: hmm this pattern is something I need to remember for later

Gankra:
sample of the new interactive ui based on indicatif 
image


(This took way longer than it had any right to, because it turns out that the last stable release of indicatif is horribly broken, and they've been just releasing 0.17.0 release candidates for over a year now.)

Gankra:
I'm also successfully using this in minidump-debugger to make the ui much more responsive, live updating the stackframes as the stackwalk happens! https://github.com/Gankra/minidump-debugger/commit/82cdfd64f9645d3155e4fa166316d16f3b728828

Gankra:
While I think there are some rough edges (excessive Mutex/Arc usage? Weird UX?) I think this is worth discussing at this point.

I should definitely benchmark the impact.

minidump-processor/src/stackwalker/mod.rs
```diff
@@ -190,66 +190,76 @@ async fn fill_source_line_info<P>(
...
+        stack.thread_name.as_deref().unwrap_or(""),
+    );
+    // Begin with the context frame, and keep getting callers until there are no more.
+    let mut has_new_frame = !stack.frames.is_empty();
+    while has_new_frame {
+        // Symbolicate the new frame
+        let num_frames = stack.frames.len();
+        let frame = stack.frames.last_mut().unwrap();
+
+        fill_source_line_info(frame, modules, symbol_provider).await;
```
Gankra, line 217: An interesting feature of this new design having you pass in a CallStack: as per @luser, if you think that a frame has an error, you can truncate the stack to that frame, make an edit to it, and then pass the CallStack back into here. The stackwalker will the happily pick up from that point.

However you will lose some of the post-processing analysis done in minidump-processor. If we really cared about that we could either pull that processing into this API or factor out that step further. This is more a musing

Gankra:
Ok this has graduated from "kinda jank but interesting" to "pretty nice and clean".

Gankra:
minidump-debugger also now looks pretty nice with the new API https://github.com/Gankra/minidump-debugger/compare/interactive

Gankra:
I started to add some driveby fixups to minidump on the basis that this is basically the "minidump-debugger feature branch". One of them I added was MacCrashInfo printing support, but all the minidumps I'm checking in firefox have num_records=0. @gabrielesvelto did we break something?

Gankra:
Also if anyone wants to test this stuff out, I have some prebuilt binaries for macos/windows. (linux requires me to figure out some [package management stuff](https://github.com/emilk/egui/tree/master/egui_glow), but I don't have a desktop linux setup (only WSL) so I can't really properly test it.)

https://github.com/Gankra/minidump-debugger/releases/tag/v0.2.0-prerelease1

willkg:
@Gankra You can find crash reports that have a non-0-records mac_crash_info section by checking for `mac_crash_info exists`:

https://crash-stats.mozilla.org/search/?mac_crash_info=%21__null__&product=Firefox&date=%3E%3D2022-07-18T13%3A15%3A00.000Z&date=%3C2022-07-25T13%3A15%3A00.000Z&_facets=signature&_sort=-date&_columns=date&_columns=signature&_columns=product&_columns=version&_columns=build_id&_columns=platform&_columns=mac_crash_info#crash-reports

Here's some:

* https://crash-stats.mozilla.org/report/index/03049fa2-fc83-4377-9cf3-b46390220725
* https://crash-stats.mozilla.org/report/index/cb04174c-fb6b-45db-878d-331510220725
* https://crash-stats.mozilla.org/report/index/7d63a66a-7633-4a9b-8f55-4c9f20220725

Gankra:
Ah perfect, thanks!

image

Gankra:
ah there we go 
image

Gankra:
image

even better

Gankra:
ok i did some hasty benchmarking and yeah if this changes perf at all (at least when disabled) it's in the noise.

minidump-processor/src/processor.rs
```diff
@@ -63,6 +64,212 @@ pub struct ProcessorOptions<'a> {
...
+            let mut stats = self.stats.lock().unwrap();
+            stats.total_threads = total_threads;
+        }
+    }
+
+    /// Record that a thread has been processed.
+    pub(crate) fn inc_processed_threads(&self) {
+        // Only bother doing this if the user cares
+        if self.subscriptions.thread_count {
+            let mut stats = self.stats.lock().unwrap();
```
flub, line 242: since you worry about all these locks, did you consider making this entire reporting infrastructure a channel?  Something like to subscribe you'd have to pass in the appropriate sender to ProcessOptions and then all of the modify functions here are sending things over events, maybe including the current check on whether the subscriber cares about the event.

minidump-processor/src/processor.rs
```diff
@@ -63,6 +64,212 @@ pub struct ProcessorOptions<'a> {
...
+#[derive(Default, Debug, Clone)]
+struct PendingProcessorStatsInner {
+    /// How many threads have been processed
+    num_threads_processed: u64,
+    /// How many threads there are in total (redundant, but convenient)
+    total_threads: u64,
+    /// The number of frames that have been walked
+    num_frames_processed: u64,
+    /// Frames that have been walked
+    new_walked_frames: Vec<WalkedFrame>,
```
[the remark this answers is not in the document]
Gankra: The API to query this stat does `drain`, so it's since the last time you queried it.

minidump-processor/src/processor.rs
Gankra: The "pull" lock approach seems a lot easier to work with for the actual intended users (interactive UIs with specific refresh rates) than a "push" approach which would require the ui to keep waking up and updating state when it doesn't necessarily care about updating things.

But I am not a Concurrency Person so someone with stronger convictions could convince me otherwise.

minidump-processor/src/processor.rs
```diff
@@ -479,5 +647,94 @@ where
...
+        let memory_list = &memory_list;
+        let modules = &state.modules;
+        let system_info = &state.system_info;
+        let unloaded_modules = &state.unloaded_modules;
+        let options = &options;
+
+        futures_util::future::join_all(
+            state
+                .threads
+                .iter_mut()
```
Swatinem, line 667: wow, I’m even surprised the borrowchecker lets you do this, but okay ;-)

minidump-processor/src/symbols.rs
```diff
@@ -159,6 +159,7 @@ pub trait SymbolProvider {
     /// Keys are implementation dependent.
     /// For example the file name of the module (code_file's file name).
     fn stats(&self) -> HashMap<String, SymbolStats>;
+    fn pending_stats(&self) -> PendingSymbolStats;
```
Swatinem, line 162: Can you document this please so implementors know what they have to do?

I tried making these things have a default impl, which might even work for non-async functions. Though the compiler complained to me about Object Safety when I tried to add a default impl for `get_file_path`.

Swatinem (approved):
lgtm.

I would very much prefer some docs (and a default impl if possible) for the added trait methods

minidump-processor/src/processor.rs
Gankra: yes this is the Super Cool Thing about the rust iterator API, each yielded reference has its lifetime tied to the original collection, and not the iterator, so

```
let a = iter_mut.next();
let b = iter_mut.next();
*a = 1;
*b = 2;
```

Is totally valid!

This is only sound because collection iterators are "one-shot" and even if you mix `next` vs `next_back`, you'll only ever get one reference to each element, and need to discard them all before getting a new iterator. It's basically a generalization of `slice::split_at_mut`. :)

minidump-processor/src/symbols.rs
Gankra: done

PR #640, opened 2022-07-29
Fix json schema re: registers

### LINKED ISSUE #639: registers not printed in json output in places i thought it should be
It's a terrible summary, but the gist of it is that I spent today writing a jsonschema for the processed crash in Socorro because reasons when I noticed that the schema [here](https://github.com/rust-minidump/rust-minidump/blob/main/minidump-processor/json-schema.md) suggests there should be registers object at

```
crashing_thread.registers
```

but the rust-minidump stackwalker output puts it at:

```
crashing_thread.frames.[0].registers
```

Also, I expected there to be corresponding registers object at:

```
threads.[crashing_thread].frames.[0].registers
```

because the schema said that the `crashing_thread` object was primarily redundant information.

Anyhow, seems like the spec and the reality are out of sync. I think Socorro ignores register information at the moment, so this is only an issue with my jsonschema thing. Also, this is probably related to #364.

willkg:
This rectifies the spec in regards to where register information is emitted.

Fixes #639.

### DIFF
--- a/minidump-processor/json-schema.md
+++ b/minidump-processor/json-schema.md
@@ -219,6 +219,21 @@ anyway.
             | "frame_pointer"  // Used frame pointers to unwind (often reliable)
             | "scan",          // Searched the callee's stack memory (SKETCHY!)
 
+          // The values the general purpose registers contained.
+          //
+          // In the default configuration, this field will only show up in
+          // frame 0 of the crashing_thread copy of this backtrace. In the
+          // future, this may be more configurable and appear in different
+          // frames.
+          //
+          // The contents of this <object> are platform-specific,
+          // but it's always a mapping from register names to <hexstring>s.
+          //
+          // e.g. "rip": "0x000000010bbc852e"
+          "registers": {
+            "some_register_name": <hexstring>,
+          }
+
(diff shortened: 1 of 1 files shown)


willkg:
@Gankra Does this look like what you were looking for?

willkg:
Huh. This is my first contribution to this repo. I thought I did other things, but apparently not.

minidump-processor/json-schema.md
```diff
@@ -219,6 +219,16 @@ anyway.
             | "frame_pointer"  // Used frame pointers to unwind (often reliable)
             | "scan",          // Searched the callee's stack memory (SKETCHY!)
 
+          // The values the general purpose registers contained (frame 0 only!).
```
Gankra, line 222: I think instead of "frame 0 only" I would say something like "In the default config, this field will only show up in frame 0 of the crashing_thread copy of this backtrace. In the future this may be more configurable and appear in different frames."

Gankra (approved):

willkg:
^^^ That updates the comment.

I don't have write access to this repo, so if you think it's good, feel free to merge it.

Gankra:
thanks!

PR #655, opened 2022-08-12
Add instruction analysis for amd64

marti4d:


minidump-processor/src/op_analysis.rs
```diff
@@ -0,0 +1,733 @@
...
+/// used to determine a target memory address, and allow for single operations involving
+/// multiple memory locations (bit blitting)
+///
+/// If the given `instruction_bytes` contain a valid instruction, and the given `context` contains
+/// valid values for all the registers used for address calculation, this function will return a
+/// (possibly-empty) list of all the memory accesses for the instruction
+///
+/// ⚠ NOTE ⚠ - Certain instructions like PUSH are not considered to access memory, even though
+/// they do technically write to the stack. Generally this shouldn't be of much concern, but
+/// there may be some situations where this detail matters
```
iximeow, line 173: hi! i'm really curious about this comment in particular -- there was [a bug](https://github.com/iximeow/yaxpeax-x86/blob/no-gods-no-/CHANGELOG#L83-L88) i'd fixed in yaxpeax-x86 where _implicit_ memory accesses (`push rax`) were not reported, but _explicit_ memory accesses (`push qword [rax]`) were. i added [a few test cases](https://github.com/iximeow/yaxpeax-x86/blob/no-gods-no-/test/long_mode/operand.rs#L32-L47) when fixing that issue, so now yaxpeax-x86 should be consistently reporting an instruction accesses memory even if it's not through an explicit operand.

since this comment seems similar to a bug i'd fixed in `1.1`, and you're using `1.1`, and i don't see explicit mention of instructions like `push` in this PR, i'm curious if you'd initially used an older version, found the bug, then added the comment referencing the now-fixed yaxpeax behavior?

(ps: it's very cool to see yax is useful here! use in tools is exactly what i'd had in mind for the crates :tada:)

iximeow: ah nvm! i see the logic now: there won't be a memory-accessing operand to iterate over, so you won't record a memory access. and if you _did_ want to record a memory access you'd need to consult the `Opcode` to know what registers to use in an address calculation (like.. `rsp`).

so then there's an addendum to offer to this comment if you're interested: cases like `push qword [rax]` _would_ be considered as accessing memory, but only to read the operand, not for the implied push afterwards :grin:

luser:
Is this a stepping-stone towards dealing with crashes involving non-canonical pointers? (Because it sure feels like it!)

minidump-processor/src/op_analysis.rs
Gankra: (just for historical record, ixi and i discussed this and it seems like the set of instructions with implicit memory operands are pretty rare and enumerable. I believe it's just the various flavours of push, pop, call, save, restore. I'm perfectly fine with landing this as is though, and waiting for yaxpeax to provide some general API for enumerating implicit operands.)

minidump-processor/src/op_analysis.rs
```diff
@@ -0,0 +1,733 @@
...
+///
+/// # Errors
+///
+/// May fail if the given bytes are too short to form an instruction, don't represent a valid
+/// instruction encoding, registers needed to calculate the address are invalid, or if the
+/// given CPU architecture is not supported or enabled by the current feature set
+///
+pub fn get_instruction_memory_access(
```
Gankra, line 181: This trailing blank doc-comment line is unconventional-enough style that I'd appreciate having them removed just for uniformity.

minidump-processor/src/op_analysis.rs
```diff
@@ -0,0 +1,733 @@
...
+    let instruction_pointer = context.get_instruction_pointer();
+
+    memory
+        .seek(SeekFrom::Start(instruction_pointer))
+        .map_err(OpAnalysisError::ReadThreadInstructionFailed)?;
+
+    // We use MAX_INSTRUCTION_LENGTH here as an optimization to avoid allocating and copying
+    // more bytes than needed to produce an instruction
+
+    let mut buffer = Vec::with_capacity(MAX_INSTRUCTION_LENGTH.into());
```
Gankra, line 139: Is it at all anticipated that MAX_INSTRUCTION_LENGTH can become absurdly large?

(I am mostly just curious if there's something fucked up on some architecture that you know about.)

minidump-processor/src/op_analysis.rs
Gankra: Also I guess if I'm bothering you about comment formatting, ending paragraphs/sentences with periods would be ideal.

minidump-processor/src/op_analysis.rs
```diff
@@ -0,0 +1,733 @@
...
+                    let index = get_reg(reg)?;
+                    let scale = scale.unwrap_or(1);
+                    index.wrapping_mul(scale.into())
+                }
+                None => 0,
+            };
+
+            let disp = disp.unwrap_or(0) as i64 as u64;
+
+            Ok(base.wrapping_add(scaled_index).wrapping_add(disp))
```
Gankra, line 285: 💯 on making all the arithmetic in here robust to evil inputs!

minidump-processor/src/op_analysis.rs
```diff
@@ -0,0 +1,733 @@
...
+
+        for idx in 0..decoded_instruction.operand_count() {
+            let operand = decoded_instruction.operand(idx);
+
+            let maybe_address = match operand {
+                Operand::DisplacementU32(disp) => Some(disp.into()),
+                Operand::DisplacementU64(disp) => Some(disp),
+                Operand::RegDeref(base) => Some(calculate_address(Some(base), None, None, None)?),
+                Operand::RegDisp(base, disp) => {
+                    Some(calculate_address(Some(base), None, None, Some(disp))?)
```
Gankra, line 310: hmm there's a potential argument here that instead of `?` this should be `.ok()`, so that you can recover *some* memory accesses even if others fail. I think that could happen if you're looking at a frame that isn't the context (top) frame, as this would result in only some of the registers being valid.

That said the caller would need to then be wary of incorrect "negative reasoning" about the absence of memory accesses (because it could just be that some of them were silently dropped for being uncomputable).

minidump-processor/src/op_analysis.rs
```diff
@@ -0,0 +1,733 @@
...
+                Operand::RegIndexBaseScaleDisp(base, index, scale, disp) => Some(
+                    calculate_address(Some(base), Some(index), Some(scale), Some(disp))?,
+                ),
+                _ => None,
+            };
+
+            if let Some(address) = maybe_address {
+                memory_accesses.push(MemoryAccess {
+                    address,
+                    size: mem_size,
```
Gankra, line 345: weird, all the different memory accesses share the same size?

minidump-processor/src/op_analysis.rs
```diff
@@ -0,0 +1,733 @@
...
+            data.expected_size = 8;
+            data.expected_addresses = &[0xfffffffffff08000];
+            access_test(&data);
+        }
+
+        #[test]
+        fn test_base_index_scale() {
+            // mov al, [r12 + r14 * 2]
+            let mut data = AccessTestData {
+                bytes: &[0x43, 0x8a, 0x04, 0x74],
```
Gankra, line 645: I think what you're doing here is cleaner but just in case you're unaware, this is also a valid way to declare a byte array/slice: `b"\x43\x8a\x04\x74"`

minidump-processor/src/process_state.rs
```diff
@@ -275,16 +293,8 @@ pub struct ProcessState {
...
-    /// The memory address implicated in the crash.
-    ///
-    /// If the process crashed, and if the crash reason implicates memory,
-    /// this is the memory address that caused the crash. For data access
-    /// errors this will be the data address that caused the fault. For code
-    /// errors, this will be the address of the instruction that caused the
-    /// fault.
-    pub crash_address: Option<u64>,
+    /// Info about the exception that triggered the dump (if one did)
+    pub exception_info: Option<ExceptionInfo>,
```
Gankra, line 297: heads up to sentry (@Swatinem)

minidump-processor/src/processor.rs
```diff
@@ -566,6 +553,50 @@ where
...
+
+        let (instruction_str, memory_accesses) = exception
+            .context(&dump_system_info, misc_info.as_ref())
+            .map(|context| {
+                let mut memory_list_cursor = MemoryListCursor::new(&memory_list);
+
+                let instruction_str = crate::op_analysis::pretty_print_thread_instruction(
+                    &context,
+                    &mut memory_list_cursor,
+                )
```
Gankra, line 572: Hmm so to be clear this code wouldn't do well at reporting something like "hey this JIT code tried to jump to the middle of an instruction", yeah? I know that's not entirely reliable to detect. I'm also not sure if the chunk of memory that minidumps grab includes any bytes before the crash address or if it's just after.

(This is great as is, just thinking out loud.)

minidump-processor/src/process_state.rs
```diff
@@ -264,6 +265,23 @@ pub struct LinuxStandardBase {
     pub description: String,
 }
 
+#[derive(Debug, Clone)]
+pub struct ExceptionInfo {
```
Gankra, line 269: This type itself is missing docs

Gankra (approved):
This basically LGTM, I mostly have a lot of idle thoughts and nits. You clearly put a ton of work into this, seem to handle overflow/untrusted-input gotchas, and have solid testing and docs!

minidump-processor/src/op_analysis.rs
```diff
@@ -0,0 +1,733 @@
...
+/// byte may be needed to represent an instruction
+///
+/// Specifying too low of a value for this may cause long instructions to fail to decode. Too
+/// high of a value won't affect functionality, but may result in unnecessary memory usage
+pub const MAX_INSTRUCTION_LENGTH: u8 = MAX_AMD64_INSTRUCTION_LENGTH;
+
+/// The maximum size of an instruction on Amd64 architecture
+///
+/// Amd64 instructions can be between 1 and 15 bytes in length
+pub const MAX_AMD64_INSTRUCTION_LENGTH: u8 = 15;
```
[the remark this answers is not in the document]
marti4d: There is a reason for this -- Currently `std::io::Read::take()` uses a `u64` as its argument, and since `From<usize>` is not implemented for `u64` it means I either have to `as` cast or use `TryFrom`, neither of which I love considering I know statically that the value is <256

minidump-processor/src/op_analysis.rs
marti4d: Yeah -- I looked at [the list of amd64 opcodes in iximeow's library](https://docs.rs/yaxpeax-x86/latest/yaxpeax_x86/long_mode/enum.Opcode.html), had a minor heart attack, and decided that perhaps some things were better left to experts 😂

marti4d:
> Is this a stepping-stone towards dealing with crashes involving non-canonical pointers? (Because it sure feels like it!)

It is indeed! I think that change will be coming quite shortly!

minidump-processor/src/op_analysis.rs
marti4d: Fair enough, and done!

minidump-processor/src/op_analysis.rs
marti4d: I'm not really aware of any architecture where an instruction would be very large. I guess it's theoretically possible these days with CPUs having so much onboard memory, but I doubt we'll be dealing with anything like that anytime soon!

minidump-processor/src/op_analysis.rs
marti4d: Now that you mention it, I have seen that before but it was buried away somewhere deep. I think I still like this better, TBH

minidump-processor/src/processor.rs
marti4d: I think it would work just fine! It will just attempt to start decoding bytes at wherever `rip` is pointing. If that happens to be a  valid instruction, it will pretty-print it even if it's garbage

marti4d: (But if yaxpeax fails to decode it because it's an invalid instruction, we won't pretty-print it)

minidump-processor/src/process_state.rs
marti4d: Thanks! Good catch!

minidump-processor/src/processor.rs
luser: > I'm also not sure if the chunk of memory that minidumps grab includes any bytes before the crash address or if it's just after.

It is typically 256 bytes centered on the instruction pointer. I implemented that in Breakpad for mac and Linux, attempting to match the default behavior from Windows, and remember writing fairly thorough tests for it.

gabrielesvelto:
> Is this a stepping-stone towards dealing with crashes involving non-canonical pointers? (Because it sure feels like it!)

Yes! And much more, see a bunch of ideas we had [here](https://wiki.mozilla.org/Crash_reporting_improvements#Disassembly_in_the_stack_walker)

minidump-processor/src/op_analysis.rs
gabrielesvelto: Operand size is encoded per-instruction on x86 and that should apply to all instructions (and in particular vector instructions can only have one memory operand IIRC)

minidump-processor/src/processor.rs
```diff
@@ -738,3 +768,76 @@ where
 
     Ok(state)
 }
+
+/// A cursor that allows a memory list to be used like a byte stream
```
luser, line 772: My commentary here is purely advisory in my status as maintainer emeritus, but are you sure this abstraction pulls its weight here? In my experience a minidump either has a single memory region that contains the memory containing the instruction pointer or it doesn't. I can't recall ever seeing an instance where there was an adjacent memory region that would have been useful to continue reading in this manner.

Does this code set you up for future success in some way with other planned work?

For comparison's sake, here's the Rust port of my `get-minidump-instructions` tool that I used many times for doing this analysis in an ad-hoc manner:
https://github.com/rust-minidump/rust-minidump/blob/4f6eae86dd16f0aaf0b6a8056989d76b691a9934/minidump-tools/src/lib.rs#L216-L247

It simply tries to find a memory region containing the instruction pointer, and uses that if available. I cannot recall ever running into a situation where having this abstraction would have produced more useful results.

Gankra: Some of this code is potentially reusable by third-party integrations like minidump-debugger. I could imagine this sort of logic being useful for debugging Full Memory Minidumps, but I'm not sure what the stream looks like in practice for that situation (i.e. can a contiguous region be fragmented across entries?).

luser: > I could imagine this sort of logic being useful for debugging Full Memory Minidumps, but I'm not sure what the stream looks like in practice for that situation

We have code for reading the memory streams from full memory dumps, each memory region is a contiguous chunk of bytes:
https://github.com/rust-minidump/rust-minidump/blob/a6a486b8992008c6865ee70fc68421c38e45ef42/minidump/src/minidump.rs#L1896-L1915

luser:
> I did this mostly because I needed some trait anyway to make it easy to unit test without synthesizing my own fake memory list. I originally had a GetMemoryAtAddress single-function trait and changed it to this. I guess I will just go back to that

I'm always sympathetic to anything that makes it easier to write tests! I think the [`SynthMinidump`](https://docs.rs/minidump-synth/latest/minidump_synth/) bits we have are full-featured enough to make this straightforward. You can see an example of creating a synthetic x86-64 dump in this test:
https://github.com/rust-minidump/rust-minidump/blob/38edd05c27588f573a447d89d07bdba252f601d9/minidump/src/minidump.rs#L6561-L6574

The processor tests have a function to build a minimal minidump that they add additional info to:
https://github.com/rust-minidump/rust-minidump/blob/a6a486b8992008c6865ee70fc68421c38e45ef42/minidump-processor/tests/test_processor.rs#L136-L149

Calling `SynthMinidump::add_memory` automatically adds the memory region you provide to the minidump and also to the memory list stream.

There are tests with examples of adding exception info:
https://github.com/rust-minidump/rust-minidump/blob/38edd05c27588f573a447d89d07bdba252f601d9/minidump/src/minidump.rs#L6697-L6709

All of this is built atop the [`test-assembler` crate](https://docs.rs/test-assembler/latest/test_assembler/), which gives you a bunch of ways to string together arbitrary data for testing purposes. The stackwalker tests are a good point of reference, they set up the stack contents to test specific unwind scenarios, you could do virtually the same thing to create memory regions with specific instructions for your tests:
https://github.com/rust-minidump/rust-minidump/blob/a6a486b8992008c6865ee70fc68421c38e45ef42/minidump-processor/src/stackwalker/amd64_unittest.rs#L116-L130

luser:
I think with a little bit of up-front work you can write a function that generates suitable synthetic minidumps for your tests so that the individual tests can be very straightforward. I assume most of your tests will just want to test specific instructions, so you could write something like `fn create_synth_dump(memory: Memory) -> SynthMinidump`. Each test would use the `Section` APIs to populate the desired instructions then create a `Memory` with that as its contents and pass it to this function. `create_synth_dump` would add that memory to the dump, along withat least one thread with stack memory + context, an exception stream + context whose instruction pointer points into `memory`, and a module and a few other necessary stock bits.

If you get stuck, one of us can help you out.

marti4d:
@luser @Gankra 

Alrighty -- In response to your feedback, I have removed the `std::io::{Seek, Read}` implementations on the memory lists, and now the public `op_analysis` convenience functions just directly take a `&MinidumpMemoryListBase` as their `memory` argument. 

The unit tests now just directly use the `get_instruction_memory_access()` function, which simply takes a `instruction_bytes: &[u8]`, so now I'm just skipping the memory list machinery in my unit tests altogether. I think it's actually better that way anyhow -- The memory list stuff is already well-covered by other tests, so this really breaks the unit tests down to "just test that your new assembly decoding logic is correct".

minidump-processor/src/processor.rs
marti4d: > > I did this mostly because I needed some trait anyway to make it easy to unit test without synthesizing my own fake memory list. I originally had a GetMemoryAtAddress single-function trait and changed it to this. I guess I will just go back to that
> 
> I'm always sympathetic to anything that makes it easier to write tests! I think the [`SynthMinidump`](https://docs.rs/minidump-synth/latest/minidump_synth/) bits we have are full-featured enough to make this straightforward. You can see an example of creating a synthetic x86-64 dump in this test:
> 
> https://github.com/rust-minidump/rust-minidump/blob/38edd05c27588f573a447d89d07bdba252f601d9/minidump/src/minidump.rs#L6561-L6574
> 
> The processor tests have a function to build a minimal minidump that they add additional info to:
> 
> https://github.com/rust-minidump/rust-minidump/blob/a6a486b8992008c6865ee70fc68421c38e45ef42/minidump-processor/tests/test_processor.rs#L136-L149
> 
> Calling `SynthMinidump::add_memory` automatically adds the memory region you provide to the minidump and also to the memory list stream.
> 
> There are tests with examples of adding exception info:
> 
> https://github.com/rust-minidump/rust-minidump/blob/38edd05c27588f573a447d89d07bdba252f601d9/minidump/src/minidump.rs#L6697-L6709
> 
> All of this is built atop the [`test-assembler` crate](https://docs.rs/test-assembler/latest/test_assembler/), which gives you a bunch of ways to string together arbitrary data for testing purposes. The stackwalker tests are a good point of reference, they set up the stack contents to test specific unwind scenarios, you could do virtually the same thing to create memory regions with specific instructions for your tests:
> 
> https://github.com/rust-minidump/rust-minidump/blob/a6a486b8992008c6865ee70fc68421c38e45ef42/minidump-processor/src/stackwalker/amd64_unittest.rs#L116-L130

Thank you for all the info here. I will probably use this for the unit testing for the stuff I build on top of this feature (like non-canonical address handling and other things in the list shared by @gabrielesvelto)

marti4d:
@Gankra As long as there are no further concerns, I believe this should be good-to-go :)

Gankra:
thanks for all the hard work, sorry for the delays!

PR #665, opened 2022-08-26
Report non-canonical addresses on Amd64 properly

marti4d:
@gabrielesvelto -- Here is the change I promised for non-canonical address detection. I tested, and it seems to be working fine on Window, Linux, and macOS.

I will be away next week, so if you want to merge this sooner-rather-than-later and you only have minor nits, I would not be offended if you just pull my changes into your own fork and fix them; otherwise, I can do it when I get back on Sept 6


Gankra:
See also https://github.com/EmbarkStudios/crash-handling/pull/53 which adds back the non-canonical case to minidump-pipeline/crash-handling tests, which I didn't have time to finish up.

minidump-processor/src/processor.rs
```diff
@@ -763,3 +769,107 @@ where
...
+    // Only Amd64 has non-canonical addresses
+    if system_info.cpu != Cpu::X86_64 {
+        return;
+    }
+
+    // Different OSes have unique ways of reporting this error
+    match system_info.os {
+        // Windows reports it as EXCEPTION_ACCESS_VIOLATION_READ, address 0xffffffffffffffff
+        Os::Windows => {
+            if exception_info.reason
```
luser, line 805: These series of `if` statements might be easier to read if you combined them into `match` statements.

marti4d: Fair enough! I attached a new commit - It does seem to read better

minidump-processor/src/processor.rs
```diff
@@ -763,3 +769,103 @@ where
...
+            Os::Linux,
+            CrashReason::LinuxGeneral(minidump_errors::ExceptionCodeLinux::SIGBUS, SI_KERNEL_U32),
+            0,
+        ) => true,
+        (Os::Linux, _, _) => false,
+        (_, _, _) => {
+            tracing::warn!("we don't currently support non-canonical analysis for your OS");
+            return;
+        }
+    };
```
gabrielesvelto, line 844: It would be nice to wrap this large check into its own function so you could do something like:
```rust
if !is_non_canonical_access(...) {
    return;
}
```
below

gabrielesvelto (approved):
LGTM with just a minor nit

gabrielesvelto (changes requested):
I forgot something important! Please update the [schema](https://github.com/rust-minidump/rust-minidump/blob/main/minidump-processor/json-schema.md) to include the new fields you're introducing with this PR.

minidump-processor/src/processor.rs
marti4d: Works for me! Done :)

marti4d:
> I forgot something important! Please update the [schema](https://github.com/rust-minidump/rust-minidump/blob/main/minidump-processor/json-schema.md) to include the new fields you're introducing with this PR.

Alrighty! Should be good-to-go, hopefully. Please let me know if you see anything else I missed.

BTW -- Is there a way I am supposed to test to ensure I didn't make a mistake with my JSON schema? I don't know if `cargo test` ended up testing it or not...?

gabrielesvelto:
Duh, I forgot to merge this :facepalm:

luser:
Very cool to see this finally happen after *years* of frustration despite my having written a C++ implementation with Breakpad!

PR #682, opened 2022-09-21
add release workflow

Gankra:
With these changes, if you type:

```
cargo release 0.15.0 --execute
```

It will:

* update all the versions in Cargo.tomls
* update a bunch of version strings in RELEASES and snapshot tests
* tag v0.15.0
* push to main (with the tag)
* cargo publish

Then CI will see the tag and:

* create a github release with the notes from RELEASES.md
* create tarballs/zips for windows/macos/linux * x64/arm64

This makes the main chore of a release "making sure the "Next Release" entry has the releases notes you want (manual, intentionally). Potentially this should have a post-release hook to add the commit SHA as I currently do manually... problem for later.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,50 @@
+name: Release
+
+permissions:
+  contents: write
+
+on:
+  push:
+    tags:
+      - v[0-9]+.*
+
+jobs:
+  create-release:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: taiki-e/create-gh-release-action@v1
+        with:
+          # (optional) Path to changelog.
+          changelog: RELEASES.md
+          # (required) GitHub token for creating GitHub Releases.
+          token: ${{ secrets.GITHUB_TOKEN }}
+
+  upload-assets:
+    strategy:
+      matrix:
+        include:
+          - target: aarch64-unknown-linux-gnu
+            os: ubuntu-latest
+          - target: aarch64-apple-darwin
+            os: macos-latest
+          - target: x86_64-unknown-linux-gnu
+            os: ubuntu-latest
+          - target: x86_64-apple-darwin
+            os: macos-latest
+          - target: x86_64-apple-darwin
+            os: macos-latest
+          - target: x86_64-pc-windows-msvc
+            os: windows-latest
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v3
+      - uses: taiki-e/upload-rust-binary-action@v1
+        with:
+          # (required) Comma-separated list of binary names (non-extension portion of filename) to build and upload.
+          # Note that glob pattern is not supported yet.
+          bin: minidump-stackwalk
+          # (optional) Target triple, default is host triple.
+          target: ${{ matrix.target }}
+          # (required) GitHub token for uploading assets to GitHub Releases.
+          token: ${{ secrets.GITHUB_TOKEN }}
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,3 +10,9 @@ members = [
 exclude = [
     "testdata/*",
 ]
+
+[workspace.metadata.release]
+shared-version = true
+consolidate-commits = true
+consolidate-pushes = true
+tag-name = "v{{version}}"
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -1,6 +1,7 @@
-# Version XXX (not yet released)
+<!-- next-header -->
+# Next Version
 
-Nothing to see here yet!
+TODO
 
 # Version 0.14.0 (2022-08-03)
 
--- a/minidump-stackwalk/Cargo.toml
+++ b/minidump-stackwalk/Cargo.toml
@@ -36,3 +36,12 @@ tracing-subscriber = "0.3.14"
 insta = "1.13.0"
 minidump-synth = { path = "../minidump-synth" }
 test-assembler = "0.1.6"
+
+[package.metadata.release]
+pre-release-replacements = [
+  {file="../RELEASES.md", search="Next Version", replace="Version {{version}} ({{date}})", min=1},
+  {file="../RELEASES.md", search="<!-- next-header -->", replace="<!-- next-header -->\n# Next Version\n", exactly=1},
+  {file="tests/snapshots/test_minidump_stackwalk__markdown-help.snap",  search="minidump-stackwalk 0.*`", replace="minidump-stackwalk {{version}}`", exactly=1},
+  {file="tests/snapshots/test_minidump_stackwalk__long-help.snap",  search="minidump-stackwalk 0.*", replace="minidump-stackwalk {{version}}", exactly=1},
+  {file="tests/snapshots/test_minidump_stackwalk__short-help.snap", search="minidump-stackwalk 0.*", replace="minidump-stackwalk {{version}}", exactly=1}
+]
--- a/minidump-stackwalk/tests/test-minidump-stackwalk.rs
+++ b/minidump-stackwalk/tests/test-minidump-stackwalk.rs
@@ -466,7 +466,13 @@ fn test_version() {
     let mut ver_parts = ver.trim().split('.');
     ver_parts.next().unwrap().parse::<u8>().unwrap();
     ver_parts.next().unwrap().parse::<u8>().unwrap();
-    ver_parts.next().unwrap().parse::<u8>().unwrap();
+    let last = ver_parts.next().unwrap();
+    if let Some((last, _prerelease)) = last.split_once('-') {
+        last.parse::<u8>().unwrap();
+    } else {
+        last.parse::<u8>().unwrap();
+    }
+
     assert!(ver_parts.next().is_none());
 }
 


Gankra:
Here's an unpublished test 0.15.0 release: https://github.com/Gankra/rust-minidump/releases/tag/v0.15.0

Gankra:
Currently the tarballs just contain the binary but [we could add files like READMES and LICENSES if folks find that useful](https://github.com/taiki-e/upload-rust-binary-action#example-workflow-include-additional-files).

Gankra:
# Experience Report for Combining cargo-release and taiki-e's CI Actions

rust-minidump is a workspace with about 6 crates and some idiosynchrocies:

* **Structure**: Most crates are libraries but one of them (minidump-stackwalk) is a CLI application
* **Coupled**: These crates are fairly heavily coupled and most changes cross-cut through them
* **Lockstep Versions**: To keep documentation/release-notes clear, I have mandated that all crates in the workspace will have the same version, which results in releases for packages which have no changes, or minor changes being escalated to major releases just because something else had a major change. This is suboptimal but I think worth it for the sake of docs (for now).
* **Version Bumps Break Tests**: There are [a few snapshot tests that contain the version number for a package](https://github.com/rust-minidump/rust-minidump/blob/main/minidump-stackwalk/tests/snapshots/test_minidump_stackwalk__long-help.snap), because we snapshot 3 flavours of `--help`  and clap puts that string there. The snapshots must therefore be updated after a version bump for CI tests to pass.
* **README Snapshot Hax**: The README for minidump-stackwalk [contains a copy of one of those snapshots](https://github.com/rust-minidump/rust-minidump/tree/main/minidump-stackwalk#minidump-stackwalk-cli-manual) to document the CLI. This snapshot is made with [the --help-markdown hack](https://github.com/rust-minidump/rust-minidump/blob/main/minidump-stackwalk/src/main.rs#L489)
* **Self-Referential RELEASES.md**: in addition to the expected need to update "upcoming release" to "version x.y.z (date)", I also try to link the commit of the release, which is a circular dependency. In my human workflow I would change "commit: TBD" to "commit: this one!" and then do a post-release commit changing it to the right SHA. However if we use tagging properly we can potentially just pre-emptively link [the github URL of the tag](https://github.com/rust-minidump/rust-minidump/releases/tag/0.14.0), and maybe that's better.

## cargo-release

* ✅ Out of the box, the fact that it defaults to --dry-run semantics is great for my confidence in fucking around
  * 😭 However I am made of hubris and by the end of this I had done about 40 commits and pushed about 15 tags to github (just finished deleting them...)
* ✅ Fantastically, my "lockstep bump all crates in the workspace" workflow is just the default behaviour of cargo-release in workspaces! 
* ✅ But I didn't want a bunch of different commits or tags, so I wanted basically all the "consolidate/shared" flags flipped on, which did exactly what I wanted!
  * 😭 I wasn't paying close enough attention so my first non-dry run didn't have these flags, adding more commits and tags for me to cleanup later
* ✅ I wanted to do various semi-wet-runs, and `--push-remote`, `--no-publish` and `--no-push` are absolute MVPs for letting me do only some of the things or test things on my own fork
* 😕 As a human I would do `cargo test` and then `cargo insta accept` to update the snapshots, but this seemed like a bunch of work since `cargo-release` only lets you specify one command to run as a pre-release hook(?), so I opted for pre-release-replacements that just assume what snapshots need to be updated where (with bad regexes I should probably fix...). This *is* a snappier experience so I'm calling this just "meh". 
* ❌ pre-release-replacements in the workspace get run on every package in the workspace, instead of once on the workspace itself. I couldn't find anyway to change this. This makes it difficult to edit a global RELEASES.md or README.md. You can either use `../` but that's sketchy and requires the workspace to be flat... or you can do what I did and put all the global edits in one of the packages.
* ❌ Because I didn't want to risk a "real" release I worked almost exclusively in prerelease versions, but pre-release-replacements (ironically, given the punned terms here) doesn't actually run on prereleases. This is where I wasted most of my time, and produced my most garbage partial-wet-runs because in the default setup there's no feedback for this. I tried so much stuff until I just looked at the code and saw a guard and log statement, and then remembered that epage had told me that `-v` is super powerful for getting more precise feedback. Once I did that I could see the logs properly saying "hey this is ignored in pre-release" and I was able to pivot temporarily to "full release but totally dry with -v" to debug/test my find-replace before doing a proper release.
* ❌ At this point the above means I'm actually not sure how I would ever try to publish a pre-release in a way that the CI packaging tasks would work (because the snapshot strings don't get updated, and that will get killed by test failures).
* 😕 had some weirdness with trying to write regexes (what syntax?) in toml strings where the parser was freaking out. I was really tired at this point and just simplified the header (removed parens) that I was searching for to get rid of all that. No idea
* 😕 you get a ton of warnings for having publishes for packages with no changes. Honestly this is fair but made my terrible workflow a bit more annoying to debug since I got a ton of warnings drowning out the things I cared about.
* 🚧I never implemented the  **Self-Referential RELEASES.md** workflow (ran out of steam, but I can see two ways to do it fairly straighforward)
* 🚧I never implemented the  **README Snapshot Hax** workflow (this one is a bit harder to picture, since I need to delete a ton of stuff across a bunch of different lines and multi-line regex stuff is usually gross... also I don't want to have a dep on shell scripting. portability!)

In spite of how this looks my overall experience was positive, but definitely I had some struggles.


## taiki-e ci workflows

I poured over a few of the different repos they have for CI, [and tokio-console's usage](https://github.com/tokio-rs/console/blob/main/.github/workflows/release.yaml), but ended up mostly just copy-pasting [the excellent cross-compiling example and doing some light edits](https://github.com/taiki-e/upload-rust-binary-action#example-workflow-cross-compilation) in [upload-rust-binary-action](https://github.com/taiki-e/upload-rust-binary-action). I just changed CHANGES.md to RELEASES.md and added windows. 🎊

My main complaint is that because all of this stuff is built out purely as github actions (which invoke shell scripts) there really wasn't a good way for me to do test/dry runs without just pushing tags to my repo until I had it refined.

I think all the issues I had were just the consequences of me not having my cargo-release workflow working right and the action complaining that RELEASES.md didn't have an entry. Fair!

epage:
> pre-release-replacements in the workspace get run on every package in the workspace, instead of once on the workspace itself. I couldn't find anyway to change this. This makes it difficult to edit a global RELEASES.md or README.md. You can either use ../ but that's sketchy and requires the workspace to be flat... or you can do what I did and put all the global edits in one of the packages.

The workspace configuration is meant for layering with the package configuration.  There isn't a "workspace release" we could even tie to.  Tying to a specific package is the intended behavior.

> Because I didn't want to risk a "real" release I worked almost exclusively in prerelease versions, but pre-release-replacements (ironically, given the punned terms here) doesn't actually run on prereleases. 
>
>  At this point the above means I'm actually not sure how I would ever try to publish a pre-release in a way that the CI packaging tasks would work (because the snapshot strings don't get updated, and that will get killed by test failures).

There is a [setting to enable this](https://github.com/crate-ci/cargo-release/blob/master/docs/reference.md#prepost-release-replacements)

(I had the opposite and with clap v4.0.0-rc.1, I was worried pre-release-replacements ran and the default saved me)

>  had some weirdness with trying to write regexes (what syntax?) in toml strings where the parser was freaking out. I was really tired at this point and just simplified the header (removed parens) that I was searching for to get rid of all that. No idea

`regex` crate.  I can imagine the escaping can be a pain.  Using toml string literals (`''`) might help.

>  you get a ton of warnings for having publishes for packages with no changes. Honestly this is fair but made my terrible workflow a bit more annoying to debug since I got a ton of warnings drowning out the things I cared about.

The intention is to help people who might want to release the minimal subset since I don't yet trust all of the logic to have people defer to cargo-release to make all decisions.

Open to alternatives on this.

> Self-Referential RELEASES.md: 

I always just link to the tag.  Works well for me.

> I never implemented the README Snapshot Hax 

Don't know what features insta offers but I update [cargo-edit's README](https://github.com/killercup/cargo-edit) using [`trycmd`](https://github.com/killercup/cargo-edit/blob/master/tests/readme.rs) (built on `snapbox`) which allows me to add wildcards that are inocuous enough to be in user-facing documentation.  Similarly, this is what I use for validating all of the example output in clap's reference and tutorials.

Gankra:
> There is a [setting to enable this](https://github.com/crate-ci/cargo-release/blob/master/docs/reference.md#prepost-release-replacements).

Ah dope, genuinely happy to know I was just failing to RTFM here and that there isn't some complicated feature work to be done. I think it's an understandable behaviour. I think maybe there's an argument for an always-on top-level warning/note about prereleases having some functionality disabled (and how to change that) but I know there's always a signal-noise tradeoff.

> regex crate

We should definitely link that and/or specify the variant we're using in the docs. (It's my pet-peeve that people just say "regex" without qualifying it because regexes are fake and everyone has weird extensions...)

> The intention is to help people who might want to release the minimal subset since I don't yet trust all of the logic to have people defer to cargo-release to make all decisions.

No yeah I think this is a good thing overall I was just trying to provide maximally accurate info on my experience.

> Don't know what features insta offers but I update [cargo-edit's README](https://github.com/killercup/cargo-edit) using [trycmd](https://github.com/killercup/cargo-edit/blob/master/tests/readme.rs) (built on snapbox) which allows me to add wildcards that are inocuous enough to be in user-facing documentation. Similarly, this is what I use for validating all of the example output in clap's reference and tutorials.

Aha, I'll take a look at that.

epage:
> We should definitely link that and/or specify the variant we're using in the docs. (It's my pet-peeve that people just say "regex" without qualifying it because regexes are fake and everyone has weird extensions...)

Done

gabrielesvelto:
I finally have some time to come around and actually try this - knowing I'll probably blow something up. But first let's merge this.

PR #713, opened 2022-10-25
Detect "null pointer in disguise"

marti4d:
This adds functionality to detect a "null pointer in disguise", which occurs when a null pointer is accessed with an offset (for example, in the context of an array or a field within a struct).

When such a situation is encountered, it will report the crash address as "0" instead of whatever non-zero offset is given by the report.


minidump-processor/src/processor.rs
```diff
@@ -873,3 +872,18 @@ fn is_non_canonical_exception(os: system_info::Os, exception_info: &crate::Excep
...
+/// Report `address == 0` if an exception happened accessing a "null pointer in disguise"
+///
+/// This function will search though all the memory accessses for the instruction and see if any
+/// of them were flagged by `op_analysis` as being a disguised nullptr. If so, we fix up the
+/// crash address to indicate `0` as the crash address instead of whatever offset
+fn fix_null_pointer_in_disguise(exception_info: &mut ExceptionInfo) {
+    if let Some(memory_accesses) = &mut exception_info.memory_accesses {
+        for access in memory_accesses.iter_mut() {
+            if access.is_likely_null_pointer_dereference {
+                access.address = 0;
```
gabrielesvelto, line 885: This is only changing the address in the corresponding memory access, not in the exception info. I'm wondering if maybe we should add another field to the `ExceptionInfo` structure to denote this type of adjustment instead of changing the raw data itself (this could be used retroactively for the non-canonical addresses too). So we could have for example:
```
enum AdjustedAddress {
  NonCanonical<u64>, // The original access was to a non-cannonical address, actual address provided here
  NullPointerWithOffset<u64>, // The base pointer was a null pointer, offset provided here
}

struct ExceptionInfo {
  pub address: u64, // The raw address as reported by the exception
  // ...
  pub adjusted_address: AdjustedAddress, // The crash address we think is the likely culprit
}
```
This would allow Socorro - or any other downstream user - to present this information in addition to the existing one instead of outright replacing it. WDYT?

jrmuizel: Yeah, I think it's valuable to be able to distinguish between a true null dereference vs a detected null dereference so that it's not confusing when looking at a crash in detail.

marti4d: Ah, good catch about the "not updating the original exception" thing.

But you're right -- Now that you say it, it makes much more sense to do this `enum` thing and include it as extra info along with the address. I have that done and uploaded now -- Thanks for the suggestion! Hopefully that works a lot better now 😁

gabrielesvelto (approved):
LGTM, thanks!

Issue #740, opened 2022-11-24
Where does SimpleSymbolSupplier look?


ePirat:
I was trying to use `minidump-stackwalk` to process some windows crash minidumps of VLC generated by Breakpad, however it never finds any symbols and I am not sure why. I was wondering, where exactly does the SimpleSymbolSupplier look for symbols? I was not able to find where it is implemented in this repository and unfortunately the trace log did not help either, it just logs:

```
TRACE unwind_thread{idx=32 tid=4744 tname=""}:unwind_frame{idx=2 fname=""}: frame pointer seems valid -- caller_ip: 0x68b41780, caller_sp: 0x0cf4fccc
TRACE unwind_thread{idx=32 tid=4744 tname=""}: locating symbols for module C:\Program Files (x86)\VideoLAN\VLC\plugins\video_output\libwinhibit_plugin.dll
TRACE unwind_thread{idx=32 tid=4744 tname=""}:symbols{module="libwinhibit_plugin.dll"}:locate_file{file_kind=BreakpadSym module="libwinhibit_plugin.dll"}: SimpleSymbolSupplier search
TRACE unwind_thread{idx=32 tid=4744 tname=""}: unwinding 0x0000000068b4177f
TRACE unwind_thread{idx=32 tid=4744 tname=""}:unwind_frame{idx=3 fname=""}: trying cfi
TRACE unwind_thread{idx=32 tid=4744 tname=""}:unwind_frame{idx=3 fname=""}: couldn't find symbols for address, cannot use cfi
```

I've passed in the path to my symbols folder which is structured like this:

```
/path/to/symbols
├── libwinhibit_plugin.dll
│   ├── 0F6AAC37BE6D68EA2EBAD0764460B05C1
│   │   └── libwinhibit_plugin.dll.sym
│   └── 4E423F5D4E9C644501F3939FBC4A5C5E1
│       └── libwinhibit_plugin.dll.sym
…
```

ePirat:
It seems actually the Windows minidumps/symbol files are a bit special as I could not get them to work with unpatched vanilla Google minidump-stackwalk either. So instead I tried it with macOS minidumps which work as expected with Googles tool but do not symbolicate with this rust stackwalk tool.

ePirat:
After some more debugging (I finally found the place in the code where it looks for the symbol files), it turns out that it actually properly finds the symbol files for my macOS crash but fails to parse all of them:

```
[...]

PR #742, opened 2022-11-24
Fix parsing symfiles that contain empty lines

ePirat:
While investigating #740 I figured out the issue was that the parser errors out on empty
lines in symfiles. This is fixed in this PR.

Note that there might be a nicer way to solve this, but I do not have much experience with Nom,
so any hints how to improve this are welcome.


ePirat:
Updated to use `multispace0` instead of actually parsing the empty line and transforming it into an no-op result.

gabrielesvelto (approved):
LGTM, thanks!

PR #724, opened 2022-11-04
Update configuration for cargo release 0.22+ and add release notes

gabrielesvelto:

PR #745, opened 2022-11-25
Fix cargo-release replacements

ePirat:
Cargo release was broken by accident when updating clap in 06348c251b3165741a6a6e744f02baef5da6c64d which removed the version line (Version: `minidump-stackwalk 0.14.0`) from the output, but missed removing the version number replacement in pre-release-replacements.

Fixes the issue observed in #724


gabrielesvelto (approved):
Thanks!

PR #754, opened 2022-12-07
Update range-map to 0.2

glandium:


gabrielesvelto (approved):
Looking good, thanks!

Issue #756, opened 2022-12-07
Add LICENSE files to releases of individual crates


molenzwiebel:
Hi there!

In order to give proper attributions for any crates I use, I use the [cargo about](https://github.com/EmbarkStudios/cargo-about) tool. This automatically gathers the licenses for each (transitive) dependency used. This tool is currently failing for all crates in this repository, as the LICENSE file is situated in the root of the repository and is therefore not included in the bundles that are pushed to crates.io.

I've gone ahead and manually elaborated the licenses in my project for now, but it'd be great if you could consider adding the licenses to the releases of the individual packages too. 

Thanks for all the work on rust-minidump!

gabrielesvelto:
I think this might be useful for [cargo-deny](https://github.com/EmbarkStudios/cargo-deny) too.

gabrielesvelto:
Done

molenzwiebel:
Thanks!

PR #757, opened 2022-12-08
Put a copy of the MIT license in every crate

gabrielesvelto:
This fixes issue #756


gabrielesvelto:
@luser I think this is the right way to do it but not having much experience with license I'd like to hear your opinion, especially because the license in the root folder is yours and I'm not sure what to do with it.

luser:
This looks fine, and if it fixes the tooling issue then we should do it. Go ahead and change the root LICENSE file to match everything else. I think the best way to phrase the authorship in the license would be something like:
`Copyright (c) 2015-2022 rust-minidump contributors`

luser (approved):

gabrielesvelto:
Thanks Ted, I'll adjust the copyright line and the root license.

PR #772, opened 2023-01-06
Improve the handling of CPU microcode values.

### LINKED ISSUE #683: Improve handling CPU microcode values
When using `--evil-json` we often provide an [annotation](https://searchfox.org/mozilla-central/rev/b1e5f2c7c96be36974262551978d54f457db2cae/toolkit/crashreporter/CrashAnnotations.yaml#228-231) that holds the CPU microcode. This should be used to populate the `system_info.cpu_microcode_version` field if we can't extract it from the minidump. Additionally we should print out this value in hexadecimal because that's the format which is most commonly used for listing versions.

afranchuk:
This reads the CPUMicrocodeVersion field from `--evil-json` and uses that if no other microcode version information is available in the minidump. It also changes output to print the microcode version as a hex string.

Closes #683.

One thing that isn't clear: should the output microcode version be 0-padded or not?

### DIFF
--- a/minidump-processor/src/process_state.rs
+++ b/minidump-processor/src/process_state.rs
@@ -917,8 +917,8 @@ Unknown streams encountered:
                 "cpu_arch": sys.cpu.to_string(),
                 "cpu_info": sys.cpu_info,
                 "cpu_count": sys.cpu_count,
-                // optional
-                "cpu_microcode_version": sys.cpu_microcode_version,
+                // optional, print as hex string
+                "cpu_microcode_version": sys.cpu_microcode_version.map(|num| format!("{:#018x}", num)),
             },
             "crash_info": {
                 "type": self.exception_info.as_ref().map(|info| info.reason).map(|reason| reason.to_string()),
--- a/minidump-processor/src/process_state.rs
+++ b/minidump-processor/src/process_state.rs
@@ -918,7 +918,7 @@ Unknown streams encountered:
                 "cpu_info": sys.cpu_info,
                 "cpu_count": sys.cpu_count,
                 // optional, print as hex string
-                "cpu_microcode_version": sys.cpu_microcode_version.map(|num| format!("{:#018x}", num)),
+                "cpu_microcode_version": sys.cpu_microcode_version.map(|num| format!("{:#x}", num)),
             },
             "crash_info": {
                 "type": self.exception_info.as_ref().map(|info| info.reason).map(|reason| reason.to_string()),
--- a/minidump-processor/src/processor.rs
+++ b/minidump-processor/src/processor.rs
@@ -418,6 +418,22 @@ where
     process_minidump_with_options(dump, symbol_provider, ProcessorOptions::default()).await
 }
 
+/// Get the microcode version from linux cpu info and evil options.
+fn get_microcode_version(linux_cpu_info: &MinidumpLinuxCpuInfo, evil: &evil::Evil) -> Option<u64> {
+    linux_cpu_info
+        .iter()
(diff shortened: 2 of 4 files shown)


minidump-processor/src/process_state.rs
```diff
@@ -917,8 +917,8 @@ Unknown streams encountered:
                 "cpu_arch": sys.cpu.to_string(),
                 "cpu_info": sys.cpu_info,
                 "cpu_count": sys.cpu_count,
-                // optional
-                "cpu_microcode_version": sys.cpu_microcode_version,
+                // optional, print as hex string
+                "cpu_microcode_version": sys.cpu_microcode_version.map(|num| format!("{:#018x}", num)),
```
gabrielesvelto, line 921: Traditionally we haven't padded this value and other places where it's used don't pad it either (think `/proc/cpuinfo` or Intel's documentation) so I'd leave it unpadded for consistency.

minidump-processor/src/processor.rs
```diff
@@ -468,17 +474,18 @@ where
...
+        .find_map(|(key, val)| {
+            if key.as_bytes() == b"microcode" {
+                val.to_str().ok()
+            } else {
+                None
+            }
+        })
+        .or_else(|| evil.cpu_microcode_version.as_ref().map(|s| s.as_str()))
+        .and_then(|val| val.strip_prefix("0x"))
+        .and_then(|val| u64::from_str_radix(val, 16).ok());
```
gabrielesvelto, line 488: Could you split this out in a helper function for better readability?

gabrielesvelto (changes requested):
This is mostly looking good, just a couple of nits to address and it's good to go

minidump-processor/src/processor.rs
afranchuk: To clarify, do you want the conversion from hex split out, or the entire block which determines the microcode version?

gabrielesvelto: The entire block, something like `get_microcode_version(linux_cpu_info, evil)`.

afranchuk: Given the use of local state and unlikelihood of reuse, would it be more appropriate to add comments rather than adding a helper function if readability is an issue? Otherwise you could make the same statement (as far as adding helper functions) for each of the following values that are extracted from streams (`linux_standard_base`, `threads`, etc).

Otherwise, I think `let cpu_microcode_version = get_microcode_version(...)` is redundant in readability (the `let cpu_microcode_version = ...` already makes the functional behavior of the right side of the `=` fairly clear). However, I respect and defer to your final judgement, since you are the code owner :smile:

gabrielesvelto: The name was just a suggestion, you could use something more expressive like `microcode_from_cpuinfo_or_evil_json()` or something else that feels appropriate. This isn't because of reuse but to start shrinking this function. We were a bit in a rush when we wrote this so we've ended up with several-hundred-lines-long functions that could use some structure to make it clearer what's going on to someone who reads them the first time.

afranchuk: Sure, that sounds like a worthy cause to me!

afranchuk: Do you mind if I refactor the code to shrink the function further? Or would it be preferable to do your requested change here and save that for a separate PR?

gabrielesvelto: Make it in a separate PR, it's fine if it's just for refactoring, I'd be very happy about that.

gabrielesvelto (approved):
LGTM, thanks!

PR #778, opened 2023-01-10
Account for calls/jmps to non-canonical addresses.

### LINKED ISSUE #749: Fix up the crashing address in cases where we're jumping to a non-canonical address on x86-64
When we encounter a memory access that hits a non-canonical address we already replace the crashing address with the one we compute by disassembling the crashing instruction. This is because the general protection fault that gets delivered to the exception handler does not contain the address itself, so we have to recover it from the contents of the registers at the point of crash. Jumping to a non-canonical address also causes a GPF so we should also replace the crashing address if the access was to a non-canonical one.

afranchuk:
Closes #749, refactoring a few things along the way.

### DIFF
--- a/minidump-processor/src/op_analysis.rs
+++ b/minidump-processor/src/op_analysis.rs
@@ -129,7 +129,7 @@ fn get_thread_instruction_bytes<'a, Descriptor>(
 #[cfg(feature = "disasm_amd64")]
 mod amd64 {
     use super::*;
-    use yaxpeax_x86::amd64::Instruction;
+    use yaxpeax_x86::amd64::{Instruction, Opcode, Operand, RegSpec};
 
     /// Amd64-specific instruction analysis
     ///
@@ -143,7 +143,8 @@ mod amd64 {
 
         let instruction_str = decoded_instruction.to_string();
 
-        let memory_accesses = get_instruction_memory_access(context, decoded_instruction)
+        let memory_accesses = GetMemoryAccess::new(context)
+            .get_instruction_memory_access(decoded_instruction)
             .map_err(|e| tracing::warn!("failed to determine instruction memory access: {}", e))
             .ok();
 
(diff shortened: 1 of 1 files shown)


minidump-processor/src/op_analysis.rs
```diff
@@ -153,62 +154,174 @@ mod amd64 {
...
+                        return Ok(());
+                    }
+                    // We assume that relative offsets (for CALL, JMP) and absolute values (CALLF,
+                    // JMPF) will be valid, so we don't check immediate operands, only registers.
+                    if let Operand::Register(reg) = decoded_instruction.operand(0) {
+                        let address = self.get_reg(reg)?;
+                        accesses.push(MemoryAccess {
+                            address,
+                            is_likely_null_pointer_dereference: address == 0,
+                            size: Some(1),
```
afranchuk, line 300: I wasn't sure whether this was appropriate or whether `None` would be better, since we don't know the actual read size.

minidump-processor/src/op_analysis.rs
```diff
@@ -153,62 +154,174 @@ mod amd64 {
...
+                        });
+                    }
+                }
+                // TODO: if the full minidump information is plumbed down here, we can support
+                // RET/IRET as well (by reading the stack to get the return address)
```
afranchuk, line 305: It would be nice to use the stack information for this. The stack info could also be used to read any indirect calls (when they fall in stack memory), like `call [rax+8]`, which is probably common for vtables. If the minidump has heap memory that could be used too!

minidump-processor/src/op_analysis.rs
gabrielesvelto: It's fine! Since x86 ISA variable-sized the first access by a jump instruction can be as small as one byte.

minidump-processor/src/op_analysis.rs
gabrielesvelto: This is a worthy follow-up, I'll file an issue for it. I was considering doing it for `ret` but you're right that it might be more interesting to deal with vtables, especially those of dead objects which would automatically lead to a non-canonical access due to the poison pattern.

gabrielesvelto (approved):
LGTM, thanks!

PR #779, opened 2023-01-12
Test for bit flips on faulting addresses if they don't map to valid memory.

### LINKED ISSUE #429: Implement a simple crash address bitflip check

afranchuk:
This is a fairly naive test. Generally speaking, it could end up being fairly easy to get a "hit" on valid memory by trying increasingly far offsets (by flipping each bit). This could be improved to account for memory region read/write/execute permissions in combination with the exception type (e.g. if the exception was a memory write error, only consider memory regions which were writable as valid intended addresses).

Closes #429.


gabrielesvelto:
> This is a fairly naive test. Generally speaking, it could end up being fairly easy to get a "hit" on valid memory by trying increasingly far offsets (by flipping each bit). This could be improved to account for memory region read/write/execute permissions in combination with the exception type (e.g. if the exception was a memory write error, only consider memory regions which were writable as valid intended addresses).

Yes, that would help. There's also another case which we encounter in the wild and it's a NULL pointer with a bit flipped. We often see crashes that look like this, the original code is:
```
if (foo_pointer) {
  do_something_with(*foo_pointer);
}
```
And the crash address is exactly NULL plus one flipped bit (e.g. 0x0000000000400000). Note that this analysis would need to be done after the other address adjustment operations because we wouldn't want to catch NULL-plus-power-of-two-offset accesses which are also common.

Like you mention in the PR doing the bit-flip detection should also happen after having retrieved the "true" address of a non-canonical access. That's a very common occurrence.

afranchuk:
I intentionally ignored the NULL case (which was described in #429) because I was confused as to the usefulness of checking for NULL. Specifically, I was thinking that the purpose of this bit flip check was to determine whether an **invalid** memory access was the result of a bit flip of a **valid** one, whereas the NULL case is an **invalid** access resulting from a bit flip of an **invalid** (NULL) one. However, the example you provided is illuminating, and it would certainly be useful to suggest that incorrect logic may have resulted from a bit flip. So I'll consider a NULL address as a potential target of bit flips errors as #429 suggested.

afranchuk:
> Like you mention in the PR doing the bit-flip detection should also happen after having retrieved the "true" address of a non-canonical access. That's a very common occurrence.

Could we perhaps change the JSON output to have `adjusted_address` as an array rather than a single value? Which consumers rely on this JSON schema, and is there a process for making breaking changes?

willkg:
Socorro (the Mozilla crash ingestion pipeline) relies on the JSON output for the rust-minidump stackwalker. We have no process for making breaking changes that I'm aware of.

However, Socorro is using 0.15.0 from 2022-11-30 and that predates  `adjusted_address`, so Socorro doesn't know anything about that field, yet.

afranchuk:
Ah, I didn't realize it was so recent. That makes breaking it less worrisome :) But it's still good to verify whether this is an appropriate change, or whether we think this particular case (divining potentially multiple adjusted addresses) will only be useful for bit-flip checking. If that is the case, it may be better to implement the bit-flip checking as a separate field rather than an `AdjustedAddress`.

willkg:
Yes, we can add a crash ingestion processor rule that looks at the array and performs whatever heuristics we want to determine if the crash report is likely from a bit flip and then stores that determination result somewhere. I think that should work out fine.

gabrielesvelto (commented):
Some tweaks I've been thinking of we could do here:
* If the crash hit a non-canonical address then try the bit-flip detection only in bits 48 through 63. If the bit-flip turned a regular address in a non-canonical one then the bit-flip should be there, not elsewhere.
* If the crash address is a canonical one only check bits 0-47, same as above, the flipped bit must be there otherwise it would have produced a general protection fault.

afranchuk:
Great idea! Do we only support amd64 right now (i.e., I don't need to _check_ that it's amd64 prior to adding this change, for now)?

afranchuk:
@gabrielesvelto is there anything else you think should be added now?

gabrielesvelto (approved):
This is fine! Let's merge this and start experimenting with it on real crashes to see what turns out

Issue #780, opened 2023-01-12
Deal with off-by-one addresses falling right outside of a symbol


gabrielesvelto:
We found crashes in the wild (like [this one](https://bugzilla.mozilla.org/show_bug.cgi?id=1800919#c10) in WebRTC code) where the crash address in the top frame falls right after the end of a symbol.  The resulting stacks are garbage as the stack walker starts scanning and finds completely unrelated stuff.

Adjusting the first frame to check for `frame.instruction_address - 1` yields the right symbol, unwinding information and stack so I'm tempted to pile this hack^CR^CR^CR^CR heuristic on top of the existing stack-walking code.

gabrielesvelto:
Note it could be that this issue is because `dump_syms` is miscalculating the size of the symbol, or the debug information is wrong. Since it's happening in non-Mozilla builds (Ubuntu, Arch) I don't have the time to trace it all the way back to the debug information.

Swatinem:
I assume the `.sym` files assume exclusive symbol end addresses? Meaning the next higher symbol starts at the same addr (start is inclusive)? Is this well documented somewhere? Does `dump_syms` (and `symbolic-debuginfo` for that matter) behave correctly here?

Also, have you looked at a snippet of disassembly around that instruction addr? Is it an invalid / trap instruction following the normal fn body?

gabrielesvelto:
> I assume the `.sym` files assume exclusive symbol end addresses? Meaning the next higher symbol starts at the same addr (start is inclusive)? Is this well documented somewhere? Does `dump_syms` (and `symbolic-debuginfo` for that matter) behave correctly here?

Yes, it's exclusive and they're both consistent with this.

> Also, have you looked at a snippet of disassembly around that instruction addr? Is it an invalid / trap instruction following the normal fn body?

Yeah, it looks like this:
```
 36e8da4:       e8 b7 cf fa ff          call   3695d60 <_ZNKSt4lessISt17basic_string_viewIcSt11char_traitsIcEEEclERKS3_S6_@@xul108+0x2f90>
 36e8da9:       00 00                   add    %al,(%rax)
```
[...]

PR #781, opened 2023-01-12
If we can't find a symbol for the top frame let's try again by substracting one from the instruction address.

gabrielesvelto:
This fixup addresses an issue we've found in the wild in Firefox crashes. Fixes issue #780.


gabrielesvelto:
@Gankra @luser @Swatinem WDYT about this? It's a hack so I'm not 100% sure I want it, but it does solve an issue we're experiencing. I'm just wondering if this would have any downsides. The only problematic situation I could think of with this is we'd have an address that's covered by CFI information but doesn't have a symbol, but an actual symbol exists just before the area covered by the CFI information. In that case we'd pick the potentially wrong symbol and its CFI information instead.

Swatinem (commented):
fwiw, in our Sentry Stackwalker, we do not provide symbol (`function_name`) information back to the stack walker at all, so on our end this would not do anything.

gabrielesvelto:
The instruction pointer came from the minidump itself, it's the crash address. The symbol and unwind information available cover the same range, so without adjusting the instruction pointer the walker finds neither the symbol nor the unwinding information.

gabrielesvelto:
I've got an additional piece of information which might explain the odd crashing address: the crash is not originating from user-space. That's something I hadn't paid attention to when looking at the problematic crashes in [bug 1800919](https://bugzilla.mozilla.org/show_bug.cgi?id=1800919). The crash reason is `SIGSEGV / SI_KERNEL` which means that the signal was sent from the kernel.

That might explain why the address doesn't fall neatly within a function's boundary; I don't know exactly how the kernel populates the address in this case. What I'm sure of is that it's no accident: the crashes we have on file are consistent among a fairly wide range of kernels (5.15.x to 6.0.x) and they all yield the same clean stack trace once the fixup is applied.

Would everybody feel better if this hack was only applied to crashes with this particular cause and only to the top frame? I could make it even optional given it covers a very specific scenario.

Swatinem:
> Would everybody feel better if this hack was only applied to crashes with this particular cause and only to the top frame? I could make it even optional given it covers a very specific scenario.

I believe that sounds like a better idea, yes.

Though I’m still suspecious of the code that causes this. Maybe lets move discussions of that to #780 ?

mstange:
> The symbol and unwind information available cover the same range

Ok, so then we should be able to change the heuristic to "is the previous address covered by unwind information" instead of "covered by symbol information". With that change, I would be fine with this.

gabrielesvelto:
Closing this, check https://github.com/rust-minidump/rust-minidump/issues/780#issuecomment-1383735331_ for the explanation.

PR #787, opened 2023-01-18
Ignore `std::io::ErrorKind::BrokenPipe`.

### LINKED ISSUE #563: Interrupting output piped to `more` causes panic
If on Windows I'm piping the output to `more`, like `minidump-stackwalk fa5fa88c-df34-443d-bac7-4e3afd805bab.dmp symbols | more`, and then interrupt the output of more by pressing Q, minidump-stackwalk generates a panic:

> [ERROR] Panic - A panic occurred at D:\Users\Jan\.cargo\registry\src\github.com-1ecc6299db9ec823\minidump-stackwalk-0.11.0\src\main.rs:522: called `Result::unwrap()` on an `Err` value: Os { code: 232, kind: BrokenPipe, message: "Die Pipe wird gerade geschlossen." }

afranchuk:
Since we are only dealing with files, this error will only result from stdio pipes closing, which we won't consider an error.

Closes #563.

### DIFF
--- a/minidump-stackwalk/src/main.rs
+++ b/minidump-stackwalk/src/main.rs
@@ -264,14 +264,25 @@ struct Cli {
     symbols_path_legacy: Vec<PathBuf>,
 }
 
-#[cfg_attr(test, allow(dead_code))]
 #[tokio::main]
 async fn main() {
+    if let Err(e) = main_result().await {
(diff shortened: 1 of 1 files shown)


gabrielesvelto (approved):
LGTM

PR #789, opened 2023-01-18
Support reading indirect call addresses and ret addresses from memory.

### LINKED ISSUE #782: Fix up the crashing address in cases where we're indirectly jumping to a non-canonical address on x86-64
This is a follow-up to issue #749. If we're dealing with a `ret` instruction or an indirect call (e.g. `call [rax+8]`, something which would be common when accessing a vtable) then the address won't be in a register but rather in the memory that's being fetched. For cases like `ret` we'll almost always have the address as it'll be on the stack. For calls it won't be as common to have the target memory area handy, but we can tweak the minidump writer to capture the areas pointed by the registers in the first frame - something we already do under certain conditions - to increase the probability of having access to the right object.

In particular this would be useful for detecting indirect calls into dead objects which will always result in a general protection fault, due to the contents of the object being filled with the poison pattern.

afranchuk:
Closes #782.

### DIFF
--- a/minidump-processor/src/op_analysis.rs
+++ b/minidump-processor/src/op_analysis.rs
@@ -87,17 +87,21 @@ pub struct MemoryAccess {
 ///
 /// Note that even if this function doesn't return an error, individual pieces of information
 /// may still be missing from the returned `OpAnalysis` structure.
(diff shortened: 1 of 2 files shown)


gabrielesvelto (approved):
Very nice, thank you!

PR #805, opened 2023-02-11
Introduce UnifiedMemoryList and UnifiedMemory 

Gankra:
...to abstract over memory streams and migrate all APIs to using them.

This makes rust-minidump work properly with modern "builtin" windows minidumps which only provide Memory64List. To see such a minidump: go to crash monitor, select an exe, right-click, "Create dump file". The resulting file will fail to stackwalk/analyze anything due to lack of memory.

This is an *enormous* breaking change to the primary `minidump` APIs, although minidump-processor's API I think works just as well. A second pass on docs might be worthwhile. Also it might be worthwhile to add an API to rust-minidump that handles fetching these streams for you.

Note that because I unconditionally only use the 64-bit stream over the 32-bit one if both exist, this will regress results if you specifically:

* Have both MemoryList streams defined
* The 64-bit one is junk and all the real values are in the 32-bit one

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ fn main() -> Result<(), Error> {
         // Use `Default` to try to make progress when a stream is missing.
         // This is especially natural for MinidumpMemoryList because
         // everything needs to handle memory lookups failing anyway.
-        let mem = UnifiedMemoryList::Memory(dump.get_stream::<MinidumpMemoryList>().unwrap_or_default());
+        let mem = dump.get_memory().unwrap_or_default();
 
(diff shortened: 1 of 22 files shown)


Gankra:
I was planning to relax today but I made the mistake of trying to mess with microsoft debuggers for fun and went down a rabbit hole debugging my debugging of my debugging.

Gankra:
Added that convenience API

afranchuk (approved):
Looks good to me, nicely cleaning up some interfaces.

PR #812, opened 2023-02-23
Use the procfs crate in minidump for parsing linux memory maps.

afranchuk:
There are a handful of tests (and thus functionality) regarding malformed input that we'd have to give up to merge this. This gives up support for parsing files with messed up whitespace and memory map lines that only have partial information (e.g. only the mapped addresses).

I don't think these two cases are important to support, however if we do want to support them I can make some changes to `procfs`.

I also ran into an odd case in the `backwards_range` test. As far as I can tell, the code in that test only incidentally failed because the old memory map parser was extremely permissive. The minidump used in that test has an empty or invalid `MinidumpLinuxMaps` stream (I think?), and the `MinidumpStream::read` call gets the entire minidump content as the stream `bytes`, which seems very weird. That definitely should produce an error as far as I'm concerned, so I'm unsure what the original intent of that test was.

### DIFF
--- a/minidump/src/minidump.rs
+++ b/minidump/src/minidump.rs
@@ -4,6 +4,8 @@
 use debugid::{CodeId, DebugId};
 use memmap2::Mmap;
 use num_traits::FromPrimitive;
+use procfs_core::prelude::*;
+use procfs_core::process::{MMPermissions, MemoryMap, MemoryMaps};
 use scroll::ctx::{SizeWith, TryFromCtx};
 use scroll::{self, Pread, BE, LE};
 use std::borrow::Cow;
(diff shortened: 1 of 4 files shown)


afranchuk:
https://github.com/eminence/procfs/issues/255

mstange:
It might be better to just move the parsing code into a new crate, rather than #ifdef'ing out all the Linux-specific code. You could have a crate which just has an API which accepts a `&[u8]` or a `&str` or a `Read`. I'm aware of three other implementations of procmap parsing: [1](https://github.com/rbspy/proc-maps/blob/edf1b96ad62e04a6e4b029f061a82bcf516a6fc1/src/linux_maps.rs#L59) [2](https://github.com/jabedude/procmaps/blob/473e2316ee54c2c49ab7a2ea79bd7c920461e08c/src/lib.rs#L176) [3](https://github.com/koute/not-perf/blob/911723cf9099cc0f9e45b921f045117535cd1f06/proc-maps/src/lib.rs#L52)

afranchuk:
I agree, and hope the `procfs` maintainer does too. One reason to try to use `procfs` over other crates is that we want to potentially read other `procfs` info in the future.

minidump/src/minidump.rs
```diff
@@ -5654,25 +5460,22 @@ mod test {
...
-        // Whitespace intentionally wonky to test robustness
-        let input = b"
+        use procfs::process::{MMPermissions, MMapPath};
 
- a90206ca83eb2852-b90206ca83eb3852 r-xp  10bac9000 fd:05 1196511 /usr/lib64/libtdb1.so\x20\x20
-c70206ca83eb2852-de0206ca83eb2852  -w-s  10bac9000 fd:05 1196511 /usr/lib64/libtdb2.so  (deleted)
-
-
-";
+        // TODO: is it okay to give up wonky whitespace support?
```
gabrielesvelto, line 5465: We can live without it given that the kernel output is unlikely to change dramatically over time

afranchuk: I was wondering whether this support existed because of real examples seen in the wild. But I didn't really think that likely given it's probably been static for quite a while in the kernel. At which point I was fearful that we need to support _very_ old systems. Though obviously it's not a difficult thing to add to `procfs`.

afranchuk:
Waiting on a new (initial) version of `procfs-core` to be published before pulling this out of draft.

afranchuk:
@gabrielesvelto `procfs-core` was finally published in October, so this can be merged.

PR #813, opened 2023-02-23
Make `minidump-common::format` structs writable through `scroll`.

afranchuk:
As far as I can tell, there was no reason for these structs to not implement `Pwrite`. This is needed for https://github.com/rust-minidump/minidump-writer/issues/8.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -1,6 +1,8 @@
 <!-- next-header -->
 # Next Version
 
+* Make all `minidump-common::format` structs writable with `scroll`.
+
 # Version 0.15.2 (2022-12-07)
 
 * Updated the num-range crate, further reducing the number of dependencies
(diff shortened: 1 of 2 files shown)


luser:
Agreed. We simply didn't have any minidump-writing code when I first switched over to using scroll.

I wouldn't block on this, but it might be worth looking at the impact on code size/build time to decide whether it's worth using cargo features to conditionally enable the `Pread` / `Pwrite` derives? Presumably `minidump-writer` doesn't need `Pread`, and `minidump-stackwalk` doesn't need `Pwrite`.

luser (approved):

afranchuk:
On my system:

|  | release compile time | size |
|-|-|-|
| with Pread/Pwrite/SizeWith | 7.3s | 15MB |
| without | 6.66s | 12MB |
```

**Comment by gabrielesvelto** [comment]
If crate size is an issue I suppose it's rather easy to add something like:
```Rust
#[cfg_attr(feature = "write_support", derive(Pwrite)]
```
I don't know what crate sizes are considered acceptable these days. The increase in compilation time seems very marginal.

luser:
Given @afranchuk's testing, I don't think it's necessary. If it becomes a problem in the future we can address it.

gabrielesvelto:
Alright, merging this

PR #819, opened 2023-03-06
Only perform bit-flip detection on 64-bit platforms.

### LINKED ISSUE #810: Disable bit-flip detection in 32-bit minidumps
This just occurred to me as I was analyzing a minidump for a 32-bit x86 build. We've seen Firefox use more than 80% of the available 32-bit address space before crashing. With that much memory mapped it's very easy to get false positives with our algorithm for bit-flip detection. We also don't enable instruction disassembly on 32-bit minidumps AFAIK, so we can't use it to verify near-NULL pointer accesses or the like. It's better to just turn it off entirely.

afranchuk:
32-bit platforms have much higher likelihood of false-positives due to larger portions of the address space being mapped.

Closes #810.

### DIFF
--- a/minidump-processor/src/processor.rs
+++ b/minidump-processor/src/processor.rs
@@ -7,6 +7,7 @@ use std::path::Path;
 use std::sync::{Arc, Mutex};
 use std::time::{Duration, SystemTime};
 
+use minidump::system_info::PointerWidth;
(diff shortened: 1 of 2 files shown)


gabrielesvelto:
Thanks, merging.

PR #820, opened 2023-03-07
Add additional details about the circumstances around possible bitflip

### LINKED ISSUE #808: Extend bit-flip detection with heuristics to calculate confidence levels
There should be a confidence level provided for _each_ possible bit-flip address provided. The possible bit flips should be exposed in JSON as an array of objects, where each object has an `address` and `confidence` field.

Heuristics may include:
* Whether the bit-flipped address caused a non-canonical address (very high confidence)
* Whether the corrected address is near many other register values
* Whether the corrected address is null

It's unclear whether it would be more useful to decide the confidence levels we want to impart, or whether we want to provide flags of these heuristic conditions and allow users to calculate confidence levels with their own weighting.

afranchuk:
values.

This includes (somewhat arbitrary) confidence values to help in determining the confidence of a bit flip.

Closes #808.

Starting as a draft because documentation needs to be updated.

### DIFF
--- a/minidump-processor/src/processor.rs
+++ b/minidump-processor/src/processor.rs
@@ -571,6 +571,7 @@ where
 
     let mut exception_info: Option<crate::ExceptionInfo> = None;
     let mut exception_context: Option<std::borrow::Cow<MinidumpContext>> = None;
+    let mut instruction_registers: BTreeSet<&'static str> = Default::default();
 
     if let Some(exception) = exception_ref {
         let reason = exception.get_crash_reason(system_info.os, system_info.cpu);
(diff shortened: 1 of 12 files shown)


minidump-processor/src/processor.rs
```diff
@@ -568,66 +568,81 @@ where
...
+                        memory_accesses: op_analysis.memory_accesses,
+                        possible_bit_flips: Default::default(),
+                    });
+                    instruction_registers = op_analysis.registers;
+                }
+                Err(e) => {
+                    tracing::warn!("failed to analyze the thread context: {e}");
+                }
+            }
+        }
```
gabrielesvelto, line 621: Given this does very specific work it might be worth splitting this out into a separate function to make `process_minidump_with_options()` a bit shorter.

minidump-processor/src/processor.rs
```diff
@@ -568,66 +568,81 @@ where
...
                 instruction_str: None,
                 memory_accesses: None,
                 possible_bit_flips: Default::default(),
-            })
-    });
+            });
+        }
+    }
 
     // Check for bit-flips of the exception address
```
gabrielesvelto, line 635: Likewise, we could drop the comment and make a tersely-named function instead.

gabrielesvelto (approved):
LGTM, I just left a couple of formatting nits and the PR needs to be rebased.

minidump-processor/src/processor.rs
afranchuk: I will do so, but there's so much contextual stuff to pass around that this function may be better refactored into a struct with methods in the future (though I won't do that now).

afranchuk: I ended up refactoring it into a struct :shrug:

minidump/src/minidump.rs
```diff
@@ -4486,7 +4486,7 @@ impl<'a> MinidumpException<'a> {
         &self,
         system_info: &MinidumpSystemInfo,
         misc: Option<&MinidumpMiscInfo>,
-    ) -> Option<Cow<MinidumpContext>> {
+    ) -> Option<Cow<'a, MinidumpContext>> {
```
afranchuk, line 4489: It's not immediately clear why the API here returns a `Cow`, though I could see it potentially accessing the underlying bytes directly. If that is the intention, the lifetime was wrong (which was causing errors for me elsewhere).

gabrielesvelto:
Looking good! I'm merging the PR and I'll cut a new release tomorrow

Issue #824, opened 2023-03-27
Potentially rename some of the crates for clarity


afranchuk:
I think some of the crates in the project could have better names, especially since #816 will be adding another crate. Changing crate names causes confusion for current consumers, however keeping poor names can cause confusion for future consumers: either decision has a downside.

This is my proposal (as if #816 occurred, anything omitted stays the same):
| crate | new name | reasoning |
|--------|--------|--------|
| `minidump-processor` (non-stackwalk logic) | `minidump-processor` | encompasses all the extra things this does |
| `minidump-processor` (stackwalk logic) | `minidump-stackwalk` | it's the stackwalk function |
| `minidump-stackwalk` | `minidump-analyzer` | it does much more than just stack walking |

The mozilla `minidump-analyzer` crate (not in this project) could have another more specific name, like `minidump-stack-info`.

afranchuk:
Using the public information on crates.io, the external consumers look to be:
https://crates.io/crates/witchcraft-server
https://crates.io/crates/udbg
https://crates.io/crates/minidumper
https://crates.io/crates/minidump-debugger (assuming this doesn't fall into the "Mozilla" category)

PR #825, opened 2023-03-27
Split stackwalking out of the minidump-processor crate.

### LINKED ISSUE #816: Split the stack walking behavior out of `minidump-processor`
The new `minidump-analyzer` binary relies on `minidump-processor` for stack walking. However, this brings along a lot of extra dependencies which are unused (e.g. `yaxpeax`). Normally this might be a situation where a feature would be appropriate, however in combination with a need to audit dependencies with `cargo-vet`, it would make things much simpler to have stack walking as a separate crate.

Besides stack walking functionality, `minidump-analyzer` also needs the new debug info symbol provider from `minidump-processor`. That will need to go into a potentially new split-out crate, or there could be _another_ crate with symbol providers. It is annoying that `cargo-vet` can't support features so it forces us to break functionality into little pieces, though ultimately there isn't that much overhead aside from e.g. coordinating future additions that need to change multiple crates. The vendoring/auditing team felt much more comfortable with splitting the crates up (if possible) rather than adding a bunch of exemptions for the enormous crates which we aren't actually using. Not to mention vendoring those large crates would be cumbersome, too (or at the very least hacky with some dummy/stubbed out/patched versions).

@Gankra I'm wondering if you have any thoughts on this?

afranchuk:
The new minidump-walk-stack crate handles it and has far fewer dependencies.

Closes #816.

Notes:
* `minidump-walk-stack` isn't a great name. Pending a decision in #824 it may be changed to be clearer.
* Do we want to re-export `minidump-walk-stack` in `minidump-processor` so that the latter keeps the same API as before (and as a convenience to users)?


gabrielesvelto (approved):
This is fine I've just been wondering a bit about the naming especially WRT to https://github.com/rust-minidump/rust-minidump/issues/824. Renaming a crate is not a big deal, we could easily move `minidump-stackwalk` to `minidump-analyzer` and put a forwarding crate in its place to guarantee that the new crate is picked up. However using the old name of a crate for a new one - such as this one - is problematic as AFAIK cargo wouldn't be able to support it.

So here's my proposal, why don't we call the new crate `minidump-unwind` to clearly distinguish it from `minidump-stackwalk`? We can always rename `minidump-stackwalk` to `minidump-analyzer` later to make the distinction clearer.

Aside from this the PR is fine.

afranchuk:
Okay, I've renamed the crate to `minidump-unwind`.

afranchuk:
I don't know why CI is failing (invalid cache?).

gabrielesvelto:
> I don't know why CI is failing (invalid cache?).

The folder is still called `minidump-walk-stack`, it's looking for `minidump-unwind` instead.

afranchuk:
> > I don't know why CI is failing (invalid cache?).
> 
> The folder is still called `minidump-walk-stack`, it's looking for `minidump-unwind` instead.

:facepalm: didn't add the _new_ folder (it just removed the old one). Probably should use `git mv` more often.

gabrielesvelto:
Merging. Once we're done with the work here I'll make a major version release since we're adding an entirely new crate.

PR #826, opened 2023-03-29
Replace `encoding` with `encoding_rs`

Jake-Shadle:
`encoding` is [unmaintained](https://rustsec.org/advisories/RUSTSEC-2021-0153.html), this PR just replaces its usage in `minidump` with `encoding_rs` and completely gets rid of it in `minidump-synth` in favor of just using `std::str::encode_utf16`

### DIFF
--- a/minidump/src/minidump.rs
+++ b/minidump/src/minidump.rs
@@ -2,8 +2,6 @@
 // file at the top-level directory of this distribution.
 
 use debugid::{CodeId, DebugId};
-use encoding::all::{UTF_16BE, UTF_16LE};
-use encoding::{DecoderTrap, Encoding};
 use memmap2::Mmap;
 use num_traits::FromPrimitive;
 use scroll::ctx::{SizeWith, TryFromCtx};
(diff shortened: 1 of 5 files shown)


gabrielesvelto (approved):
LGTM, can I merge this or is there something else missing?

Jake-Shadle:
Nope this is everything AFAIK.

minidump/src/minidump.rs
```diff
@@ -4636,7 +4634,9 @@ fn utf16_to_string(data: &[u16]) -> Option<String> {
     let len = data.iter().take_while(|c| **c != 0).count();
     let s16 = &data[..len];
     let bytes = unsafe { slice::from_raw_parts(s16.as_ptr() as *const u8, s16.len() * 2) };
-    UTF_16LE.decode(bytes, DecoderTrap::Strict).ok()
+    encoding_rs::UTF_16LE
```
luser, line 4637: Is this code correct? It's assuming LE.

Jake-Shadle: I think this place and minidump-synth both assumed LE, I thought it was better to just keep the code mostly as it was before the change, but I just assumed it was because....no one really cares about BE any longer since there's no one running BE in a meaningful capacity.

PR #828, opened 2023-04-10
Start to fix DebugInfoSymbolProvider

sfackler:
This PR fixes two bugs that prevented DebugInfoSymbolProvider from ever resolving an address to a function:

1. `partition_point` expects a closure that returns true at the head of the slice and false at the tail, but the closure originally provided did the opposite. This meant that `partition_point` always returned `0`, and `function_by_address` would therefore always return `None`.
2. `fill_symbol` was passing the raw instruction address into `function_by_address` rather than the debiased address required to resolve properly against the function addresses embedded in the binary.

As an example, here's before and after output of running minidump-stackwalk with `--use-local-debuginfo` on a minidump produced by the `diskwrite` example in the crash-handling repo: https://gist.github.com/sfackler/578de5039b6734ba4b7ce8044ca5a2d0

The function resolution process still seems a bit broken, but this gets it closer.

### DIFF
--- a/minidump-unwind/src/symbols.rs
+++ b/minidump-unwind/src/symbols.rs
@@ -438,7 +438,7 @@ pub mod debuginfo {
         pub fn function_by_address(&self, addr: u64) -> Option<&Function> {
             // Get the index of the first function that starts after our search address.
             let functions = &self.functions;
-            let index_after_location = functions.partition_point(|f| addr < f.address);
+            let index_after_location = functions.partition_point(|f| addr >= f.address);
             if index_after_location == 0 {
                 return None;
             }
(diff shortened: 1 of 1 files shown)


sfackler:
Rebased

gabrielesvelto:
Thanks for looking into this, I've done some more testing w/ and w/o the patch on a local build of Firefox and I noticed some side-effects of this change. The frames found via CFI seem to have the proper symbols, but the symbols aren't demangled (I'll open a separate issue for this), this is down to the code being buggy and not being able to find the symbols w/o the patch. Additionally the stack walker seems to find a lot of inlined frames that just aren't on the stack (with most making no sense at all). This might indicate a problem while reading inline frames from the native debuginfo.

afranchuk:
@sfackler thanks for finding this, the component for which this functionality was written is still being developed so I haven't had an opportunity to end-to-end test it (though I should have tested this on its own; I don't yet have a streamlined process for generating local minidumps but now is as good a time as any to set up a test environment). I will poke at it more now.

gabrielesvelto:
I'm merging this so we can land further fixes on top of it.

PR #829, opened 2023-04-25
Bug 1663983 - Check for memory accesses to likely guard pages.

afranchuk:
This is useful for classifying exceptions (e.g., as buffer overflows).


gabrielesvelto (approved):
LGTM

PR #835, opened 2023-05-03
Fix local debuginfo inline functions and line info

### LINKED ISSUE #832: When using local debug information inline information appears to be wrong
I've tested unwinding local debug information with a local build of Firefox and the resulting stack traces have incoherent inlining information. It appears to be scattered all over the frames so it's likely that we're reading the inline information incorrectly. Here's how to reproduce:
- Make a local build of Firefox
- Generate symbol files with `./mach buildsymbols`
- Run Firefox with `./mach run --enable-crash-reporter`, browse to *about:crashcontent*
- Do not interact with the crash reporter UI, instead copy the minidump out of `~/.mozilla/firefox/Crash Reports/pending`
- Close firefox
- Unpack the archive with the symbol files that have been generated in Firefox' object dir under `dist/firefox-114.0a1.en-US.linux-x86_64.crashreporter-symbols.zip`
- Now run `minidump-stackwalk <minidump> <path_to_symbol_files>` and `minidump-stackwalk --use-local-debuginfo <minidump>`, compare the resulting output

`dump_syms` processes inline information (see [this](https://github.com/mozilla/dump_syms/blob/fa4caf72ff6ca07c0bbf29cecf693c6294193563/src/collector.rs#L164-L340)) so we most likely have to do the same to get coherent stacks between the two unwinders.

afranchuk:
Closes #832.

Relies on #834.

### DIFF
--- a/minidump-unwind/src/symbols.rs
+++ b/minidump-unwind/src/symbols.rs
@@ -396,7 +396,8 @@ pub mod debuginfo {
     }
 
     struct DebugInfo {
-        functions: Vec<Function>,
+        // Sorted by function address, mutually exclusive
+        functions: AddressRanges<Function>,
         unwind_symbol_file: Option<SymbolFile>,
     }
 
@@ -412,7 +413,7 @@ pub mod debuginfo {
         }
 
         pub fn from_object(object: Object) -> Self {
-            let mut functions: Vec<Function> = object
+            let functions = object
                 .debug_session()
                 .ok()
                 .map(|session| {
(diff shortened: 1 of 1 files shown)


gabrielesvelto:
This generally LGTM but in the JSON output the `function_offset` fields appear to be wrong. Specifically, it seems that they're filled with the function address, not the function offset (i.e. the function address minus the module base address).

afranchuk:
That's odd, that shouldn't be from symbolic since that is opening the modules directly. I'll check and see how that's occurring.

afranchuk:
Ah, it looks like it needs the module base address _added_ to the function offset in the trait interface.

gabrielesvelto (changes requested):
Looking good, but can you rebase this so I can merge it?

afranchuk:
> Looking good, but can you rebase this so I can merge it?

Done!

PR #842, opened 2023-05-17
setup cargo-dist

Gankra:
First commit is literally just me checking in the result of that command.

Second commit is me hand-editing the old hack I made to use the modern version cargo-dist understands (if you cut a prerelease it will use the contents of the "Unreleased" section without actually mutating RELEASES.md).

This drops support for arm64 linux, which was broken anyway.

### DIFF
--- a/minidump-stackwalk/Cargo.toml
+++ b/minidump-stackwalk/Cargo.toml
@@ -38,6 +38,5 @@ test-assembler = "0.1.6"
 
 [package.metadata.release]
 pre-release-replacements = [
-  {file="../RELEASES.md", search="Next Version", replace="Version {{version}} ({{date}})", min=1},
-  {file="../RELEASES.md", search="<!-- next-header -->", replace="<!-- next-header -->\n# Next Version\n", exactly=1},
+  {file="../RELEASES.md", search="# Unreleased", replace="# Unreleased\n\nNothing Yet!\n\n\n# Version {{version}} ({{date}})", exactly=1},
 ]
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,50 +1,143 @@
+# CI that:
+#
+# * checks for a Git Tag that looks like a release
+# * creates a Github Release™ and fills in its text
+# * builds artifacts with cargo-dist (executable-zips, installers)
+# * uploads those artifacts to the Github Release™
+#
+# Note that the Github Release™ will be created before the artifacts,
+# so there will be a few minutes where the release has no artifacts
+# and then they will slowly trickle in, possibly failing. To make
+# this more pleasant we mark the release as a "draft" until all
+# artifacts have been successfully uploaded. This allows you to
+# choose what to do with partial successes and avoids spamming
+# anyone with notifications before the release is actually ready.
 name: Release
 
 permissions:
   contents: write
 
+# This task will run whenever you push a git tag that looks like a version
+# like "v1", "v1.2.0", "v0.1.0-prerelease01", "my-app-v1.0.0", etc.
+# The version will be roughly parsed as ({PACKAGE_NAME}-)?v{VERSION}, where
+# PACKAGE_NAME must be the name of a Cargo package in your workspace, and VERSION
+# must be a Cargo-style SemVer Version.
+#
+# If PACKAGE_NAME is specified, then we will create a Github Release™ for that
+# package (erroring out if it doesn't have the given version or isn't cargo-dist-able).
+#
+# If PACKAGE_NAME isn't specified, then we will create a Github Release™ for all
+# (cargo-dist-able) packages in the workspace with that version (this is mode is
+# intended for workspaces with only one dist-able package, or with all dist-able
+# packages versioned/released in lockstep).
+#
+# If you push multiple tags at once, separate instances of this workflow will
+# spin up, creating an independent Github Release™ for each one.
+#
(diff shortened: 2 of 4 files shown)


Gankra:
So just to clarify the implications of setting up [cargo-dist](https://opensource.axo.dev/cargo-dist/):

TL;DR: everything you used to do still works, because cargo-dist is based on the release-engineering experiments I originally did in this project. That is, you still cut releases with `cargo release`, which is just sugar for pushing up a tagged commit.

You just can now run the tool `cargo dist` to help you audit/replicate/update the release process locally, and your release process will get better "for free" as I improve cargo-dist. If you have any issues with cargo-dist you can message me on the mozilla matrix server or [the axo.dev discord server](https://discord.gg/DEPqCk8P) and I'll fix it / help you. You are effectively beta-testing my tool, (which this project was already sort of alpha-testing by being the tools precursor...).



 # cargo-dist basics

cargo-dist is both:

* a wrapper over `cargo build` that handles making zips/tarballs/installers/checksums
* taskcluster-for-orchestrating-itself (it generates its own CI scripts)

The purpose of making cargo-dist aware of CI is so you can reproduce its behaviour locally. If you run `cargo dist plan` on this repo you will get a report of what cargo-dist thinks is going on, and what it would want to build in CI:

image

Each machine in the CI basically just runs `cargo dist` exactly once and then uploads the artifacts it produces. If you're on a similar machine (target-triple) and you copy the `cargo dist` invocation CI did, you should ideally get the same result (not bit-identical, but to the kind of standard we expect from `cargo build`), making it much easier to debug and reproduce CI issues (`cargo dist plan` should in fact have identical behaviour on all systems, and if it doesn't let me know).

If you run `cargo dist init` on this repo you will get an interactive updater for cargo-dist-itself and [your config](https://opensource.axo.dev/cargo-dist/book/config.html), which will regenerate the CI script at the end (`-y` auto-accepts all recommendations). This is how I setup cargo-dist.



# detailed implications of this change

* The same `cargo release` workflow I originally setup on this repo [still works](https://opensource.axo.dev/cargo-dist/book/cargo-release-guide.html#virtual-workspace-with-unified-versions) and is the intended method of doing releases. The docs I just linked also include details on how you can use PRs now but I don't Love that design yet.
* There is now a "dist" cargo build profile (that you can edit however you want) that turns up LTO a bit more. When you build with cargo-dist (`cargo dist build`) this profile will be used. You can manually do this with `cargo build --profile=dist`.
   * cargo-dist will also force [crt-static](https://rust-lang.github.io/rfcs/1721-crt-static.html) when targetting windows, this will not be reproduced by plan `cargo build`.
   *  In the future [we'll push this profile to look more like firefox release builds with split debuginfo](https://github.com/axodotdev/cargo-dist/milestone/3).
* The github workflow that's triggered now creates the Github Release as a draft, and will only un-draft the release if all build tasks complete succesfully. This lets you decide what to do with a failed build (you can manually undraft it). 
  * Note that the URLs to download artifacts are different in this draft mode.
  * Currently I group the two macos builds into one machine, theoretically saving resources but increasing the latency of a release. I am increasingly disliking this because I'm Impatient.
* The aarch64-unknown-linux-gnu target has been removed, as cargo-dist [does not yet support non-trivial cross-compiles](https://github.com/axodotdev/cargo-dist/issues/74) (arm64 mac is supported because apple makes it so easy). I consider this regression acceptable because I only added this target Because I Could and it's currently broken on this project anyway.
* tar.gz has been changed to tar.xz. This is the default of cargo-dist, but you [can change it back](https://opensource.axo.dev/cargo-dist/book/config.html#unix-archive)
* Archives now are paired with .sha256 checksum files. This is experimental, and one of the features you get for using a 0.1.0-prerelease and not the stable 0.0.7.
* If you cut a prerelease the auto-generated release notes will pick up the "Unreleased" section in the changelog.
* The Github Release will include some auto-generated download/install tables
* The Github Release will include [a dist-manifest.json](https://opensource.axo.dev/cargo-dist/book/schema.html) describing the release/announcement in a machine-readable format (this is just `cargo dist plan -ojson`). This can be parsed with [cargo-dist-schema](https://docs.rs/cargo-dist-schema/0.0.7/cargo_dist_schema/).
* I have not enabled any to start, but you will be able to introduce auto-generated [installers](https://opensource.axo.dev/cargo-dist/book/installers.html) with minimal effort.

minidump-stackwalk/Cargo.toml
```diff
@@ -38,6 +38,5 @@ test-assembler = "0.1.6"
 
 [package.metadata.release]
 pre-release-replacements = [
-  {file="../RELEASES.md", search="Next Version", replace="Version {{version}} ({{date}})", min=1},
-  {file="../RELEASES.md", search="<!-- next-header -->", replace="<!-- next-header -->\n# Next Version\n", exactly=1},
-]
+  {file="../CHANGELOG.md", search="# Unreleased", replace="# Unreleased\n\nNothing Yet!\n\n\n# Version {{version}} ({{date}})", exactly=1},
```
gabrielesvelto, line 41: This should still be `../RELEASES.md` right? Or did you want to rename `RELEASES.md` to `CHANGELOG.md`?

gabrielesvelto (changes requested):

Gankra:
This should be ready to land

gabrielesvelto (approved):

PR #843, opened 2023-05-17
setup oranda

Gankra:
this is more manual, as oranda doesn't yet have an init command.

this is based on top of #842 and requires https://github.com/axodotdev/oranda/pull/322 before merging


Gankra:
We're approaching code freeze for the next release of oranda this week, and part of that is reworking a ton of config/theme stuff, so I'm going to hold off on this for a bit to avoid needless churn.

PR #844, opened 2023-05-18
don't warn on duplicate UnusedStream type streams

demoray:
This change ignores warning about duplicate `UnusedStream` streams, as long as both streams are empty, as this is extremely common for dumps collected by WER.


gabrielesvelto:
Thanks!

PR #845, opened 2023-05-24
Remove tokio to reduce the audit surface.

afranchuk:
It was minimally used.

### DIFF
--- a/breakpad-symbols/src/lib.rs
+++ b/breakpad-symbols/src/lib.rs
@@ -593,7 +593,7 @@ impl<T, E> CachedAsyncResult<T, E> {
         if guard.is_none() {
             *guard = Some(Arc::new(f().await));
         }
-        unsafe { guard.as_ref().unwrap_unchecked() }.clone()
+        guard.as_ref().unwrap().clone()
     }
 }
 
--- a/breakpad-symbols/src/lib.rs
+++ b/breakpad-symbols/src/lib.rs
@@ -38,7 +38,9 @@
 //! ```
 
 use async_trait::async_trait;
+use cachemap2::CacheMap;
 use debugid::{CodeId, DebugId};
+use futures_util::lock::Mutex as FutMutex;
 use tracing::trace;
 
 use std::boxed::Box;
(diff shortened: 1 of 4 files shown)


Gankra:
Does this really change much given minidump-stackwalk uses tokio? Also is really no one in the mozilla cargo-vet trust web using tokio?

afranchuk:
Yes! We're using `minidump-unwind` on its own (for the minidump analyzer in crash reporting). Nothing delivered from mozilla central uses tokio, so otherwise there would need to be an audit of tokio to ensure it's `safe-to-deploy`.

We _could_ actually do the audit, and we have discussed possibly trusting it ourselves. No imports in the cargo-vet registry have an audit for tokio yet. It is preferable to remove it as this is the only use in the entire dependency tree. Auditing it will be a big task; it has 500+ unsafe blocks.

Gankra:
Yeah I agree it's a ton of work, just a bit surprised it's managed to evade auditing this long. Makes sense, assuming the perf still behaves reasonably (I would assume so, code is generally bottlenecked on xul.dll's symbols sync downloading, async'ing just lets the other stuff get fetched at the same time).

afranchuk:
I expect there's a lot of `tokio` exemptions :cry: 

The new code here should be almost the same as the old code as far as perf goes. I could have written it using atomics and whatnot (actually I have an impl lying around somewhere I could copy in) but I figured it's not a high priority.

breakpad-symbols/src/lib.rs
```diff
@@ -569,6 +571,32 @@ fn module_key(module: &(dyn Module + Sync)) -> ModuleKey {
...
+    pub async fn get<'a, F, Fut>(&self, f: F) -> Arc<Result<T, E>>
+    where
+        F: FnOnce() -> Fut + 'a,
+        Fut: std::future::Future<Output = Result<T, E>> + 'a,
+    {
+        let mut guard = self.inner.lock().await;
+        if guard.is_none() {
+            *guard = Some(Arc::new(f().await));
+        }
+        unsafe { guard.as_ref().unwrap_unchecked() }.clone()
```
Gankra, line 596: there's no way this unsafe is justified, just use normal unwrap. the compiler will probably optimize it out anyway, and it won't be a fun landmine for a refactor.

Gankra (approved):

gabrielesvelto (approved):
I'm OK with this w/ the change @Gankra requested

breakpad-symbols/src/lib.rs
afranchuk: Whoops, definitely not justified.

PR #846, opened 2023-05-26
Add support for MozMacosBootargsStream

steven-michaud:
Recently I [landed a patch](https://bugzilla.mozilla.org/show_bug.cgi?id=1833680) to make Mozilla's Breakpad add kernel boot args streams to the minidumps it creates on macOS. This pull request allows rust-minidump to read these streams. I've tested it with minidumps created using a tryserver Firefox build containing my Breakpad patch. I saw no problems.

Kernel boot args seem to be specific to macOS (or maybe to the Mach kernel). They can be set by users with administrator privileges, and by apps (like installers) that request an administrator password. They change the kernel's behavior in ways that can effect apps running in userland. This has happened a number of times with Firefox. It'd be good to have this information in crash reports.

Breakpad includes a MozMacosBootargsStream in all the minidumps it creates on macOS. If there are no kernel boot args, the value is an empty string. I'd have liked to only include a MozMacosBootargsStream if the kernel boot args are non-empty, but Breakpad doesn't support that.


mstange:
The "fmt" failure means you need to run `cargo fmt`. The other test failures are "insta" snapshot test failures, see https://insta.rs/docs/cli/

steven-michaud:
Thanks, Markus.

I ran `cargo fmt` and it made some changes. Then I installed `cargo-insta` and ran `cargo-insta review`. It returned `done: no snapshots to review`. Does that mean that I just need to submit the changes made by `cargo fmt`?

mstange:
I think it means that you first need to run `cargo insta test` so that it knows about the failing tests.

mstange:
> `cargo-insta test` returned no failures.

Hmm, I'm surprised by that. Oh, maybe you need `--workspace`? [In the failure log](https://github.com/rust-minidump/rust-minidump/actions/runs/5092030882/jobs/9153125154?pr=846#step:5:12), you can see that it's running `cargo test --workspace`.

steven-michaud:
It looks like I need to edit the "snapshots" in `minidump-stackwalk/tests/snapshots`, to make them conform to the (slightly) changed output my patch produces. Is that kosher, or do I need to follow some fancy procedure to regenerate the snapshots?

mstange:
`cargo insta review` should be that procedure.

steven-michaud:
OK, now I understand ... or at least I think I do.

After running `cargo-insta review` (or maybe it was `cargo-insta test`, I see a bunch of files in `minidump-stackwalk/tests/snapshots` named `*.snap` and others named `*.snap.new`. I suppose I need to replace the first set with the second (suitably renamed).

mstange:
The purpose of `cargo insta review` is to do all the renaming / replacing for you.

steven-michaud:
Running `cargo insta review` (instead of `cargo-insta review` did what you reported. Now `cargo test` reports no failures.

mstange:
Great! And wow, I would not have expected that to make a difference, I thought `cargo x` was fully equivalent to `cargo-x` (if `cargo-x` exists).

PR #852, opened 2023-06-26
Use the code file on win64 for unwind information.

afranchuk:
The debug file may not be available, but even if it is, it doesn't contain unwind info.


gabrielesvelto (approved):
LGTM, you just need to appease clippy

PR #861, opened 2023-07-30
Generate less bloaty code for `errors::windows` enums

Swatinem:
Not really intended to be merged, feel free to ignore.

This is just an experiment to see if custom `Debug` and `from_u64/u32` code based on PHF leads to better codegen.
On my windows system, a full lto compile using `cargo bloat` shrinks the resulting binary of `minidump-stackwalk` by 100k, (from 10M to 9.9M), with a tiny regression to the debug data size.

The full story of what I did is documented here: https://swatinem.de/blog/optimizing-enums/

Ideally the Rust compiler would do this optimization internally and not generate such code in the first place.


gabrielesvelto:
This is a nice improvement! I don't mind merging a bunch of generated code but would you prefer waiting to see if the compiler issue gets some traction first?

PR #868, opened 2023-09-07
Update bitflags to version 2

glandium:


glandium:
The failures don't seem to be mine.

glandium:
Let's go with a semver compatible change first.

PR #869, opened 2023-09-08
Avoid relying on bitflags internals

glandium:
This will allow to use a bitflags 2-derived bitflags 1 implementation in mozilla-central.


gabrielesvelto:
There's some clippy & fmt issues to address but the rest seems to work

glandium:
The fmt and clippy issues are on main.

gabrielesvelto:
Sorry for the delay, I'm going to fix those in the main branch then merge

PR #872, opened 2023-09-12
Lookup debug info using code info

### LINKED ISSUE #870: implement /code_file/code_id/sym_file symbol download attempt
In [bug 1746940](https://bugzilla.mozilla.org/show_bug.cgi?id=1746940), I changed the download API for the Mozilla Symbol Server such that it can handle `/code_file/code_id/sym_file` requests. If it has a record with that `code_file` / `code_id` combination, it'll return an HTTP 302 redirect to the relevant `/debug_file/debug_id/sym_file` url.

Records are created during symbol upload. For symbol files, Mozilla Symbol Server will parse the header consisting of `MODULE` and `INFO` lines and capture `debug_filename`, `debug_id`, `code_file`, `code_id`, and `generator` information. I need to figure out how to backfill the information. I'll probably do it item-by-item based on the missing symbols emails.

Anyhow, this is currently deployed to the stage environment. I'm planning to deploy it to the prod environment next week.

This issue covers changing the rust-minidump stackwalker to try downloading the symbol using `/code_file/code_id/sym_file` in the case where the debug filename and debug id are missing or bogus.

willkg:
When fetching a symbol file from an HTTP symbol supplier and the debug filename and debug identifier are missing, try looking up the debug information using the code info. If the lookup is successful, use that debug info to fetch the symbol file from the cache and symbol servers.

This uses a new feature in Mozilla Symbols Server where requesting a download with:

```
/<code_file>/<code_id>/<code_file>.sym
```

will return an HTTP 302 with the download API url using the debug file and debug id if the Mozilla Symbols Server has a record for that file.

Fixes #870

### DIFF
--- a/breakpad-symbols/src/http.rs
+++ b/breakpad-symbols/src/http.rs
@@ -2,7 +2,7 @@
 
 use crate::*;
 use cachemap2::CacheMap;
-use reqwest::{Client, Url, redirect};
+use reqwest::{redirect, Client, Url};
 use std::fs;
 use std::io::{self, Write};
 use std::path::{Path, PathBuf};
@@ -223,26 +223,25 @@ async fn lookup_debug_info_by_code_info(
             .unwrap();
 
         let response = no_redirects_client.get(url.clone()).send().await;
-        match response {
-            Ok(res) => {
-                let res_status = res.status();
-                if res_status == reqwest::StatusCode::FOUND || res_status == reqwest::StatusCode::MOVED_PERMANENTLY {
-                    let mut new_url = res.headers().get("Location").unwrap().to_str().unwrap();
-                    if new_url.starts_with("/") {
-                        new_url = new_url.strip_prefix("/").unwrap();
-                    }
-                    let mut parts = new_url.split("/");
-                    let debug_file = String::from(parts.next().unwrap());
-                    let debug_identifier = DebugId::from_str(parts.next().unwrap()).unwrap();
-
-                    debug!("Found debug info {} {}", debug_file, debug_identifier);
-                    return Some(DebugInfoResult {
-                        debug_file: debug_file,
-                        debug_identifier: debug_identifier,
-                    });
+        if let Ok(res) = response {
+            let res_status = res.status();
+            if res_status == reqwest::StatusCode::FOUND
+                || res_status == reqwest::StatusCode::MOVED_PERMANENTLY
+            {
+                let mut new_url = res.headers().get("Location").unwrap().to_str().unwrap();
+                if new_url.starts_with('/') {
+                    new_url = new_url.strip_prefix('/').unwrap();
                 }
-            },
-            _ => (),
+                let mut parts = new_url.split('/');
+                let debug_file = String::from(parts.next().unwrap());
+                let debug_identifier = DebugId::from_str(parts.next().unwrap()).unwrap();
+
+                debug!("Found debug info {} {}", debug_file, debug_identifier);
+                return Some(DebugInfoResult {
+                    debug_file,
+                    debug_identifier,
+                });
+            }
         }
     }
 
@@ -545,12 +544,11 @@ impl SymbolSupplier for HttpSymbolSupplier {
 
         if debug_file.to_string().is_empty() || debug_id.is_nil() {
             debug!("Missing debug file or debug identifier--trying lookup with code info");
-            match lookup_debug_info_by_code_info(&self.urls, module).await {
-                Some(debug_info_result) => {
-                    debug_file = debug_info_result.debug_file;
-                    debug_id = debug_info_result.debug_identifier;
-                },
-                _ => (),
+            if let Some(debug_info_result) =
+                lookup_debug_info_by_code_info(&self.urls, module).await
+            {
+                debug_file = debug_info_result.debug_file;
+                debug_id = debug_info_result.debug_identifier;
             }
         }
 
@@ -584,7 +582,8 @@ impl SymbolSupplier for HttpSymbolSupplier {
         // Second: try to directly download sym files
         for url in &self.urls {
             // First, try to get a breakpad .sym file from the symbol server
-            let sym = fetch_symbol_file(&self.client, url, &lookup_module, &self.cache, &self.tmp).await;
+            let sym =
+                fetch_symbol_file(&self.client, url, &lookup_module, &self.cache, &self.tmp).await;
             match sym {
                 Ok(file) => {
                     trace!("HttpSymbolSupplier parsed file!");
@@ -601,7 +600,10 @@ impl SymbolSupplier for HttpSymbolSupplier {
             trace!("symbols: trying to fetch native symbols");
             // Find native files
             let mut native_artifacts = vec![];
-            native_artifacts.push(self.locate_file_internal(&lookup_module, FileKind::Binary).await);
+            native_artifacts.push(
+                self.locate_file_internal(&lookup_module, FileKind::Binary)
+                    .await,
+            );
             native_artifacts.push(
                 self.locate_file_internal(module, FileKind::ExtraDebugInfo)
                     .await,
--- a/breakpad-symbols/src/http.rs
+++ b/breakpad-symbols/src/http.rs
@@ -190,6 +190,48 @@ pub struct DebugInfoResult {
     debug_identifier: DebugId,
 }
 
+/// Perform a code_file/code_identifier lookup for a specific symbol server.
+async fn individual_lookup_debug_info_by_code_info(
+    base_url: &Url,
+    lookup_path: &str,
+) -> Option<DebugInfoResult> {
+    let url = base_url.join(lookup_path).ok()?;
+
+    debug!("Trying code file / code identifier lookup: {}", url);
+
+    // This should not follow redirects--we want the next url if there is one
+    let no_redirects_client = Client::builder()
+        .redirect(redirect::Policy::none())
+        .build()
+        .unwrap();
+
+    let response = no_redirects_client.get(url.clone()).send().await;
+    if let Ok(res) = response {
+        let res_status = res.status();
+        if res_status == reqwest::StatusCode::FOUND
+            || res_status == reqwest::StatusCode::MOVED_PERMANENTLY
+        {
+            let location_header = res.headers().get("Location")?;
+            let mut new_url = location_header.to_str().ok()?;
+            if new_url.starts_with('/') {
+                new_url = new_url.strip_prefix('/').unwrap();
+            }
+            let mut parts = new_url.split('/');
+            let debug_file_part = parts.next()?;
+            let debug_file = String::from(debug_file_part);
+            let debug_identifier_part = parts.next()?;
+            let debug_identifier = DebugId::from_str(debug_identifier_part).ok()?;
+            debug!("Found debug info {} {}", debug_file, debug_identifier);
+            return Some(DebugInfoResult {
+                debug_file,
+                debug_identifier,
+            });
+        }
+    }
+
+    None
+}
+
 /// Given a vector of symbol urls and a module with a code_file and code_identifier,
 /// this tries to request a symbol file using the code file and code identifier.
 ///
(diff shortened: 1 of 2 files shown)


willkg:
@gabrielesvelto Can you review this please? I tried to keep the changes globbed so they're not scattered all over. I couldn't figure out how to write tests for the lookup, but I tested it with 100 Windows crash reports plus the crash report I added code file and code identifier data for the `xul.dll` modules on the symbols server stage environment:

https://symbol.stage.mozaws.net/

```bash
crashid=421408bf-aa63-4e5e-8af9-8ee3a0230907
fetch-data --dumps crashdata "${crashid}"
target/debug/minidump-stackwalk \
        --no-color \
        --verbose=debug \
        --symbols-cache=~/tmp/ \
        --symbols-tmp=~/tmp/ \
        --symbols-url="https://symbols.stage.mozaws.net/" \
        --output-file="output.${crashid}.json" \
        --json \
        "crashdata/upload_file_minidump/${crashid}" \
        > "debug.${crashid}.json" \
        2>&1
```

willkg:
I'll fix the fmt and clippy issues.

breakpad-symbols/src/http.rs
```diff
@@ -470,8 +538,43 @@ impl SymbolSupplier for HttpSymbolSupplier {
...
+        if debug_file.to_string().is_empty() || debug_id.is_nil() {
+            debug!("Missing debug file or debug identifier--trying lookup with code info");
+            match lookup_debug_info_by_code_info(&self.urls, module).await {
+                Some(debug_info_result) => {
+                    debug_file = debug_info_result.debug_file;
+                    debug_id = debug_info_result.debug_identifier;
+                },
+                _ => (),
+            }
+        }
```
willkg, line 555: In order for the lookup to help with the cache, local files, and symbol servers, we have to do the lookup and then carry the new debug file and debug id through all the other functions. So I did the lookup first if the debug file or debug identifier is empty. Then I carried the (possibly new) debug file and debug identifier through all the other functions by putting the needed data in a `SimpleModule`. It was kind of painful because `SimpleModule` isn't like minidump `Module`.

If you have ideas on how to do this better, I'm all ears.

breakpad-symbols/src/http.rs
```diff
@@ -181,6 +182,73 @@ fn commit_cache_file(mut temp: NamedTempFile, final_path: &Path, url: &Url) -> i
...
+
+        debug!("Trying code file / code identifier lookup: {}", url);
+
+        // This should not follow redirects--we want the next url if there is one
+        let no_redirects_client = Client::builder()
+            .redirect(redirect::Policy::none())
+            .build()
+            .unwrap();
+
+        let response = no_redirects_client.get(url.clone()).send().await;
```
willkg, line 225: I'm on the fence about this API. On the one hand, it's pretty convenient for other use cases to support requesting by code file and code id using the download API and being able to follow redirects to the final file in S3. On the other hand, it's pretty wonky for this case. Maybe it'd be better to have a new lookup API where you give it a code file and code identifier by querystring params and it returns a debug file and debug identifier in a JSON payload.

We could support multiple ways to get the information.

Are there other things we might use this functionality for that would inform what we should do?

breakpad-symbols/src/http.rs
```diff
@@ -470,8 +537,42 @@ impl SymbolSupplier for HttpSymbolSupplier {
         &self,
         module: &(dyn Module + Sync),
     ) -> Result<SymbolFile, SymbolError> {
+        // If we don't have a debug_file or debug_identifier, then try to get it
+        // from a symbol server.
+        let mut debug_file = module.debug_file().unwrap_or_default().into_owned();
+        let mut debug_id = module.debug_identifier().unwrap_or_default();
```
gabrielesvelto, line 543: `SimpleModule` fields are all `Option<>` objects, so you could keep `debug_file` and `debug_id` options too, like:
```rust
        let mut debug_file = module.debug_file().map(|name| name.into_owned());
        let mut debug_id = module.debug_identifier();
```

breakpad-symbols/src/http.rs
```diff
@@ -470,8 +537,42 @@ impl SymbolSupplier for HttpSymbolSupplier {
...
+
+        if debug_file.to_string().is_empty() || debug_id.is_nil() {
```
gabrielesvelto, line 545: And then this would become:
```rust
        if debug_file.is_none() || debug_id.is_none() {
```

breakpad-symbols/src/http.rs
```diff
@@ -470,8 +537,42 @@ impl SymbolSupplier for HttpSymbolSupplier {
...
+            debug!("Missing debug file or debug identifier--trying lookup with code info");
+            if let Some(debug_info_result) =
+                lookup_debug_info_by_code_info(&self.urls, module).await
+            {
+                debug_file = debug_info_result.debug_file;
+                debug_id = debug_info_result.debug_identifier;
```
gabrielesvelto, line 551: And finally this would become:
```rust
                debug_file = Some(debug_info_result.debug_file);
                debug_id = Some(debug_info_result.debug_identifier);
```

breakpad-symbols/src/http.rs
```diff
@@ -470,8 +537,42 @@ impl SymbolSupplier for HttpSymbolSupplier {
...
+            },
+            debug_id: Some(debug_id),
+            code_file: if code_file.to_string().is_empty() {
+                None
+            } else {
+                Some(code_file)
+            },
+            code_identifier: module.code_identifier(),
+            ..SimpleModule::default()
+        };
```
gabrielesvelto, line 572: The above would allow you to populate `debug_file` and `debug_id` directly without the if-else blocks. Given this is a constructor you could then factor this out into the `SimpleModule` implementation.

gabrielesvelto (changes requested):

willkg:
I'm planning to get to this after I finish fixing issues with the download API in Tecken.

willkg:
^^^ That rebases against main tip and fixes code review comments.

With crash report [421408bf-aa63-4e5e-8af9-8ee3a0230907](https://crash-stats.mozilla.org/report/index/421408bf-aa63-4e5e-8af9-8ee3a0230907) if you run the stackwalker using the prod symbols server which is missing information for `xul.pdb/1BCB86F9A4FEEC254C4C44205044422E1`, it produces this stack:

```
 0  xul.dll + 0x6f5a39  
 1  xul.dll + 0x186f13d  
 2  xul.dll + 0xeca69a  
 3  xul.dll + 0x1ace297  
 4  xul.dll + 0xec6964  
 5  xul.dll + 0x1ad6dc1  
 6  xul.dll + 0xeb0621  
 7  xul.dll + 0xeb0446  
 8  xul.dll + 0x20a1f8  
 9  xul.dll + 0xe93c50  
10  xul.dll + 0xe9389c  
...
```

If you run the stackwalker using the stage symbols server which does have debug file, debug id, code file, code id information for `xul.pdb/1BCB86F9A4FEEC254C4C44205044422E1`, it produces this stack:

```
 0  xul.dll!NS_ABORT_OOM(unsigned int) [nsDebugImpl.cpp:27de51e74c19186044da535125895c96fc3a0f23 : 674 + 0x0]  
 1  xul.dll!mozilla::ArenaAllocator<32768,8>::Allocate(unsigned int) [ArenaAllocator.h:27de51e74c19186044da535125895c96fc3a0f23 : 72]
 2  xul.dll!nsPresArena<32768,mozilla::DisplayListArenaObjectId,89>::Allocate(mozilla::DisplayListArenaObjectId, unsigned int) [nsPresArena.cpp:27de51e74c19186044da535125895c96fc3a0f23 : 52 + 0x3f]    
 3  xul.dll!mozilla::nsDisplayListBuilder::Allocate(unsigned int, mozilla::DisplayListArenaObjectId) [nsDisplayList.h:27de51e74c19186044da535125895c96fc3a0f23 : 913]    
 4  xul.dll!mozilla::nsDisplayListBuilder::Allocate(unsigned int, DisplayItemType) [nsDisplayList.h:27de51e74c19186044da535125895c96fc3a0f23 : 928]    
 5  xul.dll!mozilla::nsDisplayCanvasBackgroundColor::operator new(unsigned int, mozilla::nsDisplayListBuilder*) [nsCanvasFrame.h:27de51e74c19186044da535125895c96fc3a0f23 : 155]    
 6  xul.dll!mozilla::MakeDisplayItemWithIndex(mozilla::nsDisplayListBuilder*, nsCanvasFrame*, const unsigned short) [nsDisplayList.h:27de51e74c19186044da535125895c96fc3a0f23 : 1970]    
 7  xul.dll!mozilla::nsDisplayList::AppendNewToTopWithIndex(mozilla::nsDisplayListBuilder*, nsCanvasFrame*, const unsigned short) [nsDisplayList.h:27de51e74c19186044da535125895c96fc3a0f23 : 3079]    
 8  xul.dll!mozilla::nsDisplayList::AppendNewToTop(mozilla::nsDisplayListBuilder*, nsCanvasFrame*) [nsDisplayList.h:27de51e74c19186044da535125895c96fc3a0f23 : 3072]    
 9  xul.dll!nsCanvasFrame::BuildDisplayList(mozilla::nsDisplayListBuilder*, mozilla::nsDisplayListSet const&) [nsCanvasFrame.cpp:27de51e74c19186044da535125895c96fc3a0f23 : 572 + 0x1f]    
10  xul.dll!nsIFrame::BuildDisplayListForChild(mozilla::nsDisplayListBuilder*, nsIFrame*, mozilla::nsDisplayListSet const&, mozilla::EnumSet<nsIFrame::DisplayChildFlag,unsigned int>) [nsIFrame.cpp:27de51e74c19186044da535125895c96fc3a0f23 : 4314 + 0x1a]
...
```

breakpad-symbols/src/http.rs
```diff
@@ -181,6 +182,90 @@ fn commit_cache_file(mut temp: NamedTempFile, final_path: &Path, url: &Url) -> i
...
+/// This returns a DebugInfoResult with the new debug file and debug identifier
+/// or None.
+async fn lookup_debug_info_by_code_info(
+    symbol_urls: &Vec<Url>,
+    module: &(dyn Module + Sync),
+) -> Option<DebugInfoResult> {
+    let lookup_path = match code_info_breakpad_sym_lookup(module) {
+        Some(value) => value,
+        _ => return None,
+    };
```
gabrielesvelto, line 212: Since you're returning `None` you can replace this line with:
```
let lookup_path = code_info_breakpad_sym_lookup(module)?;
```

breakpad-symbols/src/http.rs
```diff
@@ -181,6 +182,90 @@ fn commit_cache_file(mut temp: NamedTempFile, final_path: &Path, url: &Url) -> i
...
+
+    for base_url in symbol_urls {
+        let url = base_url.join(&lookup_path).ok()?;
```
gabrielesvelto, line 215: If `base_url.join()` here fails we'll return from the function which is probably not what you want. To avoid using `continue` and calls to `unwrap()` in the code past this point I propose doing something like this:
```rust
for base_url in symbol_urls {
  if let Some(result) = individual_lookup_debug_info_by_code_info(lookup_path) {
    return result;
  }
}
```
Then you structure `individual_lookup_debug_info_by_code_info()` to make extensive use of the comma operator to return `None` when it cannot proceed. For example:
```rust
fn individual_lookup_debug_info_by_code_info(lookup_path: &str) -> Option<DebugInfoResult> {
  let url = base_url.join(&lookup_path).ok()?;
  let no_redirects_client = Client::builder().redirect(redirect::Policy::none()).build().ok()?;
  let response = no_redirects_client.get(url.clone()).send().await.ok()?;
  ...
}
```
Basically at every step where you have a function that must succeed, turn the result into an `Option<T>` (if it's not already) and then use the question mark operator to proceed or return `None`. Note how I use the `Result::ok()` function to turn a `Result<T>` into an `Option<T>` where the error part of the latter becomes a `None`. This turns all the control flow into easy-to-read straight line code.

gabrielesvelto (changes requested):
This is looking good, but I think we can greatly simplify the control flow within `lookup_debug_info_by_code_info()`. Sorry for not pointing it out earlier but I'm still not very experienced when it comes to idiomatic Rust.

breakpad-symbols/src/http.rs
willkg: That's sooooo much better! Thank you!

breakpad-symbols/src/http.rs
```diff
@@ -181,6 +182,90 @@ fn commit_cache_file(mut temp: NamedTempFile, final_path: &Path, url: &Url) -> i
...
+) -> Option<DebugInfoResult> {
+    let url = base_url.join(lookup_path).ok()?;
+
+    debug!("Trying code file / code identifier lookup: {}", url);
+
+    // This should not follow redirects--we want the next url if there is one
+    let no_redirects_client = Client::builder()
+        .redirect(redirect::Policy::none())
+        .build()
+        .unwrap();
```
gabrielesvelto, line 206: You can replace the `.unwrap()` here with `.ok()?`.

breakpad-symbols/src/http.rs
```diff
@@ -181,6 +182,90 @@ fn commit_cache_file(mut temp: NamedTempFile, final_path: &Path, url: &Url) -> i
...
+    if let Ok(res) = response {
+        let res_status = res.status();
+        if res_status == reqwest::StatusCode::FOUND
+            || res_status == reqwest::StatusCode::MOVED_PERMANENTLY
+        {
+            let location_header = res.headers().get("Location")?;
+            let mut new_url = location_header.to_str().ok()?;
+            if new_url.starts_with('/') {
+                new_url = new_url.strip_prefix('/').unwrap();
+            }
```
gabrielesvelto, line 218: You can get rid of the `.unwrap()` here by rewriting this like:
```rust
new_url = new_url.strip_prefix('/').unwrap_or(new_url);
```

gabrielesvelto (approved):
Looking good, I just left a couple of comments to get rid of the last two calls to `unwrap()`. This way the result code will never panic.

willkg:
Thank you for helping me land this!

kunwyoo:
Does this patch work for locally cached symbols? i.e. using --symbols-path option with minidump-stackwalk binary

willkg:
@kunwyoo No, it does not affect locally cached symbols since it's in the http symbol supplier and relies on the Mozilla Symbols Server supporting an additional download API mode.

PR #876, opened 2023-09-18
update dependencies with incompatible updates

demoray:


gabrielesvelto:
Sorry for the delay here, any specific reason you want to update to 0.7.1 instead of 0.8.0 or 0.9.0? symbolic currently uses 0.8.0, so if we bump that too we can avoid having a duplicate crate.

demoray:
> Sorry for the delay here, any specific reason you want to update to 0.7.1 instead of 0.8.0 or 0.9.0? symbolic currently uses 0.8.0, so if we bump that too we can avoid having a duplicate crate.

When this pr was created, 0.8 and 0.9 had not been released yet.   At the time, I was looking to reduce duplicate crates.

Moving to an even later version is fine with me.

gabrielesvelto:
I updated to 0.8.0 because it aligns the project with what symbolic uses, and since it's one of the dependencies this reduces duplication. Thanks!

PR #883, opened 2023-10-12
Fix redirect url parsing (#882)

### LINKED ISSUE #882: codeinfo lookup doesn't work with try/ path
If the symbols-url value is `https://symbols.mozilla.org`, then the codeinfo lookup works and the symbol supplier finds the symbol.

If the symbols-url value is `https://symbols.mozilla.org/try`, then the codeinfo lookup works, but something fails and the symbol supplier never finds the symbol.

This fails:

```
minidump-stackwalk \
    --evil-json=./crashdata_mdsw_tmp/v1/raw_crash/20231005/a477945d-9803-4cb9-9d5c-df0da0231005 \
    --symbols-cache=/tmp/symbols/cache \
    --symbols-tmp=/tmp/symbols/tmp \
    --no-color \
    --symbols-url=https://symbols.mozilla.org/try \
    --output-file=a477945d-9803-4cb9-9d5c-df0da0231005.dump.json \
    --log-file=a477945d-9803-4cb9-9d5c-df0da0231005.dump.log \
    --json \
    --verbose=debug \
    ./crashdata_mdsw_tmp/v1/dump/a477945d-9803-4cb9-9d5c-df0da0231005
```

debug output (most of the sym files are cached, so there's no output for those):

```
DEBUG resolving host="symbols.mozilla.org"      
DEBUG connecting to 35.161.63.49:443      
DEBUG connected to 35.161.63.49:443      
DEBUG No cached session for DnsName("symbols.mozilla.org")          
DEBUG Not resuming any session          
DEBUG ALPN protocol is None          
DEBUG Using ciphersuite TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256          
DEBUG ECDHE curve is ECParameters { curve_type: NamedCurve, named_group: secp256r1 }          
DEBUG Server DNS name is DnsName("symbols.mozilla.org")          
DEBUG flushed 116 bytes      
DEBUG parsed 16 headers      
DEBUG incoming body is empty        
DEBUG pooling idle connection for ("https", symbols.mozilla.org)      
DEBUG redirect policy disallowed redirection to 'https://symbols.mozilla.org/try/xul.pdb/445AC5F4EF9816474C4C44205044422E1/xul.sym'      
DEBUG No debug file / debug id found with lookup path xul.dll/6514D0BB72EE000/xul.sym.      
DEBUG Sending warning alert CloseNotify          
```

willkg:
When the symbol supplier does a codeinfo lookup and it's successful, the symbol supplier gets back an HTTP 302 with a url to redirect to in the Location header. The path of that url may have additional parts at the beginning that should be ignored--the debug file and debug id are the third and second from the end.

This fixes the path part parsing so
individual_lookup_debug_info_by_code_info picks up the debug file and debug id correctly.

This fixes the code so it works with any of these symbols-url variations:

* `https://symbols.mozilla.org`
* `https://symbols.mozilla.org/`
* `https://symbols.mozilla.org/try`
* `https://symbols.mozilla.org/try/`

Fixes #882

### DIFF
--- a/breakpad-symbols/src/http.rs
+++ b/breakpad-symbols/src/http.rs
@@ -218,19 +218,13 @@ async fn individual_lookup_debug_info_by_code_info(
             }
 
             // new_url looks like some/path/stuff/xul.pdb/somedebugid/xul.sym and we want the debug
-            // file and debug id portions which are the third from end and second from end
-            let mut parts: Vec<&str> = new_url.split('/').collect();
-            if parts.len() < 3 {
-                return None;
-            }
-            while parts.len() > 3 {
-                parts.remove(0);
-            }
-
-            let debug_file_part = parts[0];
-            let debug_file = String::from(debug_file_part);
-            let debug_identifier_part = parts[1];
+            // file and debug id portions which are at fixed indexes from the end
+            let mut parts = new_url.rsplit('/');
+            let debug_identifier_part = parts.nth(1)?;
             let debug_identifier = DebugId::from_str(debug_identifier_part).ok()?;
+            let debug_file_part = parts.next()?;
+            let debug_file = String::from(debug_file_part);
+
             debug!("Found debug info {} {}", debug_file, debug_identifier);
             return Some(DebugInfoResult {
                 debug_file,
--- a/breakpad-symbols/src/http.rs
+++ b/breakpad-symbols/src/http.rs
@@ -216,10 +216,20 @@ async fn individual_lookup_debug_info_by_code_info(
             if new_url.starts_with('/') {
                 new_url = new_url.strip_prefix('/').unwrap_or(new_url);
             }
-            let mut parts = new_url.split('/');
-            let debug_file_part = parts.next()?;
+
+            // new_url looks like some/path/stuff/xul.pdb/somedebugid/xul.sym and we want the debug
+            // file and debug id portions which are the third from end and second from end
+            let mut parts: Vec<&str> = new_url.split('/').collect();
+            if parts.len() < 3 {
+                return None;
+            }
+            while parts.len() > 3 {
+                parts.remove(0);
+            }
+
+            let debug_file_part = parts[0];
             let debug_file = String::from(debug_file_part);
-            let debug_identifier_part = parts.next()?;
+            let debug_identifier_part = parts[1];
             let debug_identifier = DebugId::from_str(debug_identifier_part).ok()?;
             debug!("Found debug info {} {}", debug_file, debug_identifier);
             return Some(DebugInfoResult {


willkg:
@gabrielesvelto Can you review this please?

breakpad-symbols/src/http.rs
```diff
@@ -216,10 +216,20 @@ async fn individual_lookup_debug_info_by_code_info(
...
+                return None;
+            }
+            while parts.len() > 3 {
+                parts.remove(0);
+            }
+
+            let debug_file_part = parts[0];
             let debug_file = String::from(debug_file_part);
-            let debug_identifier_part = parts.next()?;
+            let debug_identifier_part = parts[1];
```
gabrielesvelto, line 232: Since you're interested in parts of the URL that are at fixed positions from the back you can use a reverse iterator:
```rust
            let mut parts = new_url.rsplit('/');
            let debug_file_part = parts.nth(1)?;
            let debug_file = String::from(debug_file_part);
            let debug_identifier_part = parts.nth(2)?;
```
`.rsplit()` works like `.split().rev()` in that the results are returned in reverse order, and `.nth()` allows you to select a specific element (returning an `Option<T>` as usual).

gabrielesvelto (changes requested):

willkg:
^^^ That reworks the code to use `rsplit`.

breakpad-symbols/src/http.rs
```diff
@@ -216,11 +216,15 @@ async fn individual_lookup_debug_info_by_code_info(
...
                 new_url = new_url.strip_prefix('/').unwrap_or(new_url);
             }
-            let mut parts = new_url.split('/');
+
+            // new_url looks like some/path/stuff/xul.pdb/somedebugid/xul.sym and we want the debug
+            // file and debug id portions which are at fixed indexes from the end
+            let mut parts = new_url.rsplit('/');
+            let debug_identifier_part = parts.nth(1)?;
+            let debug_identifier = DebugId::from_str(debug_identifier_part).ok()?;
             let debug_file_part = parts.next()?;
```
gabrielesvelto, line 225: This won't work as the previous call `parts.nth(1)` doesn't advance the iterator, but you can use `.nth(2)` instead here.

gabrielesvelto (changes requested):

breakpad-symbols/src/http.rs
willkg: I couldn't get it to work, so I looked up `nth` and if I'm reading this correctly, it does advance the iterator:

https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.nth

So then I switched the order of extracting debug_id and debug_file and changed debug_file to `nth(0)`. Then clippy said to change it to `next()`.

gabrielesvelto: Oh, I didn't know it advanced the iterator. I check in devdocs documentation and it didn't mention it, maybe it's out-of-date. Anyway, if clippy says so I won't argue :ok_hand:

gabrielesvelto (approved):

willkg:
Thank you!

willkg:
This was related to issue #870.

PR #884, opened 2023-10-18
chore: run 'cargo dist init' with cargo-dist 0.4.0-prerelease.2

Gankra:
(want to check that some new features we have planned work for y'all)


Gankra:
Nice, looks like minidump-stackwalk is out-of-the-box compatible with musl builds if y'all want them :)

gabrielesvelto:
Thanks :heart:

Issue #324, opened 2021-11-30
Fill in all the ex_list streams


Gankra:
Several official streams all have the same "ex_list" format, and are therefore potentially trivial to parse and provide a minimal implementation of. All you need is to define the struct that each entry is:

https://github.com/luser/rust-minidump/blob/5069ee30611c06ca7672f74cbfa055b089c84880/minidump/src/lib.rs#L326-L354

https://github.com/luser/rust-minidump/blob/5069ee30611c06ca7672f74cbfa055b089c84880/minidump/src/minidump.rs#L1439-L1441

Some quick checking suggests these are:

* [HandleDataStream](https://docs.microsoft.com/en-us/windows/win32/api/minidumpapiset/ns-minidumpapiset-minidump_thread_info_list)
  * **NOTE: this one [actually uses](https://docs.microsoft.com/en-us/windows/win32/api/minidumpapiset/ns-minidumpapiset-minidump_handle_data_stream#remarks) the versioning scheme!** 
* [HandleOperationListStream](https://docs.microsoft.com/en-us/windows/win32/api/minidumpapiset/ns-minidumpapiset-minidump_handle_operation_list)
* [ThreadInfoListStream](https://docs.microsoft.com/en-us/windows/win32/api/minidumpapiset/ns-minidumpapiset-minidump_thread_info_list)

Gankra:
A couple other streams are "nearly" ex_lists:

* [FunctionTableStream](https://docs.microsoft.com/en-us/windows/win32/api/minidumpapiset/ns-minidumpapiset-minidump_function_table_stream)
* TokenStream (has MINIDUMP_TOKEN_INFO_HEADER and MINIDUMP_TOKEN_INFO_LIST structs in minidumpapiset.h)

I think both of these are muddied because they are seemingly 2d arrays instead of 1d arrays (TokenStream isn't clearly documented).

Gankra:
Also HandleOperationListStream is actually kind of cool? It's like MemoryInfoList in that it lets you know if specific Handles (file descriptors) are closed/opened/invalid. Dunno if it's only applicable in really narrow situations though.

```
// Some operating systems can track the last operations
// performed on a handle.  For example, Application Verifier
// can enable this for some versions of Windows.  The
// handle operation list collects handle operations
// known for the dump target.
// Each entry is an AVRF_HANDLE_OPERATION.
```

[...]

PR #885, opened 2023-10-19
Add support for the MINIDUMP_HANDLE_DATA_STREAM stream

gabrielesvelto:
This covers part of issue #324


gabrielesvelto:
Tests are missing, I'll add them next week

gabrielesvelto:
I've manually tested this on Windows-generated minidumps that had the stream populated

gabrielesvelto:
@Gankra I've tested this with actual minidumps and everything seems to be working fine, but it seems like I messed up something in *minidump-synth*. Somehow the test fail with the value of `handle` being returned in the wrong endianness, so instead of `0x000000000000007b` I get `0x7b00000000000000`. I can't figure out what I'm doing wrong though.

minidump-synth/src/lib.rs
```diff
@@ -758,7 +782,8 @@ impl<T: ListItem> ExList<T> {
         let section = Section::with_endian(endian)
             .D32(12)
```
Gankra, line 783:
```suggestion
            .D32(12 + padding as u32)
```

Gankra (approved):

minidump-common/src/format.rs
```diff
@@ -2459,3 +2462,86 @@ bitflags! {
...
+    MiniEventInformation1,
+    MiniSectionInformation1,
+    MiniSemaphoreInformation1,
+    MiniHandleObjectInformationTypeMax,
+}
+
+// OS-specific handle object information. Microsoft headers state that it can
+// change based on the target platform. The object is larger than this structure
+// (as specified by `size_of_info`) and the remaining data is stored after the
+// `size_of_info` field. The format of this information is not specified.
```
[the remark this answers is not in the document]
gabrielesvelto: Fixed and added some extra info to the structure itself

minidump-synth/src/lib.rs
gabrielesvelto: I went with passing the entire header size and then figuring out the padding from there, it's harder to get it wrong then

gabrielesvelto: Thanks a lot for spotting what was wrong with this code

PR #891, opened 2023-11-10
Allow passing the path to a symbol files in --symbols-path

### LINKED ISSUE #890: Documentation of --symbols-path is wrong
Currently, the documentation claims that we can "provide those symbol files directly as symbols-path values (passed positionally, see the cli manual below)", or that they're a "Path to a symbol file.". However, looking at what the code does, it looks like `--symbols-path` is treated as a _directory_ where symbols are looked up by looking at `{path}/{pdbname}.pdb/{id}/{pdbname}.sym`: https://github.com/rust-minidump/rust-minidump/blob/main/breakpad-symbols/src/lib.rs#L452

This is very confusing, and not very practical. It'd be better if `--symbol-path` took the sym file directly, rather than forcing a directory layout.

roblabla:
Fixes #890

### DIFF
--- a/breakpad-symbols/src/lib.rs
+++ b/breakpad-symbols/src/lib.rs
@@ -192,6 +192,8 @@ fn replace_or_add_extension(filename: &str, match_extension: &str, new_extension
 /// A lookup we would like to perform for some file (sym, exe, pdb, dll, ...)
 #[derive(Debug, Clone)]
 pub struct FileLookup {
+    pub debug_id: String,
+    pub debug_file: String,
     pub cache_rel: String,
     pub server_rel: String,
 }
@@ -223,6 +225,8 @@ pub fn breakpad_sym_lookup(module: &(dyn Module + Sync)) -> Option<FileLookup> {
     Some(FileLookup {
         cache_rel: rel_path.clone(),
         server_rel: rel_path,
+        debug_id: debug_id.breakpad().to_string(),
+        debug_file: filename,
     })
 }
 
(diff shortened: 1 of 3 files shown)


roblabla:
I fixed the tests and formats, hopefully CI should pass now 🤞

gabrielesvelto:
@Gankra what do you think about this PR? I've got no particular preference around this but you wrote the lookup logic. AFAIK we did everything this way because that's what breakpad does, but I have no qualms in breaking with tradition if it makes things easier for users.

roblabla:
FWIW, I tried to make the change backwards-compatible. If a directory is given with `--symbols-path`, it will retain the old behavior of looking up the symbols using a symsrv-like lookup. It only uses the new query method if the `symbols-path` points to a file instead of a directory.

So this shouldn't negatively impact any existing users of the tool, I think.

ReactorScram:
This would make it easier for me because I couldn't figure out the directory structure and I just have one symbols file anyway

gabrielesvelto:
Sorry I've been super-slow on this one. I'm doing a few tests to see how this looks like when I mix paths up. In the meantime for those who need accessible symbols, the easiest way is to use `dump_syms --store <path>` when dumping your files. E.g.:
```
dump_syms --store symbols path/to/my/executable.exe
minidump-stackwalk my_minidump.dmp symbols
```
You can dump all your libraries and exes under the same folder this way, and `dump_syms` will automatically create the correct paths.

gabrielesvelto:
I checked this with a mix of folders, file paths and URLs and everything seems to work together just fine, merging.

Issue #893, opened 2023-11-13
32-bit exception information is sometimes sign-extended causing confusion when decoding Windows error codes


gabrielesvelto:
I've noticed that some Windows crashes for 32-bit Firefox & Thunderbird builds where showing odd values for the crash reason, for example [this one](https://crash-stats.mozilla.org/report/index/7b8c95b6-cf69-4f8c-aa8b-6fec10231113). The crash reason is `EXCEPTION_IN_PAGE_ERROR_READ / 0xffffffffc0000006` which is clearly wrong. Sifting through more crashes I figured out that it seems that **windbg.dll** is sign-extending Windows status codes when writing them to the minidump's exception stream `exception_information` array (which is made up of 64-bit values). To address this issue we'll have to either clear or ignore the upper bits of the elements pulled out of the `exception_information` array. I'd like the fix to be as small as possible as I don't want to accidentally throw away meaningful data just because of an implementation bug in some closed-source library.

PR #894, opened 2023-11-13
Interpret Windows error-code correctly even when they're sign-extended.

gabrielesvelto:
When interpreting Windows status code that have been stored as 64-bit values, ignore the upper 32 bits as they're likely to contain bogus data. In particular windbg.dll seems to sign-extend the 32-bit value when storing it as a 64-bit one.

Note that we're not really throwing away data with this patch as the status codes are supposed to be 32 bits, so we shouldn't have relied on the upper 32 bits being cleared in the first place.

This fixes issue #893

### DIFF
--- a/minidump/src/minidump.rs
+++ b/minidump/src/minidump.rs
@@ -4373,7 +4373,8 @@ impl CrashReason {
                 // will be present in the crash thread's instruction field anyway.
                 if record.number_parameters >= 3 {
                     // NOTE: address := info[1];
-                    let nt_status = info[2];
+                    // The status code is 32-bits wide, ignore the upper 32 bits
+                    let nt_status = info[2] & 0xffff_ffff;
                     if let Some(ty) = err::ExceptionCodeWindowsInPageErrorType::from_u64(info[0]) {
                         reason = CrashReason::WindowsInPageError(ty, nt_status);
                     }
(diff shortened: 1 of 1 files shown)


minidump/src/minidump.rs
```diff
@@ -4373,7 +4373,8 @@ impl CrashReason {
                 // will be present in the crash thread's instruction field anyway.
                 if record.number_parameters >= 3 {
                     // NOTE: address := info[1];
-                    let nt_status = info[2];
+                    // The status code is 32-bits wide, ignore the upper 32 bits
+                    let nt_status = info[2] & 0xffffffff;
```
Gankra, line 4377:
```suggestion
                    let nt_status = info[2] & 0xffff_ffff;
```

minidump/src/minidump.rs
```diff
@@ -4384,7 +4385,8 @@ impl CrashReason {
                 // invocations and the fast-fail code is stored in
                 // exception_information[0].
                 if record.number_parameters >= 1 {
-                    let fast_fail = info[0];
+                    // The status code is 32-bits wide, ignore the upper 32 bits
+                    let fast_fail = info[0] & 0xffffffff;
```
Gankra, line 4389:
```suggestion
                    let fast_fail = info[0] & 0xffff_ffff;
```

Gankra (approved):
oh this is evil, good catch

Issue #895, opened 2023-11-14
Update the macOS error codes


gabrielesvelto:
Several new error codes were introduced in recent versions of macOS. Additionally, Apple has moved its repositories on GitHub, so we should update the links to reflect where the new headers are, to help people figure out where those values came from.

PR #901, opened 2023-11-14
Add support for additional macOS guard exception types and flavors

gabrielesvelto:
This fixues issue #895

### DIFF
--- a/minidump-common/src/errors/macos.rs
+++ b/minidump-common/src/errors/macos.rs
@@ -330,7 +330,7 @@ pub enum ExceptionCodeMacResourceThreadsFlavor {
 ///
 /// See the [osfmk/kern/exc_guard.h][header] header in Apple's kernel sources
 ///
-/// [header]: https://github.com/apple/darwin-xnu/blob/main/osfmk/kern/exc_guard.h
+/// [header]: https://github.com/apple-oss-distributions/xnu/blob/main/osfmk/kern/exc_guard.h
 #[derive(Copy, Clone, PartialEq, Eq, Debug, FromPrimitive)]
 pub enum ExceptionCodeMacGuardType {
     GUARD_TYPE_NONE = 0,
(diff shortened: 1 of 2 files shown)


minidump-common/src/errors/macos.rs
```diff
@@ -339,18 +339,21 @@ pub enum ExceptionCodeMacGuardType {
...
 #[derive(Copy, Clone, PartialEq, Eq, Debug, FromPrimitive)]
 pub enum ExceptionCodeMacGuardMachPortFlavor {
-    GUARD_EXC_DESTROY = 0x00000001,
-    GUARD_EXC_MOD_REFS = 0x00000002,
-    GUARD_EXC_SET_CONTEXT = 0x00000004,
+    GUARD_EXC_DESTROY = 1,
+    GUARD_EXC_MOD_REFS = 2,
+    GUARD_EXC_INVALID_OPTIONS = 3,
+    GUARD_EXC_SET_CONTEXT = 4,
+    GUARD_EXC_THREAD_SET_STATE = 5,
```
Gankra, line 356: any reason for the change in notation?

minidump/src/minidump.rs
```diff
@@ -4726,11 +4726,20 @@ impl fmt::Display for CrashReason {
                     if let Some(mach_port_flavor) =
                         err::ExceptionCodeMacGuardMachPortFlavor::from_u64(flavor)
                     {
+                        // TODO: GUARD_EXC_STRICT_REPLY, GUARD_EXC_MOD_REFS and GUARD_EXC_IMMOVABLE have additional flags defined here:
```
Gankra, line 4729: I forget if we ever properly enforced this style but imo ideally "TODO" is "blocks merging" and "FIXME" is "one of these days..."

Gankra (approved):

minidump-common/src/errors/macos.rs
gabrielesvelto: That's how they're written in Apple's headers now, and I like to stick to the same notation so that people grep'ing stuff around can find their way in the original sources.

minidump/src/minidump.rs
gabrielesvelto: Gotcha, I'm changing it

PR #932, opened 2024-01-16
Update memmap2 to 0.9.3

glandium:


glandium:
I guess this needs a version bump...

gabrielesvelto:
> I guess this needs a version bump...

There a few fixes that have landed in the past months, I can cut a new release.

glandium:
To be clear, I meant a semver bump.

I'm thinking Mmap should probably be re-exported from minidump, or Minidump<'a, Mmap> aliased so that dependencies of the crate don't need to have a dependency on memmap2 to use it.

gabrielesvelto:
OK, I can do a major version bump after we're done with the changes (there's more fixes in the pipeline for Socorro so this week is a good time).

PR #934, opened 2024-01-17
Avoid requiring a dependency on memmap2 to use Minidump<'_, Mmap>

glandium:
The main problem is that the dependency on memmap2 then needs to be the exact same as the one the minidump crate has, which doesn't necessarily line up.

Another option would be to reexport Mmap.

### DIFF
--- a/minidump/src/minidump.rs
+++ b/minidump/src/minidump.rs
@@ -5182,11 +5182,14 @@ impl MinidumpCrashpadInfo {
     }
 }
 
+/// An index into the contents of a memory-mapped minidump.
+pub type MmapMinidump<'a> = Minidump<'a, Mmap>;
+
 impl<'a> Minidump<'a, Mmap> {
     /// Read a `Minidump` from a `Path` to a file on disk.
     ///
     /// See [the type definition](Minidump.html) for an example.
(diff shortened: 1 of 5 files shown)


luser:
> Another option would be to reexport Mmap.

Having seen that pattern in other crates, that way gets messy quickly.

minidump/src/minidump.rs
```diff
@@ -5182,6 +5182,8 @@ impl MinidumpCrashpadInfo {
     }
 }
 
+pub type MmapMinidump<'a> = Minidump<'a, Mmap>;
```
luser, line 5185: Can you put a doc comment on here, or on `Minidump::read_path` (or both) that tells users how to use this?

luser (approved):

minidump/src/minidump.rs
glandium: Interestingly, the doctest uses Minidump::read_path directly, and relies on type inference, not requiring to use Mmap at all. Only minidump-writer uses the type because it stores it in a struct.

luser (approved):
Thanks!

glandium:
Erf, github was not showing me the last approval when I requested a new review...

PR #938, opened 2024-01-17
Remove the dependency on dump_syms

gabrielesvelto:


gabrielesvelto:
@Gankra we support unwinding with native debug information and removing dump_syms untangles a lot of dependencies that make updating and vendoring the crate a lot harder. If you haven't got anything against it I'll just remove the support.

akaspeh1:
Hi there,

I've been using a symbol server compatible with Visual Studio to serve DLLs, PDBs, and EXEs via HTTP. Recently, I encountered an issue where the minidump debugger, which was using V0.19, worked fine, but the rust-minidump Version 0.19.1 (2023-11-02) didn't function as expected.

After some investigation, I realized that I needed to enable the "http" and "dump_sys" functions to ensure compatibility. I proceeded by compiling Version 0.19.1 (2023-11-02) with these functions enabled, allowing the minidump reader to retrieve debug symbols from my symbol server using the same paths as Visual Studio.

However, upon removing the "dump_sys" function in versions beyond 0.20, I've encountered a setback. 
I'm no longer able to effectively utilize my symbol server for integrating with Visual Studio and rust-minidump to gather libraries.

I'm currently seeking solutions to resolve this matter. Any insights or suggestions would be greatly appreciated.

gabrielesvelto:
We now have functionality to use native debug information directly, so it's just a matter of teaching the `HttpSymbolSupplier` to return those instead of Breakpad .sym files when they are available. Can you open an issue with this so we can track it?

PR #940, opened 2024-01-20
Remove `Unwind` trait and further cleanups

Swatinem:
I remembered that the `Unwind` trait is purely internal to the `minidump-unwind` crate.
It exists only because its implemented on foreign types from the `minidump-common` crate.

I started out just removing the `#[async_trait]` proc macro, as AFIT is now stable in latest Rust.
But then I just removed that trait completely, as it is pretty pointless to have it as a trait in the first place.

I then continued down a rabbit hole, cleaning up the wall of arguments passed through to all the unwinder code and replaced that with a struct holding all those arguments, as that results in much cleaner code.

You can review this as individual commits, as there is a progression to the cleanups.


gabrielesvelto:
This is a very nice cleanup. I'm busy with something else so I can't do an in-depth review ATM but I'll try to get to it ASAP if no one else did in the meantime.

PR #941, opened 2024-01-24
Add initial riscv64 support

makotokato:
The defines for riscv64 are from Google's breakpad. This is tested on Ubuntu 23.04/riscv64 using generated minidump by Google's breakpad.

### DIFF
--- a/minidump-unwind/src/riscv64.rs
+++ b/minidump-unwind/src/riscv64.rs
@@ -1,0 +1,313 @@
+// Copyright 2015 Ted Mielczarek. See the COPYRIGHT
+// file at the top-level directory of this distribution.
+
+use super::impl_prelude::*;
+use crate::{SymbolProvider, SystemInfo};
+use minidump::{
+    CpuContext, MinidumpContext, MinidumpContextValidity, MinidumpModuleList, MinidumpRawContext,
+    UnifiedMemory,
+};
+use std::collections::HashSet;
+use tracing::trace;
+
+type RiscvContext = minidump::format::CONTEXT_RISCV64;
+type Pointer = <RiscvContext as CpuContext>::Register;
+
+const POINTER_WIDTH: Pointer = std::mem::size_of::<Pointer>() as Pointer;
+const FRAME_POINTER: &str = "s0";
+const LINK_REGISTER: &str = "ra";
+const STACK_POINTER: &str = "sp";
+const PROGRAM_COUNTER: &str = "pc";
+const CALLEE_SAVED_REGS: &[&str] = &[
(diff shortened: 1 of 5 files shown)


minidump-unwind/src/riscv64.rs
```diff
@@ -0,0 +1,313 @@
...
+        // enforce progress and avoid infinite loops.
+
+        let sp = frame.context.get_stack_pointer();
+        let last_sp = self.get_register_always("sp");
+        if sp <= last_sp {
+            // Arm leaf functions may not actually touch the stack (thanks
+            // to the link register allowing you to "push" the return address
+            // to a register), so we need to permit the stack pointer to not
+            // change for the first frame of the unwind. After that we need
+            // more strict validation to avoid infinite loops.
```
gabrielesvelto, line 293: Likewise this looks like a copy-paste error, does the logic here hold true for RISC-V too? I'm not sufficiently familiar with the ABI to be able to tell.

minidump-unwind/src/riscv64.rs
```diff
@@ -0,0 +1,313 @@
...
+
+        // Ok, the frame now seems well and truly valid, do final cleanup.
+
+        // A caller's ip is the return address, which is the instruction
+        // *after* the CALL that caused us to arrive at the callee. Set
+        // the value to 4 less than that, so it points to the CALL instruction
+        // (arm64 instructions are all 4 bytes wide). This is important because
+        // we use this value to lookup the CFI we need to unwind the next frame.
+        let ip = frame.context.get_instruction_pointer();
+        frame.instruction = ip - 4;
```
gabrielesvelto, line 309: This is certainly wrong for RISC-V as the “C” extension for compressed instructions allows for 2-bytes wide instructions and all jumps have an address that is a multiple of 2 bytes.

gabrielesvelto (changes requested):

minidump-unwind/src/riscv64.rs
```diff
@@ -0,0 +1,313 @@
+// Copyright 2015 Ted Mielczarek. See the COPYRIGHT
+// file at the top-level directory of this distribution.
```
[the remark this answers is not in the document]
luser: There's a LICENSE file in the root of the repository, I think that should cover it? I would recommend dropping the license comments from individual files. I probably picked that habit up from Breakpad.

hack3ric:
Hi, is this PR getting worked on? This PR seems mergeable without the minidump-unwind part, so I am planning to submit another PR and making at least minidump and minidump-common work on riscv64.

PR #944, opened 2024-01-31
Validate `stack_memory` once in `walk_stack`

Swatinem:
Doing a little bit of profiling, I found that `UnifiedMemory::get_memory_at_address` is quite hot. This lead me to a comment that I added previously about doing overflow checks on every single access. Instead of doing that on every access, rather do it once ahead of using it within `walk_stack`.


gabrielesvelto:
This is a really nice improvement, thanks!

PR #947, opened 2024-02-06
Remove `chrono`

Jake-Shadle:
`procfs-core` uses `chrono` as a default feature, this is unneeded, I only noticed it because `minidump-writer` already removed the dependency, but it was still present in the lockfile due to `minidump`.

Also fixes the CI badge that wasn't fixed when `master` was renamed to `main`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -27,21 +27,6 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


gabrielesvelto (approved):
Very nice, thanks!

PR #954, opened 2024-02-12
Use wholesym and framehop in minidump-unwind to greatly improve the performance of debuginfo-based unwinding.

afranchuk:
These changes to `minidump-unwind` allow for much faster debuginfo unwinding and symbol retrieval.

As before, support for debuginfo in `minidump-unwind` is optional at compile-time. Now we also separately support enabling _only_ debuginfo unwinding (without symbol support) at compile-time and runtime, since symbol support may not be needed for all purposes (though it is still quite fast, using `wholesym` when it is needed).

All debuginfo code has been moved to its own file.


minidump-unwind/src/symbols/debuginfo.rs
```diff
@@ -0,0 +1,488 @@
...
+            tracing::error!("failed to map {} for debug info: {e}", path.display());
+            return None;
+        }
+    };
+
+    let objfile = match object::read::File::parse(
+        // # Safety
+        // We broaden the lifetime to static, but ensure that the Mmap which provides the data
+        // outlives all references.
+        unsafe { std::mem::transmute::<_, &'static [u8]>(mapped.as_ref()) },
```
gabrielesvelto, line 361: I'm not super-fond of this but I also don't see a better way to do it here given we end up with two references

gabrielesvelto (approved):

gabrielesvelto:
@afranchuk I did some changes to the main branch, can you rebase this before we merge it please?

PR #955, opened 2024-02-15
Bug 1847098 - Parse version number for Elf files

lissyx:


minidump/src/minidump.rs
```diff
@@ -1136,14 +1136,26 @@ impl Module for MinidumpModule {
...
+                let ver = format!(
+                    "{}.{}.{}.{}",
+                    self.raw.version_info.file_version_hi >> 16,
+                    self.raw.version_info.file_version_hi & 0xffff,
+                    self.raw.version_info.file_version_lo >> 16,
+                    self.raw.version_info.file_version_lo & 0xffff
+                );
+                Some(Cow::Owned(ver))
+            } else {
+                // Assume Elf
```
afranchuk, line 1149: I'm not a fan of this assumption (e.g., where would mingw binaries land here?) but if all we can do is check `self.os`, it might be the best we can do. But wouldn't `product_version` also be potentially useful on windows? In which case maybe it'd be better if either this string didn't look like `A.B.C.D` _or_ we have a different function to get the product version (though I realize that under the ELF assumption we are interpreting `file_version` differently, which is annoying).

lissyx: i had a first version (but the information version was stored differently) where I was doing more precise checking, and I was unsure this was really needed. Mostly inspired from https://github.com/rust-minidump/rust-minidump/blob/a66b5c5025f79db47f6c90184e1e9121e0b76f76/minidump/src/minidump.rs#L1123-L1131 `debug_file` function. Do you think it would be a better fit?

I can't speak for `product_version` on Windows though, existing code does not take care of it?

lissyx: > (though I realize that under the ELF assumption we are interpreting file_version differently, which is annoying).

We don't really have a choice here, cf https://bugzilla.mozilla.org/show_bug.cgi?id=1847098#c18

afranchuk: I think it's probably okay as-is. If we encounter some issue we'll know where to look!

lissyx: Fine, I was about to fix it :p

PR #961, opened 2024-02-27
Add fuzz projects to dependabot.

afranchuk:
The fuzz projects need to be updated alongside the main project, otherwise build-fuzz is more likely to fail from time to time.


afranchuk:
Note that I'd like to use yaml references rather than duplicating the `updates` items, but it's not clear from dependabot docs whether an additional unrecognized key (which is needed to set the reference) will cause failure or not.

PR #995, opened 2024-05-07
Update framehop and change `object` to have minimal features.

afranchuk:
Consumers can add more features to change object reading if desired.

The way the crate is written, at least one of `coff`/`elf`/`macho`/`pe`/`wasm`/`xcoff` must be enabled for `object::read::File` to be available. While we could limit the features even further and make consumers choose which formats to support, this instead assumes that they want to be able to read any of the formats since we need to choose at least one anyway.


gabrielesvelto (approved):
Not a fan of the duplicated dependencies but we can address that before we cut a new release by updating the other crates too.

afranchuk:
> Not a fan of the duplicated dependencies but we can address that before we cut a new release by updating the other crates too.

Likewise, however it's only if you enable the debuginfo with symbols (though that is default).

PR #1001, opened 2024-05-21
Add reporting of memory map count

### LINKED ISSUE #1000: Report the number of memory mappings on Linux to help diagnose limit crashes.
On Linux, there is a limit of ~65536 memory mappings that can be made using `mmap()`. After that, future attempts may return an error code. ([Here is an example crash report](https://crash-stats.mozilla.org/report/index/995b0816-afd2-4926-ae93-7c0600240430)).

Since we [already have the list of maps](https://github.com/rust-minidump/rust-minidump/blob/7080a729b41549b729050a9615f8c0026e1da53b/minidump/src/minidump.rs#L179) in the Minidump, let's print this information from Minidump-Analyzer and display it on Socorro. That way, an engineer can quickly look at the number and notice that it's rather close to the system limit, and that something might be amok.

BrianShTsoi:
Resolves  #1000
Looking for lots of feedback, I'm definitely unsure about a lot of the design decisions I should make

1. I noticed that everything under `crash info` comes from `exception_info`. Should `memory_map_count` also be inside `exception_info`?
2. `memory_map_count` is derived from `memory_info`. Is it fair to assume that it is always valid to use `memory_info`, or are there some situations that I overlooked that might lead to non-sense values (eg. `memory_info.iter().count()` just returns 0 because both iters inside `memory_info` is empty)?
3. Is it better to get `memory_map_count` from `memory_info.iter().count()`, and avoid implementing functions for `MinidumpLinuxMapInfo` and `MinidumpMemoryInfo`, or is it better to directly get the vector lengths from these two structs and avoid the iter?
4. Not sure if I positioned my code correctly

According to CONTRIBUTING.md, I still need to
1. Update JSON Schema
2. I think it is fine to leave this under feature stable-basic?
3. Ensure my feature is tested [in test-minidump-stackwalk](https://github.com/rust-minidump/rust-minidump/blob/master/minidump-stackwalk/tests/test-minidump-stackwalk.rs) (Not entirely sure how, I'll try to read through and understand the test code first)


BrianShTsoi:
Unable to add reviewers for some reason so manually pinging reviewers for now @marti4d @gabrielesvelto

gabrielesvelto:
> Unable to add reviewers for some reason so manually pinging reviewers for now @marti4d @gabrielesvelto

I think I know what the problem is, I'll look into it so you can ask for reviews.

gabrielesvelto:
>     1. I noticed that everything under `crash info` comes from `exception_info`. Should `memory_map_count` also be inside `exception_info`?

No, `exception_info` contains the data of the exception that triggered the crash, like the faulting address and the reason. The memory map count is part of the state of the system so it doesn't belong there. I'd also move it out of `crash_info` and into the top-level of the schema. Also it should be flagged as optional, as we won't always have it.

>     2. `memory_map_count` is derived from `memory_info`. Is it fair to assume that it is always valid to use `memory_info`, or are there some situations that I overlooked that might lead to non-sense values (eg. `memory_info.iter().count()` just returns 0 because both iters inside `memory_info` is empty)?

`memory_info` abstracts over the type of memory information we have but this particular measurement will be Linux-specific. I think that the most correct way to implement it is by counting mappings in `MinidumpLinuxMapInfo`. This will naturally restrict it to Linux minidumps. Still, it's worth noting that we might be missing that particular stream, so even on Linux the field might be empty. One of the important things to keep in mind when working with minidumps is that they might be missing bits and pieces, because they were assembled at a stage when the process was already compromised.

> According to CONTRIBUTING.md, I still need to
> 
>     1. Update JSON Schema

Yes, this is important, as we use this to communicate changes to the server-side machinery that processes crashes.

>     2. I think it is fine to leave this under feature stable-basic?

Yes.

>     3. Ensure my feature is tested [in test-minidump-stackwalk](https://github.com/rust-minidump/rust-minidump/blob/master/minidump-stackwalk/tests/test-minidump-stackwalk.rs) (Not entirely sure how, I'll try to read through and understand the test code first)

Generally speaking we process minidumps in the tests. You'll find a pre-generated Linux minidump under `testdata/linux-mini.dmp`, you can use it to test the feature (and naturally, you can also process a non-Linux minidump to make sure that we don't get a count there).

gabrielesvelto:
There's another thing that's worth mentioning: given this field doesn't mean anything on Windows and macOS it shouldn't be printed out at all in the human-readable output of those operating systems (and should not be present, or be null in the JSON output).

BrianShTsoi:
I see. Thank you for the feedback! I'll review them and change my code.

BrianShTsoi:
Also, I don't think @marti4d is on the list of possible reviewers?

luser:
> Ensure my feature is tested [in test-minidump-stackwalk](https://github.com/rust-minidump/rust-minidump/blob/master/minidump-stackwalk/tests/test-minidump-stackwalk.rs) (Not entirely sure how, I'll try to read through and understand the test code first)



> Resolves #1000 Looking for lots of feedback, I'm definitely unsure about a lot of the design decisions I should make
> 
>     1. I noticed that everything under `crash info` comes from `exception_info`. Should `memory_map_count` also be inside `exception_info`?
> 
>     2. `memory_map_count` is derived from `memory_info`. Is it fair to assume that it is always valid to use `memory_info`, or are there some situations that I overlooked that might lead to non-sense values (eg. `memory_info.iter().count()` just returns 0 because both iters inside `memory_info` is empty)?
> 
>     3. Is it better to get `memory_map_count` from `memory_info.iter().count()`, and avoid implementing functions for `MinidumpLinuxMapInfo` and `MinidumpMemoryInfo`, or is it better to directly get the vector lengths from these two structs and avoid the iter?
> 
>     4. Not sure if I positioned my code correctly
> 
> 
> According to CONTRIBUTING.md, I still need to
> 
>     1. Update JSON Schema
> 
>     2. I think it is fine to leave this under feature stable-basic?
> 
>     3. Ensure my feature is tested [in test-minidump-stackwalk](https://github.com/rust-minidump/rust-minidump/blob/master/minidump-stackwalk/tests/test-minidump-stackwalk.rs) (Not entirely sure how, I'll try to read through and understand the test code first)

n.b. you can also write tests for lower layers like `minidump-processor` and use the `SynthMinidump` types to create synthetic test minidumps. See, e.g., `test_linux_proc_limits`:
https://github.com/rust-minidump/rust-minidump/blob/7080a729b41549b729050a9615f8c0026e1da53b/minidump-processor/tests/test_processor.rs#L257-L308

BrianShTsoi:
Just pushed the new code to address all the feedback. `memory_map_info` is now at the top level. After implementing a method for `MinidumpLinuxMapInfo`, `memory_map_info` is now only outputed (or not null in json) if `MinidumpLinuxMapInfo` is used. Tests specifically for Linux are added in test-minidump-stackwalk (though the tests still need some fixing).

THere are still some problems.
1. Given 
> Still, it's worth noting that we might be missing that particular stream, so even on Linux the field might be empty.
> 
there is the possibility that `MinidumpLinuxMapInfo` does not exist even if on Linux, and my current implementation in human output will just hide the field. Is this an acceptable handling of the situation, or is it better to explicitly output the fact that the `memory_map_count` info is not available (ie. outputing "memory_map_count: not available" if the system is Linux, while hiding the field entirely for Windows/MacOS)

2.  My current implementation works properly on `testdata/linux-mini.dmp`. However it does not output the field when I run it on some Linux dmp files from the crash reporter such as the example given in the original issue [Eg1](https://crash-stats.mozilla.org/report/index/995b0816-afd2-4926-ae93-7c0600240430#tab-details) and some others [Eg2](https://crash-stats.mozilla.org/report/index/1c6f8d91-d906-4b5a-9846-7444d0240524#tab-details), [Eg3](https://crash-stats.mozilla.org/report/index/a866a1ab-ce10-4989-82d3-4cadd0240524#tab-rawdump). After some investigation, I believe (despite what the comments documented), these dmp files produced valid streams for both `MinidumpMemoryInfoList` and `MinidumpLinuxMapInfo`, and because of the handling of this supposedly rare case in [`UnifiedMemoryInfoList`'s constructor](https://github.com/rust-minidump/rust-minidump/blob/7080a729b41549b729050a9615f8c0026e1da53b/minidump/src/minidump.rs#L2700), only `MinidumpMemoryInfoList` is used. If I change the constructor to choose `MinidumpLinuxMapinfo` over `MinidumpMemoryInfoList`, the current implementation works properly outputs the `memory_map_count` (and gives a reasonable number that's slightly above 65530, as all of them crashed due to `mmap` failures)

I'm not sure what the best way is to handle this situation (I considered detecting whether the system is Linux through some other means, but I'm not sure about this) 

3. The `linux_human` test that I added often fails because the lines generated under `Unimplemented streams encountered` have a different order each time. I presume it doesn't run deterministically because that part is processed asynchronously. Should I not use snap for this kind of tests, or should I find/generate some linux dmp files that will be ran deterministically, or should I do something else?

I also have some other questions, that are not directly related to the code details, but rather just for my high level understanding. 
1. When we say the `memory_map_count` field is only meaningful for linux, is it that the number really has no meaning for other systems, say Windows, or is it that it still represents the number of memory regions the process has, it's just that we only care about it when it is a linux machine (because only linux has the `max_map_count` limit)?

2. While I'm fixing my code I found [the `print` method of `MinidumpLinuxMapInfo`](https://github.com/rust-minidump/rust-minidump/blob/7080a729b41549b729050a9615f8c0026e1da53b/minidump/src/minidump.rs#L2616), and its "parent" in [`UnifiedMemoryInfoList`](https://github.com/rust-minidump/rust-minidump/blob/7080a729b41549b729050a9615f8c0026e1da53b/minidump/src/minidump.rs#L2769),  which I believe was never called? The counter part in [`MinidumpMemoryInfoList`](https://github.com/rust-minidump/rust-minidump/blob/7080a729b41549b729050a9615f8c0026e1da53b/minidump/src/minidump.rs#L2460) is used once when outputing with the `--dump` option. I'm just wondering is there a reason why for this arrangement?

4. I'm also wondering what is the difference between [`UnifiedMemoryList`](https://github.com/rust-minidump/rust-minidump/blob/7080a729b41549b729050a9615f8c0026e1da53b/minidump/src/minidump.rs#L600) and [`UnifiedMemoryInfoList`](https://github.com/rust-minidump/rust-minidump/blob/7080a729b41549b729050a9615f8c0026e1da53b/minidump/src/minidump.rs#L166)?

Apologies for this huge "dump" (pun intended) of text😅. Thank you to everyone who is giving me feedback

gabrielesvelto:
> there is the possibility that `MinidumpLinuxMapInfo` does not exist even if on Linux, and my current implementation in human output will just hide the field. Is this an acceptable handling of the situation, or is it better to explicitly output the fact that the `memory_map_count` info is not available (ie. outputing "memory_map_count: not available" if the system is Linux, while hiding the field entirely for Windows/MacOS)

Having "not available" in the human readable output would be a nice touch, for the JSON output you can just ignore the field.

>     2. My current implementation works properly on `testdata/linux-mini.dmp`. However it does not output the field when I run it on some Linux dmp files from the crash reporter such as the example given in the original issue [Eg1](https://crash-stats.mozilla.org/report/index/995b0816-afd2-4926-ae93-7c0600240430#tab-details) and some others [Eg2](https://crash-stats.mozilla.org/report/index/1c6f8d91-d906-4b5a-9846-7444d0240524#tab-details), [Eg3](https://crash-stats.mozilla.org/report/index/a866a1ab-ce10-4989-82d3-4cadd0240524#tab-rawdump). After some investigation, I believe (despite what the comments documented), these dmp files produced valid streams for both `MinidumpMemoryInfoList` and `MinidumpLinuxMapInfo`, and because of the handling of this supposedly rare case in [`UnifiedMemoryInfoList`'s constructor](https://github.com/rust-minidump/rust-minidump/blob/7080a729b41549b729050a9615f8c0026e1da53b/minidump/src/minidump.rs#L2700), only `MinidumpMemoryInfoList` is used. The evidence that leads me to think this is if I change the constructor to choose `MinidumpLinuxMapinfo` over `MinidumpMemoryInfoList`, the current implementation works properly outputs the `memory_map_count` (and gives a reasonable number that's slightly above 65530, as all of them crashed due to `mmap` failures)

That sounds reasonable. All those minidumps have both streams populated: `LinuxMaps` and `MemoryInfoList`. However if I dump them the number of entries in `MemoryInfoList` looks reasonable, could you check it? Maybe there's something else going on in the logic because it should work. You can dump the raw stream using `minidump-stackwalk --dump <file>`.

> I'm not sure what the best way is to handle this situation (I considered detecting whether the system is Linux through some other means, but I'm not sure about this)

It's better to base ourselves on the streams, the minidump might be corrupt in bizarre ways and that'll make the logic more complex.

>     3. The `linux_human` test that I added often fails because the lines generated under `Unimplemented streams encountered` have a different order each time. I presume it doesn't run deterministically because that part is processed asynchronously. Should I not use snap for this kind of tests, or should I find/generate some linux dmp files that will be ran deterministically, or should I do something else?

File an issue for that and we'll have to fix it so they're printed out in a predictable order.

> I also have some other questions, that are not directly related to the code details, but rather just for my high level understanding.
> 
>     1. When we say the `memory_map_count` field is only meaningful for linux, is it that the number really has no meaning for other systems, say Windows, or is it that it still represents the number of memory regions the process has, it's just that we only care about it when it is a linux machine (because only linux has the `max_map_count` limit)?

On other platforms there is no limit I know of on the number of mappings, so it's just a number which doesn't really give any information to the user.

>     2. While I'm fixing my code I found [the `print` method of `MinidumpLinuxMapInfo`](https://github.com/rust-minidump/rust-minidump/blob/7080a729b41549b729050a9615f8c0026e1da53b/minidump/src/minidump.rs#L2616), and its "parent" in [`UnifiedMemoryInfoList`](https://github.com/rust-minidump/rust-minidump/blob/7080a729b41549b729050a9615f8c0026e1da53b/minidump/src/minidump.rs#L2769),  which I believe was never called? The counter part in [`MinidumpMemoryInfoList`](https://github.com/rust-minidump/rust-minidump/blob/7080a729b41549b729050a9615f8c0026e1da53b/minidump/src/minidump.rs#L2460) is used once when outputing with the `--dump` option. I'm just wondering is there a reason why for this arrangement?

I honestly don't remember, I'll check later.

>     3. I'm also wondering what is the difference between [`UnifiedMemoryList`](https://github.com/rust-minidump/rust-minidump/blob/7080a729b41549b729050a9615f8c0026e1da53b/minidump/src/minidump.rs#L600) and [`UnifiedMemoryInfoList`](https://github.com/rust-minidump/rust-minidump/blob/7080a729b41549b729050a9615f8c0026e1da53b/minidump/src/minidump.rs#L166)?

The former is for the memory regions contained in the minidump and the latter for the memory regions contained in the process. The former is obviously a subset of the latter, and contains only the bits that we actually saved in the minidump with the memory that they contain.

gabrielesvelto:
The `MinidumpLinuxMaps::print()` is added for completeness. Since the crate can be used as a library and someone might want to print that stuff (even though we don't use it in minidump-stackwalk).

BrianShTsoi:
Thank you for the feedback! I go fix my code right away. 
> That sounds reasonable. All those minidumps have both streams populated: LinuxMaps and MemoryInfoList. However if I dump them the number of entries in MemoryInfoList looks reasonable, could you check it? Maybe there's something else going on in the logic because it should work. You can dump the raw stream using minidump-stackwalk --dump <file>.

Yes the number of entries in `MemoryInfoList` outputed from the dump looks reasonable. Originally I was fixated on only extracting the `memory_map_count` value from `memory_info.maps()` when the `MinidumpInfo` gets converted to `ProcessState` and so I got stuck as `memory_info.maps()` gives None if `MemoryInfoList` exists. I didn't realize that I can just add a field in `MinidumpInfo` and extract the value directly after the stream is read. I'll change it to that.

afranchuk (approved):
This looks great!

I agree with @gabrielesvelto, you should create a separate issue to make the printing of "unimplemented streams encountered" lines deterministic.

gabrielesvelto:
This should be good to land, @BrianShTsoi can you rebase this on top of the current main branch please?

gabrielesvelto (approved):
Thank you @BrianShTsoi, I'm merging this

PR #1003, opened 2024-05-28
Update the locked `backtrace` (used by tokio) to remove duplicate `object` crates.

afranchuk:


afranchuk:
I'm not sure the CI failures are related to this change?

gabrielesvelto:
As @BrianShTsoi noticed in this [comment](https://github.com/rust-minidump/rust-minidump/pull/1001#issuecomment-2135575167) it's probably because of the `time` crate, could you update that one too since you're at it?

afranchuk:
Yeah I figured as much, just wasn't sure if we had it already updated in some other PR. It's appropriate to include that here.

gabrielesvelto:
You'll need to update the `time` crate also in `minidump/fuzz/Cargo.lock`, `minidump-processor/fuzz/Cargo.lock` and `minidump-unwind/fuzz/Cargo.lock`.

afranchuk:
Hmm, I didn't check them manually but assumed they had updated there too.

gabrielesvelto (approved):

PR #1006, opened 2024-05-31
Sort unknown and unimplemented streams for deterministic output

### LINKED ISSUE #1004: Lines under `Unimplemented streams` of `linux_human` test not outputted deterministically
As mentioned in https://github.com/rust-minidump/rust-minidump/pull/1001#issuecomment-2129874085, 
> The linux_human test that I added often fails because the lines generated under Unimplemented streams encountered have a different order each time

BrianShTsoi:
Resolves #1004 

Let me know if there's a better way to sort them.

### DIFF
--- a/minidump-processor/src/processor.rs
+++ b/minidump-processor/src/processor.rs
@@ -879,14 +879,8 @@ impl<'a> MinidumpInfo<'a> {
             .collect();
 
         // Collect up info on unimplemented/unknown modules
-        let mut unknown_streams = dump
-            .unknown_streams()
-            .collect::<Vec<MinidumpUnknownStream>>();
-        unknown_streams.sort_by(|a, b| (a.stream_type).cmp(&(b.stream_type)));
-        let mut unimplemented_streams = dump
-            .unimplemented_streams()
-            .collect::<Vec<MinidumpUnimplementedStream>>();
-        unimplemented_streams.sort_by(|a, b| (a.stream_type as u32).cmp(&(b.stream_type as u32)));
+        let unknown_streams = dump.unknown_streams().collect();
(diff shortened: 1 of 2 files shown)


afranchuk (approved):
This will fix the problem (and the ordering you chose is sensible), however did you consider changing the underlying data structure used for `Minidump::streams` to have it already sorted? This would fix both methods.

BrianShTsoi:
Ah I did not thought of that. Let me look into it.

BrianShTsoi:
So I looked into it and I can change `Minidump::streams` to be a `BTreeMap` to have it always sorted. 
I am just wondering what are the implications on performance for this change? Asking because I tried to look into the code, and from what I understand most of the use of `Minidump::streams` (mostly with `.get()`) don't really need it to be sorted, and all the situations where we sort the streams are for outputs. There are also [precedents](https://github.com/rust-minidump/rust-minidump/blob/c01c10a7ca51ab44712cbaf83ff83accef6aae27/minidump/src/minidump.rs#L5688) of sorting the streams right before outputting. 

(Tbh I also just don't have a clear idea of how much we should care about performance in rust-minidump in general)

BrianShTsoi:
So `Minidump::unknown_streams()` and `Minidump::unimplemented_streams()` isn't really use anywhere else apart from the output, but since performance difference is negligible, in the end I used `BTreeMap`. It seems to be a neater solution than sorting inside `Minidump::unknown_streams()` and `Minidump::unimplemented_streams()` (or my original implementation, which sorts the result of these function calls)

afranchuk:
Looks good to me! I think that's a reasonable outcome; now there won't be unexpected non-determinism in the future.

PR #1007, opened 2024-06-05
Update to framehop 0.12.2.

afranchuk:
This removes the `thiserror` and `thiserror-no-std` dependencies. We also internalize the `object` `ModuleSectionInfo` impl for now as it'll be removed from `framehop` and `framehop` uses an incompatible version of `object` right now.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8,8 +8,8 @@ version = "0.22.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "6e4503c46a5c0c7844e948c9a4d6acd9f50cccb4de1c48eb9e291ea17470c678"
 dependencies = [
- "fallible-iterator 0.3.0",
- "gimli",
+ "fallible-iterator",
+ "gimli 0.29.0",
 ]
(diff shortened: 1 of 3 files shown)


gabrielesvelto (approved):
LGTM

afranchuk:
@gabrielesvelto note that this duplicates `gimli` versions until `wholesym` gets a new release (that will happen shortly).

afranchuk:
@gabrielesvelto also note that though it produces duplicate `gimli` versions, that's only when the `debuginfo-symbols` feature is enabled (so it won't be a problem for the `minidump-unwind` use in minidump-analyzer). I'm not sure whether you're okay with that right now to make a new release or whether you would prefer to wait for other things to be updated (backtrace already [merged](https://github.com/rust-lang/backtrace-rs/commit/22bb95bd76ec52daaa742d4933950b02f78d6e25) my updates to `object` and are making a new release shortly from the looks of it).

gabrielesvelto:
I don't mind some short-term duplication, especially if we know it's going away soon.

PR #1011, opened 2024-06-17
Fix symfiles containing newlines at the end of CFI sections failing to parse

### LINKED ISSUE #1010: Failure to parse valid symfile
The CFI parsing (and other sub-item parsing that happens at the end of the file) has an issue causing it to not properly ignore empty lines, leading to a symfile with two empty lines at the end after an CFI block to fail parsing even though it is a perfectly valid one:

```
[…]
STACK CFI INIT 8c80 1110 .cfa: sp 0 + .ra: x30
STACK CFI 8c9c .cfa: sp 336 + .ra: .cfa -8 + ^ x19: .cfa -24 + ^ x20: .cfa -32 + ^ x21: .cfa -40 + ^ x22: .cfa -48 + ^ x23: .cfa -56 + ^ x24: .cfa -64 + ^ x25: .cfa -72 + ^ x26: .cfa -80 + ^ x27: .cfa -88 + ^ x28: .cfa -96 + ^ x29: .cfa -16 + ^


```

The reason for this is, that the sub-item parsing code will jump back to the upper-level parser and continue parsing lines with that, however if we are at the end of the file, there is no line to parse anymore so the trailing empty line handling added in #742 does not work for this edge-case.

ePirat:
Fix #1010

### DIFF
--- a/breakpad-symbols/src/sym_file/parser.rs
+++ b/breakpad-symbols/src/sym_file/parser.rs
@@ -3,7 +3,7 @@
 
 use nom::branch::alt;
 use nom::bytes::complete::{tag, take_while};
-use nom::character::complete::{hex_digit1, multispace0, space1};
+use nom::character::complete::{hex_digit1, space1};
 use nom::character::{is_digit, is_hex_digit};
 use nom::combinator::{cut, map, map_res, opt};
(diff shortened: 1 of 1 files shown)


gabrielesvelto (approved):
Thanks a lot

PR #1013, opened 2024-06-18
Fallback to a module's code_file info for empty debug_file

ePirat:
If the debug_file metadata is empty, fallback to code_file to allow
to maybe still find a symfile instead of hopelessly looking for a
broken path (<empty string>.sym).

This fixes symbol file lookup for minidumps generated from GCC-built
Windows binaries, as GCC does not support PDB generation.

Additionally add some more trace logging to the paths lookups to make it
easier to debug such issues.


gabrielesvelto (changes requested):
Could you add a small test? I would expect for this case for `module.debug_file()` to be `None` rather than an empty string so I'd like to see how does this work in practice, I suppose I'm missing something.

Could you also update the function description to reflect the change? You can also drop the line that says

> `extension` is the expected extension for the symbol filename, generally *sym* if Breakpad text format symbols are expected.

Because that's clearly not valid anymore.

ePirat:
Thanks for the review!

> Could you add a small test? I would expect for this case for module.debug_file() to be None rather than an empty string so I'd like to see how does this work in practice, I suppose I'm missing something.

I am not sure whats the best way to do that, should I just generate a minidump from VLC and put that somewhere in the repository to use for the test? Or is there some better way to generate actually small "dummy" minidumps for tests?

>> extension is the expected extension for the symbol filename, generally sym if Breakpad text format symbols are expected.
> Because that's clearly not valid anymore.

How is that not valid anymore though? Nothing about the extension changes, just the file name that is looked up.

gabrielesvelto:
> I am not sure whats the best way to do that, should I just generate a minidump from VLC and put that somewhere in the repository to use for the test? Or is there some better way to generate actually small "dummy" minidumps for tests?

The `minidump-synth` crate allows you to generate dummy minidumps for testing, see how it's used in the other tests.

> How is that not valid anymore though? Nothing about the extension changes, just the file name that is looked up.

We used to have an argument called `extension` but now we don't, it's been refactored out of the function a while ago and now we're always using `.sym` as the extension from what I can see.

luser:
> I am not sure whats the best way to do that, should I just generate a minidump from VLC and put that somewhere in the repository to use for the test? Or is there some better way to generate actually small "dummy" minidumps for tests?

I assume you have a minidump that reproduces the issue you're trying to fix, right? If you run `minidump-stackwalk --dump` on it and paste a snippet of the module list (a single entry from a mingw-built module should be sufficient) we can advise you on how to replicate the contents in a unit test.

ePirat:
@luser Here is an example module entry:

```txt
module[0]
MINIDUMP_MODULE
  base_of_image                   = 0x720000
  size_of_image                   = 0xf2000
  checksum                        = 0xec763
  time_date_stamp                 = 0x30b530ae 1995-11-24T02:40:46Z
  module_name_rva                 = 0x513a
  version_info.signature          = 0xfeef04bd
  version_info.struct_version     = 0x10000
  version_info.file_version       = 0x30000:0x140000
  version_info.product_version    = 0x30000:0x140000
  version_info.file_flags_mask    = 0x0
  version_info.file_flags         = 0x0
  version_info.file_os            = 0x4
  version_info.file_type          = 0x1
  version_info.file_subtype       = 0x0
  version_info.file_date          = 0x0:0x0
  cv_record.data_size             = 25
  cv_record.rva                   = 0xe3d2
  misc_record.data_size           = 0
  misc_record.rva                 = 0x0
  (code_file)                     = "C:\Program Files (x86)\VideoLAN\VLC\vlc.exe"
  (code_identifier)               = "30b530aef2000"
  (cv_record).cv_signature        = 0x53445352
  (cv_record).signature           = b27d686d-9181-08cc-368d-b67fb54c4630
  (cv_record).age                 = 1
  (cv_record).pdb_file_name       = ""
  (misc_record)                   = (null)
  (debug_file)                    = ""
  (debug_identifier)              = "b27d686d-9181-08cc-368d-b67fb54c4630-1"
  (version)                       = "3.0.20.0"
```

I assume I need to create a `ModuleCrashpadInfo`? However I am unsure about the differences between `add_annotation_object`, `add_simple_annotation` and `add_list_annotation` and which is the right one to test my case.

luser:
> I would expect for this case for `module.debug_file()` to be `None` rather than an empty string so I'd like to see how does this work in practice, I suppose I'm missing something.

Unfortunately for PDB7.0 debug info the PDB filename is represented in the minidump as a nul-terminated string, not an offset to a string (which could be `null`) like most other strings: https://github.com/rust-minidump/rust-minidump/blob/d235cbbfb73c4f985214d08dee04560d7ea2ae67/minidump-common/src/format.rs#L506

luser:
> I am not sure whats the best way to do that, should I just generate a minidump from VLC and put that somewhere in the repository to use for the test? Or is there some better way to generate actually small "dummy" minidumps for tests?

Upon further review, since you're making changes in the `breakpad-symbols` crate, you should be able to avoid all of that and just write a test in the style of some of the existing tests in that crate, such as `test_code_info_breakpad_sym_lookup`:
https://github.com/rust-minidump/rust-minidump/blob/d235cbbfb73c4f985214d08dee04560d7ea2ae67/minidump-common/src/format.rs#L506

You can see that it uses a `SimpleModule` struct which implements the same required trait as `MinidumpModule`, but allows you to directly specify the module info. Your test should invoke `breakpad_sym_lookup`, since that's what you're changing.

gabrielesvelto:
@ePirat I don't want to hold this up for too long, if you don't have time to add a test case I'll merge the PR as is and file an issue for the tests as a follow up

PR #1014, opened 2024-06-19
Implement inconsistent crash detection

### LINKED ISSUE #704: Detect crashes with inconsistent/impossible crash addresses
A recent Firefox crash spike (bug [1796282](https://bugzilla.mozilla.org/show_bug.cgi?id=1796282)) was caused by a CPU bug. This took a while to figure out as we have only a handful of people within Mozilla that can identify those. However the crashes had an obvious hint that this crash could have been impossible: the disassembled crash address was valid (and not a non-canonical address like it appeared in many crashes) and more importantly it fell within an accessible and readable region of memory, so the crash reason (`EXCEPTION_ACCESS_VIOLATION_READ`) coupled with that address made no sense.

Given we have the memory info list available in each minidump we could detect these instances and flag these crashes as impossible or inconsistent so that they're easier to analyze.

Note that there's more hints of inconsistency that could be detected: `EXCEPTION_ACCESS_VIOLATION_EXEC` crashes in areas that are executable, or `EXCEPTION_ACCESS_VIOLATION_*` crashes when the disassembled instruction doesn't contain any memory accesses (we've seen those in CPU bugs too for example, the crashing instruction was a jump).

BrianShTsoi:
Aims to resolve #704


minidump-processor/src/op_analysis.rs
```diff
@@ -18,6 +18,7 @@
 
 use minidump::{MinidumpContext, MinidumpRawContext, UnifiedMemory};
 use std::collections::BTreeSet;
+use yaxpeax_x86::amd64::Instruction;
```
luser, line 21: This needs to be gated behind `#[cfg(feature = "disasm_amd64")]` like the other bits in this module, since `yaxpeax-x86` is an optional dependency tied to that feature.

minidump-processor/src/process_state.rs
```diff
@@ -217,12 +218,18 @@ pub struct ExceptionInfo {
     pub adjusted_address: Option<AdjustedAddress>,
     /// A string representing the crashing instruction (if available)
     pub instruction_str: Option<String>,
+    /// INCOMPLETE: Crashing instruction, later can replace instruction_str
+    pub instruction: Option<Instruction>,
```
luser, line 222: Putting this here makes this part of [this crate's public API](https://docs.rs/minidump-processor/latest/minidump_processor/struct.ExceptionInfo.html), which isn't great since it's a type from an external crate. It's also an amd64-specific type which doesn't make sense for this struct which is used for minidumps from any architecture.

The `yaxpeax-arch` crate has an [`Instruction` trait](https://docs.rs/yaxpeax-arch/0.3.1/yaxpeax_arch/trait.Instruction.html) that is implemented for this struct (and the parallel structs in other arch-specific yaxpeax crates), but it doesn't provide much functionality.

I'd suggest wrapping this in a newtype or an enum, even if it doesn't look very useful right now, to make it clear that this is an amd64 instruction and avoid directly exposing `yaxpeax` types in this crate's API.

Also, to reiterate the point from my previous comment: this dependency is currently marked as optional, so you need to account for that.

BrianShTsoi:
Restructured my code to keep `yaxpeax-x86` where it was (inside `amd64` module in `op_analysis.rs`). Added some enums & derivation of memory operation from `mov`.

Some notes below, mostly for myself and when I ask for suggestions.
I still need to fix:
* Move `bitflip::MemoryOperation` out of `bitflip`, and use it in `op_analysis.rs` as well?
* Consider `AdjustedAddress`?
* Separate "no memory operation" from `Unknown`, and properly handle `Unknown`

I still need to add:
* Handle memory operations in non-windows cases
* Detect cases where crash reason's memory access is not the address found in instruction (by checking the `memory_accesses` vector?)
* More instructions than just `mov`
* Human and json output
* Tests

minidump-processor/src/op_analysis.rs
```diff
@@ -48,6 +50,12 @@ pub enum OpAnalysisError {
 pub struct OpAnalysis {
     /// A string representation of the instruction for humans to read
     pub instruction_str: String,
+    /// The memory operation performed by the instruction
+    ///
+    /// Note that `None` means that memory operation could not be determined,
+    /// `Some(NoOperation)` means it was successfully determined that the crash
+    /// doesn't access memory
+    pub memory_operation: Option<MemoryOperation>,
```
luser, line 58: I'm not sure, but it might be nicer to drop the `Option` and just add a variant like `MemoryOperation::Unknown`?

minidump-processor/src/op_analysis.rs
```diff
@@ -183,6 +194,29 @@ mod amd64 {
         ret
     }
 
+    impl MemoryOperation {
+        fn from_instruction(i: Instruction) -> Option<Self> {
```
luser, line 198: Maybe this should be more explicitly named, like `from_amd64_instruction`?

minidump-processor/src/process_state.rs
```diff
@@ -416,6 +421,48 @@ impl PossibleBitFlip {
...
+                )
+            }
+            CrashReasonInconsistency::InconsistentMemoryOperation {
+                crash_reason_operation,
+                instruction_operation,
+            } => {
+                write!(
+                    f,
+                    "Crash reason is {} but instruction is {}",
+                    crash_reason_operation.to_string(),
```
luser, line 458: The `to_string` calls here are redundant since these types impl `Display` (clippy will warn you about this depending on the configuration).

minidump-processor/src/processor.rs
```diff
@@ -1144,44 +1199,122 @@ fn try_detect_null_pointer_in_disguise(memory_accesses: Option<&[MemoryAccess]>)
...
+        NoOperation,
         Read,
         Write,
         Execute,
+        UnknownOperation,
+    }
+
+    impl std::fmt::Display for MemoryOperation {
+        fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+            write!(
```
luser, line 1217: FYI if you're just writing string constants you can [use `f.write_str` directly](https://doc.rust-lang.org/nightly/std/fmt/struct.Formatter.html#method.write_str).

luser (approved):
This looks reasonable to me from an implementation standpoint. I didn't review the analysis itself closely.

BrianShTsoi:
Thank you for the feedback! I've addressed them.

BrianShTsoi:
Todo:
* Any other opcodes that can be used to derive memory operation
* Check for inconsistencies with other crash reasons that may or may not be memory related
* Json output and schema
* Tests

minidump-processor/src/op_analysis.rs
```diff
@@ -183,6 +193,71 @@ mod amd64 {
...
+                Opcode::CALL
+                | Opcode::CALLF
+                | Opcode::JMP
+                | Opcode::JMPF
+                | Opcode::JMPE
+                | Opcode::RETURN
+                | Opcode::RETF
+                | Opcode::IRET
+                | Opcode::IRETD
+                | Opcode::IRETQ => Self::Execute,
```
BrianShTsoi, line 255: This is wrong and should be fixed. Exec violation should be checked with a different way. These instructions only modify `rip`. If there is access violation exec, it will crash in the instruction after.

minidump-processor/src/processor.rs
```diff
@@ -1144,44 +1200,117 @@ fn try_detect_null_pointer_in_disguise(memory_accesses: Option<&[MemoryAccess]>)
...
+                        || memory_info.is_executable()
+                }
+            }
+        }
+
+        /// Return whether this memory operation is definitely allowed in the given memory region.
+        /// If operation is `Undetermined`, this method returns false.
+        /// If operation is `UnknownOperation`, this method returns true if the given memory region
+        /// allows all operations, and returns false otherwise.
+        pub fn is_definitely_allowed_for(&self, memory_info: &UnifiedMemoryInfo) -> bool {
```
afranchuk, line 1279: We could probably just call this `is_allowed_for`.

minidump-processor/src/processor.rs
```diff
@@ -1144,44 +1200,117 @@ fn try_detect_null_pointer_in_disguise(memory_accesses: Option<&[MemoryAccess]>)
...
+                Self::Execute => memory_info.is_executable(),
+                Self::UnknownOperation => {
+                    memory_info.is_readable()
+                        && memory_info.is_writable()
+                        && memory_info.is_executable()
+                }
+            }
+        }
+
+        pub fn consistent_operation(
```
afranchuk, line 1294: Add a doc comment!

afranchuk (approved):
Looks pretty good, just a few suggestions!

BrianShTsoi:
Done
* Removed memory operation derivation from instructions. It makes more sense to treat memory operations on a per operand basis. Restructured accordingly. 
* Derivation of the direction of memory access,  privileged instructions and datatype alignments will be more complete once `yaxpeax` implements the relevant functionality. For now, enumerate through the most common opcodes.
* Added handling of some non memory access related exceptions

Todo
* Rewrite how the `memory_accesses` vector is generated and separate out `indirect` accesses (which are really `rip` changes rather than memory accesses)
* Add implicit memory accesses
* Add memory access direction derivation for more opcodes 
* Human output
* JSON output and schema
* Tests

minidump-processor/src/op_analysis.rs
```diff
@@ -330,46 +622,116 @@ mod amd64 {
...
+        pub fn get_rip_update_from_instruction(
+            self,
+            decoded_instruction: Instruction,
+        ) -> Result<Option<MemoryAddressInfo>, OpAnalysisError> {
+            let rip_update = |address| {
+                Some(MemoryAddressInfo {
+                    address,
+                    is_likely_null_pointer_dereference: address == 0,
+                    is_likely_guard_page: false,
+                    size: Some(1),
```
marti4d, line 698: See -- The fact that there is nothing reasonable for you to put in these two fields is part of the reason why I think a split between `MemoryAddressInfo` and `MemoryAccess` is needed.

minidump-processor/src/op_analysis.rs
```diff
@@ -330,46 +622,116 @@ mod amd64 {
...
+                | CommonOpcode::MOV
+                | CommonOpcode::MOVAPS
+                | CommonOpcode::MOVUPS
+                | CommonOpcode::SUB => (),
+            }
+            Ok(())
+        }
+    }
+
+    impl<'a> GetInstructionPointerUpdate<'a> {
```
marti4d, line 674: This is another example of a verb masquerading as a noun. Perhaps it makes more sense to have this:

```Rust
struct InstructionPointerUpdate {
    address_info: MemoryAddressInfo,
}

impl InstructionPointerUpdate {
    pub fn from_instruction(
        context: &MinidumpContext,
        memory_list: Option<&minidump::UnifiedMemoryList<'_>>,
        stack_memory: Option<minidump::UnifiedMemory<'_, '_>>,
        decoded_instruction: Instruction,
    ) -> Result<Option<InstructionPointerUpdate>, OpAnalysisError> {
    }
}
```

minidump-processor/src/processor.rs
```diff
@@ -809,6 +832,137 @@ impl<'a> MinidumpInfo<'a> {
         }
     }
 
+    /// Check for inconsistencies between crash reason and crashing instruction
+    pub fn check_for_crash_inconsistencies(&self, exception_details: &mut ExceptionDetails<'a>) {
+        use minidump_common::errors::ExceptionCodeMacArithmeticPpcType as MacArithPpc;
```
marti4d, line 837: nit: Try grouping these `use` statements so they don't occupy quite so much room:

```Rust
use minidump_common::errors::{
    ExceptionCodeMacArithmeticPpcType as MacArithPpc, 
    ExceptionCodeMacArithmeticX86Type as MacArithX86,
    // ...
};

**Review by marti4d — CHANGES_REQUESTED** [review]
It's starting to take some pretty good shape, Brian! I think there's more logic needed in determining the operand access types, and I've got some style/structure notes, but this is already looking quite promising :)

File: minidump-processor/src/op_analysis.rs [hunk 13]
```diff
@@ -48,19 +48,24 @@ pub enum OpAnalysisError {
...
+    /// Whether the instruction pointer is being updated by the instruction
+    pub instruction_pointer_update: Option<MemoryAddressInfo>,
     /// A list of all registers which were used by this instruction.
     pub registers: BTreeSet<&'static str>,
 }
 
 /// Details about a memory access performed by an instruction
 #[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
-pub struct MemoryAccess {
+pub struct MemoryAddressInfo {
```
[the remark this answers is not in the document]
BrianShTsoi: Yep this makes sense. I'll split them into 2 structs separate.

minidump-processor/src/op_analysis.rs
```diff
@@ -183,8 +227,214 @@ mod amd64 {
...
+            !matches!(self, Self::UncommonInstructionAccess)
+        }
+        // TODO: Derive access type using `yaxpeax` instead
+        fn explicit_from_instruction(instruction: Instruction, index: u8) -> Option<Self> {
+            let Some(opcode) = CommonOpcode::from_amd64_opcode(instruction.opcode()) else {
+                return Some(Self::UncommonInstructionAccess);
+            };
+            match opcode {
+                CommonOpcode::ADD | CommonOpcode::SUB => {
+                    if index == 0 {
```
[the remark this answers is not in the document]
BrianShTsoi: Technically this is accounted for? since `OperandAccessType::explicit_from_instruction` was only called (in `GetMemoryAccess::explicit_access`) when it is determined the operand at that index is a memory operand.
But yea the structure is unnecessarily unintuitive. 
Right now it loops through the operands, checks for memory operands, then looks at the opcode and determine the access direction.
A better structure would be to look at the opcode first, and then loop through the operands and determine access types (if any). I'll change it to that.

minidump-processor/src/op_analysis.rs
```diff
@@ -255,67 +544,70 @@ mod amd64 {
     }
 
     impl<'a> GetMemoryAccess<'a> {
```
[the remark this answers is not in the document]
BrianShTsoi: Though in this case I guess I don't even need wrap it around a struct, I can just return the vector directly.

BrianShTsoi:
Reorganized much of `op_analysis.rs` according to feedback.
Also changed the code behavior as a fix of [Bug 1909661](https://bugzilla.mozilla.org/show_bug.cgi?id=1909661). The code now treats U32 displacement as a sign extended 64-bit memory access (as specified in Intel's Reference Manual). 
This is relevant to the inconsistency detection feature, as we need `memory_accesses` to be accurate in order to check if the implied access  from the crash address + crash reason is actually performed by the instruction.

minidump-processor/src/op_analysis.rs
```diff
@@ -154,222 +234,341 @@ mod amd64 {
...
             accesses: &mut Vec<MemoryAccess>,
-            decoded_instruction: &Instruction,
+            opcode: CommonOpcode,
+            instruction: Instruction,
+            context: &MinidumpContext,
         ) -> Result<(), OpAnalysisError> {
-            // Shortcut -- If the instruction doesn't access memory, just return an empty list
-            let mem_size = match decoded_instruction.mem_size() {
+            // Shortcut -- If the instruction doesn't access memory, just return
+            let mem_size = match instruction.mem_size() {
```
marti4d, line 375: Hmm... I have a concern here -- I'm not totally sure that Yaxpeax will account for implicit memory accesses with this method. If not, then you will exit too early here without considering implicit accesses.

Perhaps you could check that this does what you want it to?

minidump-processor/src/op_analysis.rs
```diff
@@ -154,222 +234,341 @@ mod amd64 {
...
-                    }),
-                    other_operand => {
-                        if let Some(op_info) = RegOperandInfo::try_from_operand(other_operand) {
-                            Some(self.calculate_address(mem_size, op_info)?)
-                        } else {
-                            None
-                        }
-                    }
-                };
+            Self::add_common_opcode_implicit_access(accesses, opcode, mem_size, context)?;
```
marti4d, line 391: Assuming that Yaxpeax doesn't account for implicit accesses, that means that `mem_size` is not useful here, and you will have to determine the memory access size based on the opcode (I think almost all of them will be 8 bytes).

minidump-processor/src/op_analysis.rs
```diff
@@ -48,30 +48,110 @@ pub enum OpAnalysisError {
...
+}
+
+pub trait MemoryAccessVecExt {
+    fn contains_access(&self, address: u64, access_type: MemoryAccessType) -> bool;
+}
+
+impl MemoryAccessVecExt for Vec<MemoryAccess> {
+    fn contains_access(&self, address: u64, access_type: MemoryAccessType) -> bool {
+        self.iter().any(|access| {
+            let size = access.size.unwrap_or(64) as u64;
```
marti4d, line 128: I'm going to go out on a limb here and say that you were probably thinking _bits_, but that number is in **bytes**. That's a pretty big memory access for one instruction! 😂

As for how to handle this -- Just `return false`. There's only a few rare, super-weird instructions on x64 that don't know the size of memory they access, and we don't really have to care about them.

minidump-processor/src/op_analysis.rs
```diff
@@ -154,222 +234,341 @@ mod amd64 {
...
+                        is_likely_null_pointer_dereference: address == 0,
+                        is_likely_guard_page: false,
+                    },
+                })
+            };
+
+            match instruction.opcode() {
                 Opcode::CALL | Opcode::CALLF | Opcode::JMP | Opcode::JMPF | Opcode::JMPE => {
-                    if decoded_instruction.operand_count() != 1 {
+                    if instruction.operand_count() != 1 {
```
marti4d, line 548: This actually seems worth using `assert!()` for -- If this is false, there is a bug in Yaxpeax that needs fixing and we want it to fail loudly.

marti4d (changes requested):
This is already looking really good, Brian! I think there are a couple of minor bugs I've noticed, and I've got some more suggestions for improving structure and readability, but woah this is really taking shape quite nicely!

minidump-processor/src/op_analysis.rs
```diff
@@ -48,30 +48,110 @@ pub enum OpAnalysisError {
...
 /// Details about a memory access performed by an instruction
 #[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
 pub struct MemoryAccess {
-    /// The address of the memory access
-    pub address: u64,
-    /// Whether or not this memory access is likely the result of a null-pointer dereference
-    pub is_likely_null_pointer_dereference: bool,
-    /// Whether or not this memory access was part of a likely guard page.
-    pub is_likely_guard_page: bool,
+    // Information
```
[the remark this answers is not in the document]
BrianShTsoi: Oh yea for sure, I probably got distracted by something else when I was writing it and then forgot about it🤦

minidump-processor/src/op_analysis.rs
```diff
@@ -48,30 +48,110 @@ pub enum OpAnalysisError {
...
+pub enum MemoryAccessType {
+    Read,
+    Write,
+    ReadWrite,
+
+    // TODO: Remove this variant once `yaxpeax` is used to derive access type of every operand
+    UncommonInstructionAccess,
+}
+
+pub trait MemoryAccessVecExt {
```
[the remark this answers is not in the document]
BrianShTsoi: I see. That makes sense!

minidump-processor/src/op_analysis.rs
BrianShTsoi: Yaxpeax gives `mem_size` for `ret`, `call`, `push` and `pop` and the likes, so this should be fine. For some reason it gives `mem_size` for `retf` but not `iretd`. I'll further investigate.

minidump-processor/src/op_analysis.rs
```diff
@@ -154,222 +234,341 @@ mod amd64 {
...
+        ) -> Result<(), OpAnalysisError> {
+            if !operand.is_memory() {
+                return Ok(());
+            }
 
-                if let Some(access) = maybe_access {
-                    accesses.push(access);
+            let access_type = match (opcode, idx) {
+                (CommonOpcode::ADD | CommonOpcode::SUB, 0) => MemoryAccessType::ReadWrite,
+                (CommonOpcode::ADD | CommonOpcode::SUB, 1) => MemoryAccessType::ReadWrite,
```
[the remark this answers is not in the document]
BrianShTsoi: Oh yea good catch, silly me🤦

minidump-processor/src/op_analysis.rs
```diff
@@ -154,222 +234,341 @@ mod amd64 {
...
+                }
+                (CommonOpcode::LEA, 0 | 1) => return Ok(()),
+                _ => {
+                    tracing::warn!("instruction had unexpected memory operand");
+                    return Ok(());
                 }
+            };
+
+            if let Some(op_info) = MemoryOperandInfo::try_from_operand(operand) {
+                let address_info = MemoryAddressInfo::from_memory_operand_info(op_info, context)?;
```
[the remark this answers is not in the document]
BrianShTsoi: Ah ok this makes sense

minidump-processor/src/op_analysis.rs
BrianShTsoi: So it seems that for yaxpeaax doesn't give `mem_size` for `iretd` and `iretq`, but does so for the other common opcodes that have implicit access 🤔. I'm also not sure if it even makes sense here to claim that `iret` access the memory pointed by the instruction pointer.

minidump-processor/src/op_analysis.rs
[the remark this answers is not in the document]
BrianShTsoi: I will create `MemoryAddressInfo::try_from_operand()`, which calls `MemoryOperandInfo::try_from_operand()` and proceed with the logic of the original `MemoryAddressInfo::from_memory_operand_info()`. 
`MemoryOperandInfo::try_from_operand()` will remain a separate function since it is also used in `get_register()`

minidump-processor/src/process_state.rs
```diff
@@ -416,6 +422,37 @@ impl PossibleBitFlip {
     }
 }
 
+#[derive(Debug, Clone)]
+pub enum CrashReasonInconsistency {
+    IntDivByZeroNotPossible,
+    PrivInstructionCrashWithoutPrivInstruction,
+    NonCanonicalAddressFalselyReported,
+    AccessViolationWhenAccessAllowed,
+    CrashingAccessNotFoundInMemoryAccesses,
```
gabrielesvelto, line 431: We want these to be machine-readable in the JSON output so it's fine for them to be descriptive labels but not quite human-readable phrases. However we use snake-case names for the rest of the JSON output, so it'd be better to print them out that way for consistency.

minidump-processor/src/process_state.rs
```diff
@@ -416,6 +422,37 @@ impl PossibleBitFlip {
...
+}
+
+impl std::fmt::Display for CrashReasonInconsistency {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            CrashReasonInconsistency::IntDivByZeroNotPossible => {
+                f.write_str("Crash reason is integer division by zero but instruction is not")
```
gabrielesvelto, line 438: "Crash reason is an integer division by zero but the crashing instruction is not a division"

minidump-processor/src/process_state.rs
```diff
@@ -416,6 +422,37 @@ impl PossibleBitFlip {
...
+            }
+            CrashReasonInconsistency::PrivInstructionCrashWithoutPrivInstruction => {
+                f.write_str("Crash reason is priveleged instruction but instruction is not")
```
gabrielesvelto, line 441: "Crash reason is a privileged instruction but crashing instruction is not a privileged one"

gabrielesvelto (changes requested):
A bit of bike shedding on my part WRT to the type/field/enum names plus the changes to the JSON code need to be reflected in the JSON schema file under `minidump-processor/json-schema.md`

minidump-processor/src/process_state.rs
```diff
@@ -416,6 +422,62 @@ impl PossibleBitFlip {
...
+                f.write_str("Crash reason is access violation exception but access is allowed")
+            }
+            CrashInconsistency::CrashingAccessNotFoundInMemoryAccesses => f.write_str(
+                "Crash address not found among the memory accesses of the crashing instruction",
+            ),
+        }
+    }
+}
+
+impl serde::Serialize for CrashInconsistency {
```
afranchuk, line 456: Usually people don't directly implement `Serialize` unless necessary. In this case, there are a few alternatives using `#[derive(Serialize)]`:
1. You can use `#[serde(rename_all = "snake_case")]` on the `enum` item, which I believe will result in the same serialization as implemented here.
2. You can use `#[serde(into = "String")]` and add `impl From<CrashInconsistenty> for String` (though this isn't an obvious conversion so this is less preferable).

minidump-processor/json-schema.md
```diff
@@ -177,6 +186,19 @@ anyway.
...
+    /// A list of inconsistencies detected by comparing crash reason/crash address with 
+    /// crashing instruction and memory information
+    "crash_inconsistencies": [
+      {
+        "inconsistency": "int_div_by_zero_not_possible"
+          | "priv_instruction_crash_without_priv_instruction"
+          | "non_canonical_address_falsely_reported"
+          | "access_violation_when_access_allowed"
+          | "crashing_access_not_found_in_memory_accesses"
+      }
```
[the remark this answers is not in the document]
BrianShTsoi: Hmm originally I thought it might also output other information like the access direction of the instruction vs the access direction of the crash, but it's shown elsewhere so it's not really necessary. Now that you mention it I can probably just leave this as a list of strings.

minidump-processor/src/op_analysis.rs
```diff
@@ -154,222 +225,402 @@ mod amd64 {
...
+                return Ok(());
+            }
 
-                if let Some(access) = maybe_access {
-                    accesses.push(access);
+            let access_type = match opcode {
+                AccessDerivableOpcode::ADD | AccessDerivableOpcode::SUB => match idx {
+                    0 => MemoryAccessType::ReadWrite,
+                    1 => MemoryAccessType::Read,
+                    _ => panic!("add/sub instruction had unexpected memory operand"),
```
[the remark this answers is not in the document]
BrianShTsoi: Yea I don't think it's guaranteed by yaxpeax, but it's guaranteed by the spec and so it implies a bug in yaxpeax.

I think @marti4d mentioned that if there is a yaxpeax bug, we definitely want to find out, and so we want to panic and fail loudly. 

On one hand just returning an error makes sense in that technically only the instruction/disassembler is problematic, and the rest of the `minidump-stackwalk` is running fine. On the other hand panicking can raise our attention quicker about a disassembler bug (I'm not too sure how much more noticeable? Not too familiar with how crash-stat actually handles panic vs just an error + `tracing::warn!`)

minidump-processor/src/op_analysis.rs
```diff
@@ -378,69 +629,241 @@ mod amd64 {
...
+        MOV,
+        MOVAPS,
+        MOVUPS,
+        POP,
+        PUSH,
+        RETF,
+        RETURN,
+        SUB,
+        /// See https://bugzilla.mozilla.org/show_bug.cgi?id=1831370
+        UCOMSISS,
```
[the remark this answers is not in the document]
BrianShTsoi: This is an embarrassing typo🤦

minidump-processor/src/op_analysis.rs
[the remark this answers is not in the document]
afranchuk: Yeah let's leave it as a panic.

minidump-processor/src/op_analysis.rs
```diff
@@ -378,69 +629,218 @@ mod amd64 {
...
-            &self,
-            access_size: Option<u8>,
-            register_operand_info: RegOperandInfo,
-        ) -> Result<MemoryAccess, OpAnalysisError> {
-            let mut access = MemoryAccess {
+    impl MemoryOperandInfo {
+        pub fn try_from_operand(op: Operand) -> Option<Self> {
+            let mut info = MemoryOperandInfo::default();
+            match op {
+                Operand::DisplacementU32(disp) => info.disp = Some(disp as i32 as i64),
```
BrianShTsoi, line 737: Note: This results in a different behaviour from the current `Rust-minidump`. 
Referring to [Bug 1909661](https://bugzilla.mozilla.org/show_bug.cgi?id=1909661), `yaxpeax` gives the displacement as a `u32`, but the CPU will treat the displacement as a signed number. It is debatable whether `yaxpeax`'s interpretation is correct. Currently `Rust-minidump` treats the displacement as a `u32` and cast it to a `u64` address (ie. no sign extension), resulting in a different address than the one actually being accessed.
Since this features require the memory address to be accurate, this implementation sign extends the displacement by casting the `u32` to an `i32` before then casting to `i64` to obtain the actual address accessed.

marti4d (commented):
This looks great, Brian! I just have the one question I was curious about, but as far as I'm concerned this is ready-to-merge!

minidump-processor/src/processor.rs
```diff
@@ -806,7 +823,173 @@ impl<'a> MinidumpInfo<'a> {
...
+        let crash_address = info.address.0;
+        // `adjusted_address` might not show that there is non-canonical address if there is
+        // also a disguised null pointer, so we check `memory_access_list` directly
+        if info
+            .instruction_properties
+            .as_ref()
+            .is_some_and(|properties| properties.is_access_derivable)
+            && info.memory_access_list.as_ref().is_some_and(|access_list| {
+                !access_list.contains_access(crash_address, MemoryAccessType::Read)
+                    && !access_list.contains_access(crash_address, MemoryAccessType::ReadWrite)
```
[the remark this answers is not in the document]
BrianShTsoi: Ah these are here because there are cases where it crashed with read violation at 0xffffffffffffffff, which is supposed to represent a non-canonical crash, but the instruction is actually reading 0xffffffffffffffff (and so it is consistent even though there is no use of non-canonical address), so these are added to distinguish that. 

But after you pointed out this I do realize this is platform specific and putting them here is out of place. I'll rename a few functions and reorganize to make it more sensible.

marti4d (approved):
Awesome! You know what @BrianShTsoi -- I think this thing is ready-to-land! Ship it! 🎉

marti4d:
@gabrielesvelto -- I think we are looking good-to-go here :)

gabrielesvelto (approved):
This is great work Brian, and thanks to everybody who helped & chimed in.

BrianShTsoi:
Thank you everyone!

PR #1015, opened 2024-06-21
Update a number of dependencies to remove some duplicates.

afranchuk:
Notably this still ends up with duplicated addr2line (and thus gimli) due to backtrace using a slightly older version.


afranchuk:
This might be good enough and the best we can hope for before a new release.

gabrielesvelto (approved):
LGTM

PR #1019, opened 2024-08-06
Add support for Fat MachO files in the minidump-unwind debuginfo symbol provider

afranchuk:
I have not tested this yet, but it should do the trick. I will test it out through `minidump-stackwalk`.

### DIFF
--- a/minidump-unwind/src/symbols/debuginfo.rs
+++ b/minidump-unwind/src/symbols/debuginfo.rs
@@ -5,6 +5,7 @@ use cachemap2::CacheMap;
 use framehop::Unwinder;
 use memmap2::Mmap;
 use minidump::{MinidumpModuleList, MinidumpSystemInfo, Module};
+use object::read::{macho::FatArch, Architecture};
 use std::cell::UnsafeCell;
 use std::fs::File;
 use std::path::{Path, PathBuf};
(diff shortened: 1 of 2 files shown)


minidump-unwind/src/symbols/debuginfo.rs
```diff
@@ -424,6 +449,39 @@ fn load_unwind_module(module: &dyn Module) -> Option<(Mmap, framehop::Module<Mod
...
+            match arch.data(fatfile_data) {
+                Err(e) => {
+                    tracing::error!(
+                        "failed to read data from fat macho file {}: {e}",
+                        path.display()
+                    );
+                    None
+                }
+                Ok(data) => Some(data),
+            }
```
gabrielesvelto, line 480: You can probable use `map_or_else()` here.

gabrielesvelto (approved):
I left just a small nit, the rest is fine.

minidump-unwind/src/symbols/debuginfo.rs
afranchuk: That's one I've needed to reach for infrequently enough that I forget about it :smile:

gabrielesvelto (approved):

afranchuk:
I have verified this works.

PR #1032, opened 2024-09-25
Limit non-canonical inconsistency detection to selected opcodes

### LINKED ISSUE #1031: Account for other general protection fault possibilities when detecting inconsistencies in non-canonical crashes
As mentioned in the discussion following https://github.com/rust-minidump/rust-minidump/issues/732#issuecomment-2370428678, unaligned accesses (and other cases of general protection fault) causes the OS to report the same crash addresses and exceptions as a crash caused by the use of non-canonical addresses. Currently we can't deduce whether the instruction is performing an unaligned accesses, so we should limit the inconsistency detection to instructions whose only possible cause of GPF is using non-canonical addresses.

BrianShTsoi:
Resolves  #1031

Also renamed a few functions/variables to better distinguish between general protection fault and non-canonical crash.

### DIFF
--- a/minidump-processor/src/op_analysis.rs
+++ b/minidump-processor/src/op_analysis.rs
@@ -75,6 +75,14 @@ pub struct InstructionProperties {
     pub is_access_derivable: bool,
     pub is_division: bool,
     pub is_privileged: bool,
+
+    // TODO: remove this field once we properly account for other causes of GPF (eg. unaligned access)
+    /// This field is used to support detecting inconsistencies in non-canonical crashes
+    /// True means that the instruction only gives General Protection Fault when non-canonical address is used
+    /// False means that GPF can be caused by other cases, or that it is undetermined
+    /// Since we only detect inconsistencies in non-canonical crashes if it is an `AccessDerivableOpcode`
+    /// This field is false for opcodes that are not `AccessDerivableOpcode`
+    pub is_only_gpf_when_non_canonical: bool,
 }
 
 #[derive(Clone, Debug)]
@@ -278,6 +286,8 @@ mod amd64 {
                 is_access_derivable: is_access_derivable(instruction.opcode()),
                 is_division: InstructionProperties::is_division(instruction),
                 is_privileged: InstructionProperties::is_privileged(instruction),
+                is_only_gpf_when_non_canonical:
+                    InstructionProperties::is_only_gpf_when_non_canonical(instruction),
             }
         }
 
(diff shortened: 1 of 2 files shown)


minidump-processor/src/op_analysis.rs
```diff
@@ -338,6 +348,18 @@ mod amd64 {
...
+
+        /// Since we only detect inconsistencies in non-canonical crashes if we can derive all its access,
+        /// this function always return false for opcodes that are not `AccessDerivableOpcode`
+        fn is_only_gpf_when_non_canonical(instruction: Instruction) -> bool {
+            let Some(opcode) = AccessDerivableOpcode::from_opcode(instruction.opcode()) else {
+                return false;
+            };
+            !matches!(
+                opcode,
+                AccessDerivableOpcode::MOVAPS | AccessDerivableOpcode::UCOMISS
```
gabrielesvelto, line 360: `MOVAPS` description in Intel manual states that it raises a general-protection fault when the address is unaligned, but `UCOMISS` instruction doesn't mention any alignment requirements, nor that it raise #GP for other conditions. Do we have crashes where the crashing instruction is `UCOMISS` and we get a #GP without a non-canonical address? IIRC we had some odd crashes involving `UCOMISS` but they were cause by a bug in QEMU, not by actual CPU behavior.

gabrielesvelto (commented):
This is mostly OK, I just left a comment to address.

minidump-processor/src/op_analysis.rs
BrianShTsoi: Ah I was under the impression that `UCOMISS` has alignment requirements. Should have looked at the Intel manual more closely. And no I did not see any #GP crashes with `UCOMISS` without non-canonical address.
The QEMU bug causes an actual access violation instead of a GP, since it's accessing some extra bytes that it's not supposed to.

BrianShTsoi: @gabrielesvelto Is the QEMU bug the reason why you suggested in https://github.com/rust-minidump/rust-minidump/issues/732#issuecomment-2373323897 to drop non-canonical inconsistency detection for `UCOMISS`? If so it's probably fine to keep `UCOMISS` and only drop the detection for `MOVAPS`?

gabrielesvelto: Possibly, it was a mistake on my part.

BrianShTsoi: Ah ok, no worries. I'll change it to only dropping `MOVAPS`.

gabrielesvelto (approved):
Looking good

PR #1035, opened 2024-10-11
Add soft error stream for minidump-writer #31

marti4d:
This is a pretty simple change to allow Minidump-Writer to have the new soft error stream.

See: https://github.com/rust-minidump/minidump-writer/issues/31


marti4d:
Obviously the new stream needs to be actually integrated into the reader, but since the minidump reader will ignore streams it doesn't recognize, I consider that it can be done later once the minidump-writer code is done.

PR #1069, opened 2025-02-25
Add soft error stream to human and json output

marti4d:
Note: I wanted the snapshot tests to also test the new Soft Error stream, so I hex-edited one of the .dmp files and injected a fake soft error stream into it.

To do this, the steps I took were:

1. The minidump starts with a count of how many streams there are, so I increased that by one.
2. Then there is a "Stream Directory", which is an array of 12-byte entries of form `{ type: u32, size: u32, rva: u32 }`. I needed to add one, but there was not 12 bytes between the last entry in the array and the start of the first stream. So I moved the contents of the first stream by appending them to the end of the file and fixed up the first rva to point to it.
3. Now I was able to clobber the unused bytes from the original first stream to extend the Stream Directory by 12 bytes to point to my new `MozSoftErrors` stream. I set the `size` and `rva` to point to the end of the minidump file.
4. Finally, I appended the soft error bytes to the end of the minidump file.

As you can see from the snapshots, it worked great!


minidump/src/minidump.rs
```diff
@@ -3917,6 +3923,26 @@ impl<'a> MinidumpStream<'a> for MinidumpLinuxProcLimits<'a> {
...
+impl<'a> MinidumpStream<'a> for MinidumpSoftErrors<'a> {
+    const STREAM_TYPE: u32 = MINIDUMP_STREAM_TYPE::MozSoftErrors as u32;
+
+    fn read(
+        bytes: &'a [u8],
+        _all: &'a [u8],
+        _endian: scroll::Endian,
+        _system_info: Option<&MinidumpSystemInfo>,
+    ) -> Result<MinidumpSoftErrors<'a>, Error> {
+        let json_str = std::str::from_utf8(bytes).map_err(|_| Error::DataError)?;
```
marti4d, line 3935: Note: I would have rather parsed this all the way through to a `serde_json::Value` (since it is an error for this to be anything but valid JSON), but `serde_json` is not already a dependency of this crate, and I didn't think it worth adding it for this.

testdata/linux-mini.dmp
marti4d, line 1: See my general comments for an explanation of what I changed about this file.

minidump-common/src/format.rs
gabrielesvelto, line 1: I've put these fixes in a separate commit and pushed it to main so that clippy would stop complaining, you can remove them.

minidump/src/minidump.rs
gabrielesvelto: I think this is a good choice given this is the base crate and has traditionally only contained the raw unparsed contents of a minidump. We put semantics on them in minidump-processor.

gabrielesvelto (changes requested):
The changes LGTM but you have to describe the new stream in minidump-processor/json-schema.md. Also add a message to the `RELEASES.md` file.

minidump/src/minidump.rs
marti4d: Fair enough :) JSON is weird sometimes so it's hard to know where to draw the line -- It kind-of straddles the boundary between "having semantics" and "just being a wire format".

PR #1077, opened 2025-03-07
Use workspace settings for packages

marti4d:
Now that Cargo allows this, it seems better to just manage these settings in a single `Cargo.toml` file (especially the version!)


gabrielesvelto (approved):
LGTM

PR #1085, opened 2025-03-15
Use workspace Cargo.toml as SSOT for dependencies

marti4d:
Cargo keeps adding new features to make it easier for workspace members to use the workspace `Cargo.toml` file as the Single Source of Truth for things they all have in common.

This centralizes all the dependencies of the various workspace members, which eliminates version drift, removes a bunch of tedious manual editing and reviewing of individual `Cargo.toml` files, and allows us to do releases for this project by just editing a single `Cargo.toml` file 🎉


marti4d:
@gabrielesvelto If you want, you can double-check that I didn't miss anything, but the fact that `Cargo.lock` didn't change at all provides some comfort :)

marti4d:
@gabrielesvelto To answer a question you probably have (because I did as well): "Wait, so if you list it as a dependency in the workspace, does that mean it will always be compiled, even if it is an unused feature or a dev-dependency?"

The answer is -- Nope! Cargo is smart enough to Do the Right Thing™ here. It basically just treates the workspace `Cargo.toml` file like a store of metadata about requested versions and features, but if no actual workspace members use that dependency given their set of features, targets, etc then the dependency will not be downloaded, compiled, etc.

gabrielesvelto (approved):
This is a very nice cleanup, we should be good to pull this into m-c when we cut the next release of rust-minidump

gabrielesvelto:
@marti4d should I merge this now? I think it just needs rebasing.

PR #1087, opened 2025-03-17
Extend the scan range when coming from a CFI frame

Dav1dde:
I am currently exploring making JIT compiled frames unwind better. My main example I am working with is a Unity minidump which calls from native into (mono) JIT compiled code into more native code, where it eventually crashes.

Rough frames:

```
[Native: Bar] <- Crash
[Native: Foo]
[JIT: C]
[JIT: B]
[JIT: A]
[Native: `jit_invoke`]
```

All native frames have full unwind and debug symbols available, but I noticed that the stack scanner, did not extend far enough to find the native `jit_invoke` frame, it stopped in the middle of the JIT compiled code.

Expected Frames:

```
[Native: Bar]
[Native: Foo]
[0x00008123]
[Native: `jit_invoke`]
```

Actual Frames:

```
[Native: Bar]
[Native: Foo]
[0x00008123]
```

The idea is to default to the extended scan range, not only for pointers provided by the context, but also when we have a reasonably confidence that the provided values are valid.

Now this is would not be needed, if we extend the validity check to also be able to recognize JITed instructions. That requires another heuristic (e.g. proximity of the instruction addr to the previous instruction addr) or proper consideration of `MemoryInfoList` / `LinuxMaps` streams. With the real minidumps I have, these streams are *not* available (still looking into why crashpad doesn't seem to add the `MemoryInfoList` stream on Windows, for Linux there is straight up no support implemented).

In any case, extending the scan range seems to cover these cases.

---

I only added a test for `amd64` but made the change for all backends. Let me know if you want me to copy the same test multiple times or if I should only make the change in `amd64` (the platform which I actually currently work with).

### DIFF
--- a/minidump-unwind/src/amd64.rs
+++ b/minidump-unwind/src/amd64.rs
@@ -249,11 +249,14 @@ where
 
     // Breakpad devs found that the first frame of an unwind can be really messed up,
     // and therefore benefits from a longer scan. Let's do it too.
-    let scan_range = if let FrameTrust::Context = args.callee_frame.trust {
-        extended_scan_range
-    } else {
-        default_scan_range
-    };
+    // Additionally, especially with JIT code, using an extended scan range when coming
+    // from a trusted source allows skipping over undetectable JIT frames.
+    let scan_range =
+        if let FrameTrust::Context | FrameTrust::CallFrameInfo = args.callee_frame.trust {
+            extended_scan_range
+        } else {
+            default_scan_range
+        };
 
     for i in 0..scan_range {
         let address_of_ip = last_sp.checked_add(i * POINTER_WIDTH)?;
--- a/minidump-unwind/src/amd64_unittest.rs
+++ b/minidump-unwind/src/amd64_unittest.rs
@@ -530,6 +530,68 @@ async fn test_scan_with_symbols() {
     }
 }
 
+#[tokio::test]
+async fn test_scan_with_symbols_sandwiched() {
+    // Test which makes sure when entering a section of frames without information,
+    // coming from a frame with high confidence can be scanned over correctly.
+    //
+    // This may happen when going from AOT native frames with unwind information
+    // into JIT compiled frames without unwind information. Those frames should
+    // ideally be all skipped.
+    let (mut f, mut stack, mut expected, expected_valid) = init_cfi_state();
+
+    let return_address = 0x00007500b0000110u64;
+
+    let frame1_rsp = Label::new();
+    let frame2_rsp = Label::new();
+    stack = stack
+        .D64(0x00007400c0005510)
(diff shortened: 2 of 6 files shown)


luser:
> Now this is would not be needed, if we extend the validity check to also be able to recognize JITed instructions. That requires another heuristic (e.g. proximity of the instruction addr to the previous instruction addr) or proper consideration of `MemoryInfoList` / `LinuxMaps` streams. With the real minidumps I have, these streams are _not_ available (still looking into why crashpad doesn't seem to add the `MemoryInfoList` stream on Windows, for Linux there is straight up no support implemented).

I can't speak to crashpad, but we do have an abstraction in the `minidump` crate over `MinidumpMemoryInfoList` / `LinuxMaps`: https://github.com/rust-minidump/rust-minidump/blob/32e01a0e54d987025aa64486ebc4154f7f6b16d2/minidump/src/minidump.rs#L151-L169

It's used as part of a few analyses in `minidump_processor`: https://github.com/rust-minidump/rust-minidump/blob/32e01a0e54d987025aa64486ebc4154f7f6b16d2/minidump-processor/src/processor.rs#L596

It does not currently get passed down into `walk_stack` for use in unwinding, however.

Dav1dde:
Yeah, I was just trying to say that even if there was support extending the scan range still makes sense in my eyes.

> It does not currently get passed down into walk_stack for use in unwinding, however.

I plan to take a look at that as well.

Dav1dde:
Updated branch with clippy fixes from master.

gabrielesvelto:
@Dav1dde I apologize the review is taking some time but I have a high-priority task that has been hounding me for the past few weeks. What I'm doing in the meantime is using [socc-pair](https://crates.io/crates/socc-pair) to compare the results from your PR to the existing mechanism to ensure we don't have regressions. It's a bit of a manual process so it's taking a minute, sorry for that.

gabrielesvelto:
I did the testing here and the results were good but I forgot to get back at you, sorry. @Dav1dde could you rebase this and I'll re-run the tests?

Dav1dde:
No worries, it's rebased!

gabrielesvelto:
Thanks! I'll merge this later today as I see no downsides from it, but I spotted [a regression](https://github.com/rust-minidump/rust-minidump/issues/1122) from one of the previous changes and we're investigating it.

Dav1dde:
@gabrielesvelto is this still something you're interested in looking into?

gabrielesvelto:
Yes, sorry I forgot about this. This is fine, in my tests it provides mixed results, sometimes helping and sometimes not. In the case where it doesn't help it usually finds extra frames at the bottom of the stacks that aren't really valid, but this is just an extension of an existing issue. I'll merge this.

PR #1114, opened 2025-06-18
Updated cargo-dist to version 0.28.6 and switched to the astral fork

gabrielesvelto:
Upstream is unfortunately unmaintained as the company making it seems to have closed down. This is a stop-gap fix but we'll have to look for a more permanent solution down the road.


afranchuk (approved):
I'm guessing all of the explicit anonymous lifetimes were a clippy thing?

**EDIT**: I see the commit says it's just from new warnings.

PR #1119, opened 2025-07-22
Fall back to loading the code file if the debug file fails to load.

### LINKED ISSUE #1018: Unresolved symbols with PDB and --use-local-debuginfo
On Windows (amd64), using libraries built with PDBs (/ZI), I cannot get symbols resolved in stack frames: I only get addresses and offsets.

My invocation: `minidump-stackwalker.exe crash.dmp --use-local-debuginfo --no-interactive --output-file=report.log`

I'm no Rust programmer, but after debugging a bit I traced it to a failure to insert symbols into the symbol cache in `wholesym_symbol_interface::Impl::new`, seemingly because `effective_debug_file` returns the PDB filename (which seemed to be the right thing) given `unwind_info: false`. However, later on in `library_info_for_binary_at_path` that PDB fails to be loaded and we just end up in the error case `"Unrecognized file"` in `load_binary_at_location`, so nothing gets ever inserted in the symbols hashmap.

Now if I switch that `unwind_info` flag to `true` and let it load the library, it seems I get symbols resolved correctly (could it be resolved later?). I'm pretty sure that flag is there for a good reason, but I can't get it to work without changing it.

Any idea what could be going on?

afranchuk:
Closes #1018.

### DIFF
--- a/minidump-unwind/src/symbols/debuginfo.rs
+++ b/minidump-unwind/src/symbols/debuginfo.rs
@@ -6,6 +6,7 @@ use framehop::Unwinder;
 use memmap2::Mmap;
 use minidump::{MinidumpModuleList, MinidumpSystemInfo, Module};
 use object::read::{macho::FatArch, Architecture};
+use std::borrow::Cow;
 use std::cell::UnsafeCell;
 use std::fs::File;
 use std::path::{Path, PathBuf};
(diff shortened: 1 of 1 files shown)


minidump-unwind/src/symbols/debuginfo.rs
```diff
@@ -349,50 +395,90 @@ mod object_section_info {
...
+impl<'a> EffectiveDebugFile<'a> {
+    /// Get the file path(s) with debug information for the given module.
+    ///
+    /// If `unwind_info` is true, returns the path that should contain unwind information.
+    fn for_module(module: &'a dyn Module, need_unwind_info: bool) -> Self {
+        // Windows x86_64 always stores the unwind info _only_ in the binary.
+        // It's "okay"ish to use `cfg` here since we expect the debuginfo symbol provider to be
+        // used on the system where the module information was created (i.e., the module ought to
+        // match the cfg).
+        let ignore_debug_file = need_unwind_info && cfg!(all(windows, target_arch = "x86_64"));
```
gabrielesvelto, line 419: Thanks for adding the comment here. This still feels a bit icky but I guess we don't have a way to make this dependency explicit without being able to query how `SymbolLoader` is configured (and that's not available in its API). I still can't think of a scenario in which this could possibly fail.

gabrielesvelto (changes requested):
Just a small change, the rest LGTM

minidump-unwind/src/symbols/debuginfo.rs
```diff
@@ -349,50 +395,90 @@ mod object_section_info {
...
+                    if let Some(parent) = code_file.parent() {
+                        let path = parent.join(&file);
+                        if path.exists() {
+                            return EffectiveDebugFile {
+                                path: path.into(),
+                                fallback: Some(code_file),
+                            };
+                        }
                     }
                 }
```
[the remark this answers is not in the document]
afranchuk: I was hesitant to do that because it will change the behavior. The current behavior will return a relative debug file if it exists and the absolute path doesn't exist, whereas always changing to an absolute path first will obviously remove that. I'm not sure whether that's an important behavior to retain or not (it seems pretty hacky for the binary to have a relative path that would resolve relative to the PWD and not the binary itself, so probably not?).

minidump-unwind/src/symbols/debuginfo.rs
gabrielesvelto: I guess that's doable? It would be a matter of adding a flag to `wholesym_symbol_interface::Impl` (where we instance it we have access to the `MinidumpSystemInfo`) and to `DebugInfoSymbolProvider::build()` so you could pass that bit all the way down to `EffectiveDebugFile::for_module()`. I don't know if it's worth it though.

minidump-unwind/src/symbols/debuginfo.rs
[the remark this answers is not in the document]
gabrielesvelto: Oh right, I hadn't thought about that. Yeah that'd be odd, but let's not add a change of behavior in this patch. We can do it later, disregard my previous comment.

minidump-unwind/src/symbols/debuginfo.rs
luser: So yeah the `Module` trait doesn't currently expose the OS info:
https://github.com/rust-minidump/rust-minidump/blob/1ae2d079d8c6978ca964a5db068d434e8d387ea5/minidump-common/src/traits.rs#L11

but this code is operating directly on `MinidumpModule`s:
https://github.com/rust-minidump/rust-minidump/blob/1ae2d079d8c6978ca964a5db068d434e8d387ea5/minidump-unwind/src/symbols/debuginfo.rs#L180
which do have an `os` field:
https://github.com/rust-minidump/rust-minidump/blob/1ae2d079d8c6978ca964a5db068d434e8d387ea5/minidump/src/minidump.rs#L1109

Annoyingly the `Module` trait is defined in `minidump-common` but the `Os` enum is defined in `minidump`, which is one layer. You *could* add a `fn os(&self) -> Os` to the trait, and even add a default implementation for it that just returns `Os::Unknown`, but you'd have to first move that enum down into `minidump-common`.

Alternately, the only callers of `effective_debug_file` are `wholesym_symbol_interface::Impl::new` and `load_unwind_module`, the former of which actually has a `&MinidumpModuleList` and the latter of which is only called by `DebugInfoSymbolProviderBuilder::build`, which *also* has a `&MinidumpModuleList`, so you could change the type of the module parameter in these functions to be `&MinidumpModule` instead of `&dyn Module` and just call `MinidumpModule::os` directly.

afranchuk: Yeah I'll just do that since it's an internal function.

afranchuk: Ah, but that'd require making `os` a public field. Was there a reason it was kept private?

afranchuk: Also we actually need to know the architecture, too, which isn't so readily available :/

gabrielesvelto (approved):

minidump-unwind/src/symbols/debuginfo.rs
luser: > Ah, but that'd require making `os` a public field. Was there a reason it was kept private?

I can't say I know of any particular reason.

PR #1127, opened 2025-11-03
Fix panic with malformed minidumps when reading object info

Dav1dde:
We encountered this panic in production, with a customer sent minidump. This treats the object info as missing if it is unknown/invalid. Maybe we should handle this with a separate error, but this looks fine to me (?).

[Sentry Issue](https://sentry.my.sentry.io/share/issue/31715566fd874e06a8dc8d4ccde813ce/) with stacktrace (note: need to click to show non in app frames).

### DIFF
--- a/minidump/src/minidump.rs
+++ b/minidump/src/minidump.rs
@@ -1780,18 +1780,21 @@ impl MinidumpHandleDescriptor {
         offset: usize,
         ctx: HandleDescriptorContext,
     ) -> Option<MinidumpHandleObjectInformation> {
-        if offset != 0 {
-            ctx.bytes
-                .pread_with::<md::MINIDUMP_HANDLE_OBJECT_INFORMATION>(offset, ctx.endianess)
-                .ok()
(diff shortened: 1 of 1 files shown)


gabrielesvelto:
Can you tell me what's in `MINIDUMP_HANDLE_DATA_STREAM`? It looks like we might be misinterpreting a stream of `MINIDUMP_HANDLE_DESCRIPTOR` elements with one of `MINIDUMP_HANDLE_DESCRIPTOR_2` elements. The size difference between the two is 8 bytes.

Dav1dde:
The header:

```
[/Users/dav1d/workspaces/rust/rust-minidump2/minidump/src/minidump.rs:1942:9] size_of_header = 16
[/Users/dav1d/workspaces/rust/rust-minidump2/minidump/src/minidump.rs:1942:9] size_of_descriptor = 40
[/Users/dav1d/workspaces/rust/rust-minidump2/minidump/src/minidump.rs:1942:9] number_of_descriptors = 412
```

PR #1128, opened 2025-11-03
Fix 1.91 clippy lints

Dav1dde:
Fixed with: `cargo clippy --workspace --all-targets --no-deps --all-features --fix`


gabrielesvelto:
> Stable of what version? `is_multiple_of` was added in 1.87.
> 
> @gabrielesvelto is there an MSRV? If so I can add it as `rust-version` to the Cargo.toml and fix up the issues which come up. But this will still require effort in auditing all dependencies also adhere to the same MSRV.

Not really, I tend to stick to stable and that's it.

PR #1134, opened 2025-11-10
Bump prost to 0.14

hareshkh:


Cargo.lock
```diff
@@ -1129,8 +1129,8 @@ dependencies = [
  "linear-map",
  "linux-perf-event-reader",
  "memchr",
- "prost",
- "prost-derive",
+ "prost 0.13.5",
+ "prost-derive 0.13.5",
```
hareshkh: Upstream PR here - https://github.com/mstange/linux-perf-data/pull/20/files

hareshkh:
@gabrielesvelto : Will need the linux-perf-data PR https://github.com/mstange/linux-perf-data/pull/20/files to merge and then a new linux-perf-data dep to be released. When we depend on that version of linux-perf-data, duplicate dependency should go away.

PR #1153, opened 2026-05-28
minidump-processor: Fix no-default-features build 

### LINKED ISSUE #1139: minidump-processor does not build with default-features = false
Refiling #1036 because…it isn't fixed, even though the field in question is now called `memory_access_list`. In a brand new `cargo init`-ed crate, adding 

```
[dependencies]
minidump-processor = { version = "0.26", default-features = false }
```

fails to build.

jrose-signal:
Fixes #1139 by making basic accessors on MemoryAccessList always available. An alternative solution would cfg out uses of those accessors instead, but this approach is consistent with future disassemblers being added in the future.

Additionally, adds a CI test invocation to catch regressions in the future.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -141,3 +141,12 @@ jobs:
         with:
           command: test
           args: --workspace --all-features --examples --bins
+
+      # This has to be done on a per-crate basis because minidump-stackwalk depends on
+      # minidump-processor's default features.
+      - uses: actions-rs/cargo@v1
+        env:
+          PWD: ${{ env.GITHUB_WORKSPACE }}
+        with:
+          command: test
+          args: -p minidump-processor --no-default-features --all-targets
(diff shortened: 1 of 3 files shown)


gabrielesvelto:
Sorry for the delay, I'm trying to catch up with the reviews. Could you please put the changes to appease clippy in a separate patch and rebase?

jrose-signal:
Clippy is failing on main. I'm happy to rebase once you fix it, but they're not related to this PR.

gabrielesvelto:
Sure, let me push a fix to main.

gabrielesvelto (changes requested):
`get_thread_instruction_bytes()` needs to be gated with `#[cfg(feature = "disasm_amd64")]` and the now unused parameters to that we pass into it must be marked with `#[allow(unused)]` to quiet warnings. Same for `OpAnalysisError` for which we don't use all the members if `disasm_amd64` is disabled.

jrose-signal:
Oh, warning-clean even in this mode too. Okay, sorry about that, will do!

gabrielesvelto:
Thank you! Merging

PR #1154, opened 2026-06-04
processor: bitflip: add a detractor with off-by-one detection heuristics

### LINKED ISSUE #960: Off-by-one errors can be mistaken for bit-flips
I noticed a small issue with bit-flip detection. Whenever we have a one-past-the-end access to a buffer it can often be mistaken for a bit-flip. We might consider ignoring bit-flips if we detect them in the 16-64 bytes right after a valid mapping.

schopin-mozilla:
When we're reasonably sure the segfault is caused by the common off-by-one programmer error, we can lower the bitflip confidence score to make developers have a second look.

Closes #960

### DIFF
--- a/minidump-processor/src/processor.rs
+++ b/minidump-processor/src/processor.rs
@@ -751,8 +751,20 @@ impl<'a> MinidumpInfo<'a> {
         };
         if let Some((address, bit_range)) = bit_flip_address {
             let memory_op = MemoryOperation::from_crash_reason(&info.reason);
+            let access = info.memory_access_list.as_ref().and_then(|list| {
+                list.iter().find(|access| {
+                    let size = access.size.unwrap_or_default();
+                    let base = access.address_info.address;
+                    (base..base.saturating_add(size as u64)).contains(&info.address.0)
+                })
+            });
+
+            // If we have a crashing memory access, use the base address for
+            // bitflip heuristics, as the exception address would just be a
+            // derivative of it.
+            let access_address = access.map(|a| a.address_info.address).unwrap_or(address);
             info.possible_bit_flips = bitflip::try_bit_flips(
-                address,
+                access_address,
                 None,
                 bit_range,
                 exception_details.context.as_deref(),
--- a/minidump-processor/src/processor.rs
+++ b/minidump-processor/src/processor.rs
@@ -763,6 +763,10 @@ impl<'a> MinidumpInfo<'a> {
             // bitflip heuristics, as the exception address would just be a
             // derivative of it.
             let access_address = access.map(|a| a.address_info.address).unwrap_or(address);
+            // The size of the access that actually faulted, used to gauge whether the crash looks
+            // like an off-by-one rather than a bit flip.
+            let memory_access_size = access.and_then(|a| a.size);
+
             info.possible_bit_flips = bitflip::try_bit_flips(
                 access_address,
                 None,
@@ -770,6 +774,7 @@ impl<'a> MinidumpInfo<'a> {
                 exception_details.context.as_deref(),
                 &self.memory_info,
                 memory_op,
+                memory_access_size,
             );
 
             // If we have an exception context, we can check the registers involved in the
@@ -788,6 +793,7 @@ impl<'a> MinidumpInfo<'a> {
                             // the base address, possibly combined with some offset, is still in
                             // the same memory region).
                             memory_op,
+                            memory_access_size,
                         ));
                     }
                 }
@@ -1508,6 +1514,7 @@ mod bitflip {
         exception_context: Option<&MinidumpContext>,
         memory_info: &UnifiedMemoryInfoList,
         memory_operation: MemoryOperation,
+        memory_access_size: Option<u8>,
     ) -> Vec<PossibleBitFlip> {
         let mut addresses = Vec::new();
         // If the address maps to valid memory, don't do anything else.
@@ -1515,6 +1522,18 @@ mod bitflip {
             if memory_operation.is_possibly_allowed_for(&mi) {
                 return addresses;
             }
+        }
+
+        // The address does not map to accessible memory. Measure how far it is from the nearest
+        // allocation: a fault landing right next to one is more likely an off-by-one than a bit
+        // flip (see `BitFlipDetails::confidence`).
+        let distance_to_closest_mapping =
+            distance_to_closest_mapping(address, memory_operation, memory_info);
+
+        // Quick check for obvious off-by-one
+        match (distance_to_closest_mapping, memory_access_size) {
+            (Some(distance), Some(access)) if distance < access as u64 => return addresses,
+            _ => (),
         }
 
         let create_possible_address = |new_address: u64| {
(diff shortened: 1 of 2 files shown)


schopin-mozilla:
I tested this on 62 recent crashes that had a non-null bitflip score.
Out of those, 55 had the same score, 7 had different scores:

```
4895808b-3337-46fc-8202-276170260604: 0.0 vs 0.125 - DIFF
5dabdf27-bfe3-4291-9c62-6f3610260604: 0.0 vs 0.6625000238418579 - DIFF
5ee12dd8-697c-4118-84de-a24990260604: 0.0 vs 0.625 - DIFF
71bbd3fd-a455-417c-a7cb-136ce0260604: 0.03748574107885361 vs 0.25 - DIFF
7ea36031-2109-456c-8af3-3c14a0260604: 0.6477457284927368 vs 0.6625000238418579 - DIFF
e6e00814-ce07-4fce-919e-1fe180260604: 0.05529674142599106 vs 0.625 - DIFF
ec929bbf-5634-4df6-9e68-c26970260604: 0.0 vs 0.125 - DIFF
```

4 of them had their score drop to 0, with their distance to the closest mapping being literally 1.
The drop from 25% to 3% is because the distance is 96 bytes, so still reasonably close but not clear-cut.
Similarly, the drop from 62% to 5% is due to a distance of 40 bytes.
The slight drop from 66% to 64% is with a distance of 3917 bytes, so it is still conceivably an off-by-one into a very large struct, but it's a bit far-fetched.

gabrielesvelto:
As we discussed offline there's a fundamental change we need to do to get more accurate results: take into account the actual access address and not the address reported in the crash when they differ. This is something that might happen with unaligned accesses straddling page boundaries. Say we're accessing a page that is mapped following one that is unmapped. The access address will be within the first page but the reported crash address will be that of the second page. Using the actual access address makes it very clear that the access was some kind of out-of-bounds access rather than a bit-flip.

schopin-mozilla:
I have added the base address adjustment as a second commit, and also fixed the snapshot-based `test_linux_json_pretty` that was broken by the new fields.

minidump-processor/src/process_state.rs
```diff
@@ -299,6 +305,27 @@ mod confidence {
     // Detractors
     pub const POISON: f32 = MEDIUM;
     pub const ORIGINAL_LOW: f32 = MEDIUM;
+
+    /// One memory page (in bytes), the window over which we think it possible for the access
+    /// to the result of an off-by-one.
+    pub const PAGE_SIZE: u64 = 4096;
```
gabrielesvelto: File an issue and add a corresponding TODO here to revisit this when we enable bit-flip detection on ARM. AArch64 on macOS and iOS uses 16 KiB pages so we'll have to accommodate for that.

minidump-processor/src/process_state.rs
```diff
@@ -299,6 +305,27 @@ mod confidence {
     // Detractors
     pub const POISON: f32 = MEDIUM;
     pub const ORIGINAL_LOW: f32 = MEDIUM;
+
+    /// One memory page (in bytes), the window over which we think it possible for the access
+    /// to the result of an off-by-one.
```
gabrielesvelto: `to the result` -> `to be the result`

gabrielesvelto (changes requested):
OK, a few changes are in order. I like this approach as it does reduce the number of false positives but we need to be more selective with what we consider a bit-flip and what we don't. If the confidence drops below a certain threshold we simply shouldn't consider one potential access as a bit-flip and remove the entry entirely. This has some positive upsides not only for actual buffer overflows where confidence drops to 0 (like in [8716d49d-c09f-44af-8327-3e9330260615](https://crash-stats.mozilla.org/report/index/8716d49d-c09f-44af-8327-3e9330260615) but also for others where the address is bogus and gets detected as a bit-flip only because it happens in a memory range where there's a cluster of pages mapped to each other (e.g. [cef7afda-dd72-4958-9ea3-8c6950260617](https://crash-stats.mozilla.org/report/index/cef7afda-dd72-4958-9ea3-8c6950260617)). Note the `e4e4` pattern in that address, it's one of our poison patterns (for uninitialized memory) and denotes we've read garbage.

Interestingly another pattern that I've noticed indicates that with this tweak the confidence for stack-overflows mistaken for bit-flips drops significantly, see for example [3627d141-43a8-4ca1-868c-e09ef0260618](https://crash-stats.mozilla.org/report/index/3627d141-43a8-4ca1-868c-e09ef0260618). This is indeed not a bit-flip.

Which leads to a few tweaks we have to do to the heuristic: we have guard-page detection already. I've noticed that in many cases where we detect a guard-page the confidence also goes down. This is good but we could just skip bit-flip detection entirely in that case.

Additionally we really need more granularity when establishing if a page is accessible: we need to be able to tell apart read access to readable pages, write accesses to writable pages and instruction fetches to executable pages. The `is_accessible()` call is too broad.

There's another aspect I hadn't considered. We often have accesses in the form base register + index. In  those cases we also tend to mess up things because we check every register and not just the aggregate access. We should evaluate what's the better strategy in that case.

I know this broadens the scope a bit but it has the potential of improving the heuristic far more. However you'll need to test it on more potential candidates, so enter socc-pair. It's a tool that lets you automatically compare your local rust-minidump build with the results on socorro. It's been unmaintained for a while so you'll need to build it from [here](https://github.com/gabrielesvelto/socc-pair) but I'll  try to grab it and fix some of the issues. In the meantime you can use it as follows:

* In a working directory make a bunch of directories that will receive the results: `mkdir comparisons  socorro-cache  symbols  symbols-tmp`
* Now you can run comparisons by pointing `socc-pair` to your local `rust-minidump` checkout and providing it with crash IDs (and your appropriate socorro token). `socc-pair --run-local path/to/rust-minidump/minidump-stackwalk --socorro-cache socorro-cache --symbols-url https://symbols.mozilla.org --symbols-cache symbols --crash-id <crash_id> --api-token <api_token>`
* If you want to do it in batches pipe `supersearch` output into it, like so:
```
URL="https://crash-stats.mozilla.org/api/SuperSearch/?possible_bit_flips_max_confidence=%21__null__"
supersearch --num 250 --supersearch-url "${URL}" | xargs --max-procs=<10-20-should-do> -I{} socc-pair --run-local path/to/rust-minidump/minidump-stackwalk/ --socorro-cache socorro-cache --ignore-field proc_limits --ignore-field frame_count --ignore-field symbol_url --symbols-url https://symbols.mozilla.org --symbols-tmp symbols-tmp --symbols-cache symbols --crash-id "{}" --api-token <api_token> --output-file "comparisons/{}.comp"
```

gabrielesvelto:
Last but not least we should take into account the output of the memory tester, that's a pretty strong signal. For example take [this crash](https://crash-stats.mozilla.org/report/index/305829ab-4c20-4d37-84c1-4cb970260615). We detected a bit-flip and the `MemtestOutput` annotation has `memtest_failed` set to true. We should bump the confidence to 100% in that case noting that we've indeed detected faulty hardware on that box.

schopin-mozilla:
> I know this broadens the scope a bit but it has the potential of improving the heuristic far more.

That's "a bit" of an understatement there!
Rather than trying to broaden the scope of this PR from a narrow, targeted improvement to a generic "improve the bitflip heuristics", can't we instead focus on getting the specific off-by-one patches through, and then tackle the rest of the lot? Most changes you talk about are fairly self-contained.

This would also allow us to actually study the impact of each change independantly.

gabrielesvelto:
> That's "a bit" of an understatement there!

Yeah.

> Rather than trying to broaden the scope of this PR from a narrow, targeted improvement to a generic "improve the bitflip heuristics", can't we instead focus on getting the specific off-by-one patches through, and then tackle the rest of the lot? Most changes you talk about are fairly self-contained.

OK, then let's narrow this down to only detecting accesses happen directly across or immediately after a readable/writable page. So make the `off_by_one()` calculation essentially a boolean where we don't consider the access a bit-flip at all. We'll keep the fuzzier but still interesting distance calculation for later when we have more data from in-the-wild crash reports. Also you should distinguish between read/write accesses and corresponding permissions. I wouldn't bother with executable permissions as an off-by-one instruction read is kind of an odd beast I don't think we'd encounter very often.

> This would also allow us to actually study the impact of each change independantly.

OK, let's dig deeper into the other crashes after you've landed this.

gabrielesvelto (approved):
LGTM, thanks!

gabrielesvelto:
Tests need some fixup before we can merge

schopin-mozilla:
The tests should be fixed, they had an implicit dependency on the disasm
feature, we just skip them now.

PR #1155, opened 2026-06-07
Fix ARM64 FP unwind LR preservation for entry CFI

### LINKED ISSUE #1138: ARM64 frame-pointer unwinding doesn't preserve link register (x30), breaking subsequent CFI unwinding
On ARM64, when frame-pointer unwinding successfully produces a caller frame, subsequent CFI-based unwinding fails because the link register (x30/lr) isn't tracked. This causes stack traces to be shorter with symbols than without symbols.

In [minidump-unwind/src/arm64.rs](https://github.com/rust-minidump/rust-minidump/blob/df483d5774cae0fa1bb847eaa3dd8f4a15c7e9d9/minidump-unwind/src/arm64.rs#L213-L222), the `get_caller_by_frame_pointer()` function only sets three registers as valid for the caller context:

```
  let mut caller_ctx = ArmContext::default();
  caller_ctx.set_register(PROGRAM_COUNTER, caller_pc);
  caller_ctx.set_register(FRAME_POINTER, caller_fp);
  caller_ctx.set_register(STACK_POINTER, caller_sp);

  let mut valid = HashSet::new();
  valid.insert(PROGRAM_COUNTER);
  valid.insert(FRAME_POINTER);
  valid.insert(STACK_POINTER);
```

The link register (x30) is not included. This becomes a problem when the next frame's instruction pointer lands at a function entry point where CFI rules specify:

```
STACK CFI INIT 2380 70 .cfa: sp 0 + .ra: x30
```

This rule says "return address is in register x30". When eval_cfi_expr() in breakpad-symbols/src/sym_file/walker.rs evaluates .ra: x30, it calls:

```
  walker.get_callee_register("x30")
```

Since x30 wasn't marked valid by frame-pointer unwinding, this returns None and CFI unwinding fails. The unwinder then falls back to scan, which often fails or produces incorrect results.

## Example

ARM64 minidump with crash in raise() → crash_abort() → get_response() → main():

```
  Without symbols (18 frames):
  Thread 0 (crashed)
   0  libc.so.6 + 0x8d7b4
   1  libc.so.6 + 0x8d750
   2  libc.so.6 + 0x3e200
   3  crash-service + 0x1908
   4  crash-service + 0x26a8
   5  crash-service + 0x15a8
   ... (continues to frame 17)
```

```
  With symbols (4 frames):
  Thread 0 (crashed)
   0  libc.so.6!raise + 0x54
   1  libc.so.6!raise [raise.c:26]
   2  libc.so.6!abort + 0xdc
   3  crash-service!crash_abort() [main.cpp:65]
```

Unwinding stops at `crash_abort()` because the return address points to `get_response()` function entry (0x26a8), where CFI says .ra: x30, but x30 isn't available.

kuzaxak:
dump_syms 2.3.6+ now emits AArch64 function-entry CFI with .ra: x30, fixing mozilla/dump_syms#744. rust-minidump still dropped x30 after frame-pointer unwinding, so a following CFI unwind at a function entry could not evaluate that return-address rule and the walk would stop or fall back to scanning.

Recover the caller's saved LR from the caller frame record when ARM64 frame-pointer unwinding has a readable caller FP, and mark x30 valid. Also allow a CFI result with unchanged sp when it changes pc, matching function-entry rows whose CFA is still sp+0 while preserving the backwards/same-frame guard.

Regression coverage builds an FP -> CFI chain matching the new dump_syms output.

Fixes: https://github.com/rust-minidump/rust-minidump/issues/1138

Related: https://github.com/mozilla/dump_syms/issues/744

Related: https://github.com/mozilla/dump_syms/pull/745

Issue-tracker context: https://bugzilla.mozilla.org/show_bug.cgi?id=1526052

### DIFF
--- a/breakpad-symbols/src/lib.rs
+++ b/breakpad-symbols/src/lib.rs
@@ -603,6 +603,9 @@ pub trait FrameWalker {
     fn set_cfa(&mut self, val: u64) -> Option<()>;
     /// Set whatever registers in the caller should be set based on the return address (e.g. rip).
     fn set_ra(&mut self, val: u64) -> Option<()>;
+    /// Report where the CFI rule range covering the instruction begins
+    /// (the `STACK CFI INIT` row's start address).
+    fn set_cfi_rules_start_address(&mut self, _addr: Option<u64>) {}
 }
 
 /// A simple implementation of `FrameSymbolizer` that just holds data.
--- a/breakpad-symbols/src/sym_file/mod.rs
+++ b/breakpad-symbols/src/sym_file/mod.rs
@@ -514,7 +514,13 @@ impl SymbolFile {
                     count += 1;
                 }
 
-                walker::walk_with_stack_cfi(&info.init, &info.add_rules[0..count], walker)
+                walker.set_cfi_rules_start_address(Some(module.base_address() + info.init.address));
+                let result =
+                    walker::walk_with_stack_cfi(&info.init, &info.add_rules[0..count], walker);
+                // Reset even on failure so the address cannot leak into
+                // another provider's evaluation.
+                walker.set_cfi_rules_start_address(None);
+                result
             } else {
                 None
             }
--- a/minidump-processor/src/processor.rs
+++ b/minidump-processor/src/processor.rs
@@ -867,27 +867,25 @@ impl<'a> MinidumpInfo<'a> {
             CrashReason::MacArithmeticPpc(MacArithPpc::EXC_PPC_ZERO_DIVIDE)
             | CrashReason::MacArithmeticX86(MacArithX86::EXC_I386_DIV)
             | CrashReason::LinuxSigfpe(LinuxSigfpe::FPE_INTDIV)
-            | CrashReason::WindowsGeneral(ExceptionCodeWindows::EXCEPTION_INT_DIVIDE_BY_ZERO) => {
+            | CrashReason::WindowsGeneral(ExceptionCodeWindows::EXCEPTION_INT_DIVIDE_BY_ZERO)
                 if exception_details
                     .info
                     .instruction_properties
                     .as_ref()
-                    .is_some_and(|p| !p.is_division)
-                {
-                    inconsistencies.push(Inconsistency::IntDivByZeroNotPossible);
-                }
+                    .is_some_and(|p| !p.is_division) =>
+            {
+                inconsistencies.push(Inconsistency::IntDivByZeroNotPossible);
             }
 
             CrashReason::WindowsGeneral(ExceptionCodeWindows::EXCEPTION_PRIV_INSTRUCTION)
-            | CrashReason::WindowsNtStatus(NtStatusWindows::STATUS_PRIVILEGED_INSTRUCTION) => {
+            | CrashReason::WindowsNtStatus(NtStatusWindows::STATUS_PRIVILEGED_INSTRUCTION)
                 if exception_details
                     .info
                     .instruction_properties
                     .as_ref()
-                    .is_some_and(|p| !p.is_privileged)
-                {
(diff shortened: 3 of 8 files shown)


kuzaxak:
@gabrielesvelto Could you please check it? Following the original PR in dump_sys.

gabrielesvelto:
This LGTM but I want to run a comparison on our crashes to make sure there aren't any regressions. I should be done by Friday.

gabrielesvelto:
FYI I'm running this comparison now over around 6k crashes, it took me a bit longer than anticipated.

gabrielesvelto:
The comparison's result were interesting. On Linux I've found consistent improvements across all the dumps I could get my hands on, but on macOS I found some regression which I need to investigate. I'll try to figure out what's going on before merging.

kuzaxak:
@gabrielesvelto could you share some description or summary? So I can try to reproduce myself and try to address.

kuzaxak:
@gabrielesvelto while you're investigating I tried to work out what could regress specifically on macOS, and I found at least one concrete mechanism, reproduced it in a test, and prototyped a fix.

## Suspected mechanism

On macOS, `.sym` CFI comes from compact unwind info (`symbolic-cfi` skips `eh_frame` when `__unwind_info` is present), so every function gets a single whole-function row with no prologue deltas. Frameless encodings and merged `CompactUnwindOp::None` ranges produce rows like:

```
STACK CFI INIT 5eb10 210 .cfa: sp .ra: x30
```

| frame | main | this PR before fix |
|---|---|---|
| 2 | `frame_caller` — FramePointer | `frame_caller` — CallFrameInfo |
| 3 | `outer` — CallFrameInfo | `frame_caller` again — **Scan (duplicate)** |
| 4 | — | `outer` — Scan |

So the signature to look for in your regressed dumps: FramePointer→CallFrameInfo trust flips at frames whose CFI row is `.cfa: sp[ N +] .ra: x30`, followed by a duplicated caller or a scanned/truncated tail.

gabrielesvelto:
> So the signature to look for in your regressed dumps: FramePointer→CallFrameInfo trust flips at frames whose CFI row is `.cfa: sp[ N +] .ra: x30`, followed by a duplicated caller or a scanned/truncated tail.

Yes, I think I saw at least two of those (digging them out now) but I've also seen other issues. For example in [this crash](https://crash-stats.mozilla.org/report/index/fe6f3085-a3dd-4e53-a82b-b931a0260519#allthreads) if you go to Thread 10 once your patch is applied the frame after `nsThreadManager::Init()` is found via stack-scanning instead of CFI.

gabrielesvelto:
In [this crash](https://crash-stats.mozilla.org/report/index/65dc8fbb-e2fb-4bc4-b19d-96f200260521#allthreads) I get exactly what you described. In Thread 4 there's a frame with `CFURLCopyResourcePropertyForKey` found via the frame-pointer. With your patch applied it becomes two frames with `CFURLCopyResourcePropertyForKey` the first found via CFI and the second via stack-scanning.

kuzaxak:
> In [this crash](https://crash-stats.mozilla.org/report/index/65dc8fbb-e2fb-4bc4-b19d-96f200260521#allthreads) I get exactly what you described. In Thread 4 there's a frame with `CFURLCopyResourcePropertyForKey` found via the frame-pointer. With your patch applied it becomes two frames with `CFURLCopyResourcePropertyForKey` the first found via CFI and the second via stack-scanning.

Good, then I will prep a fix not to touch these cases. Will push to the same branch.

kuzaxak:
Done, would be cool if you can run it against the test data you found to see if fix works for all the cases you have.

gabrielesvelto:
Tried on the crashes where I saw the regressions and it's working fine. I have to figure out what's going with the clippy workflow though before we can land.

gabrielesvelto:
This is working fine, merging it.

PR #1162, opened 2026-07-10
feat(minidump-unwind): add configurable UnwindStrategy to SymbolProvider

aholza:
## Problem

`get_caller_frame()` has always tried CFI (call frame information, e.g.
`.eh_frame`/`.debug_frame`) before the frame-pointer chain, hardcoded per
architecture. Debuggers like GDB do the opposite by default: walk the frame
pointer chain first, falling back to CFI. For binaries built with frame
pointers preserved (`-fno-omit-frame-pointer`) whose CFI happens to be
present but inaccurate — e.g. FPO'd system libraries with truncated or
wrong CFI — CFI-first can walk into garbage before falling back, where
frame-pointer-first would have produced a correct unwind immediately.
There was no way for a caller to opt into that ordering.

## Solution

- Adds `UnwindStrategy` (`CfiFirst` — the default, matching prior hardcoded
  behavior — or `FramePointerFirst`) to `minidump-unwind::symbols`.
- Adds `SymbolProvider::unwind_strategy()`, a default method returning
  `UnwindStrategy::CfiFirst`, so existing implementors are unaffected
  unless they explicitly override it.
- The per-architecture `get_caller_frame()` implementations (`amd64`,
  `x86`, `arm`, `arm64`, `arm64_old`) now consult
  `args.symbol_provider.unwind_strategy()` to decide which of CFI or the
  frame pointer chain to attempt first. Stack scanning remains the
  unconditional last resort in both orders — this only reorders the first
  two methods, never removes the fallback chain.
- `mips` has no frame-pointer unwinder, so it's left untouched (CFI, then
  scan) — there's nothing to reorder there.

This is purely additive: no existing behavior changes for any caller that
doesn't override the new default method.

## Motivating use case

Built for [getsentry/symbolicator](https://github.com/getsentry/symbolicator),
which is adding a `frame_pointer_first` config option (see companion PR
[getsentry/symbolicator#TODO]) to better match GDB's default unwinding
behavior when symbolicating minidumps. In one real case, a Windows
minidump's `rpcrt4.dll` frame has FPO'd (i.e. absent/misleading) CFI;
`cfi_first` degrades to `trust: scan` there while `frame_pointer_first`
resolves it via `trust: fp`.

## Testing

- Full `minidump-unwind` test suite (76 tests, all passing).
- New integration coverage in the downstream Symbolicator PR:
  a same-input comparison test asserting both strategies agree when CFI is
  trustworthy, and a regression test against a real Windows minidump
  showing the FPO divergence described above.

### DIFF
--- a/minidump-unwind/src/amd64.rs
+++ b/minidump-unwind/src/amd64.rs
@@ -423,11 +423,23 @@ where
 {
     // .await doesn't like closures, so don't use Option chaining
     let mut frame = None;
-    if frame.is_none() {
-        frame = get_caller_by_cfi(ctx, args).await;
-    }
-    if frame.is_none() {
-        frame = get_caller_by_frame_pointer(ctx, args);
+    match args.symbol_provider.unwind_strategy() {
+        UnwindStrategy::CfiFirst => {
+            if frame.is_none() {
+                frame = get_caller_by_cfi(ctx, args).await;
+            }
+            if frame.is_none() {
+                frame = get_caller_by_frame_pointer(ctx, args);
+            }
+        }
+        UnwindStrategy::FramePointerFirst => {
+            if frame.is_none() {
+                frame = get_caller_by_frame_pointer(ctx, args);
+            }
+            if frame.is_none() {
+                frame = get_caller_by_cfi(ctx, args).await;
+            }
+        }
     }
     if frame.is_none() {
         frame = get_caller_by_scan(ctx, args).await;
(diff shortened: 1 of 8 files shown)


Dav1dde:
As someone working on the mentioned Symbolicator repo. I have yet to see a reason why we would need this.

gabrielesvelto:
We have never experienced this particular behavior, could you provide a concrete example?

PR #1163, opened 2026-07-16
minidump: fix off-by-one in MinidumpLinuxMapInfo::memory_range

schopin-mozilla:
Despite a commentary to the contrary in the code, the end address in /proc/<pid>/maps is exclusive. This had for consequence that a lot of mappings were silently dropped as "overlapping" rather than simply adjacent. That would notably be the case for guard pages.

Luckily, this only happens for minidumps that only have LinuxMapInfo streams but not MemoryInfoList, since the latter has precedence.

Running the new code over 500 Linux crashes compared to reports generated from the mainline resulted in -146/+11 candidate bitflips, the removals being because the base address would now be correctly identified as mapped, while the *new* candidates stem from those candidates having one of their corrected address now landing in a mapping.

### DIFF
--- a/minidump/src/minidump.rs
+++ b/minidump/src/minidump.rs
@@ -2668,11 +2668,11 @@ impl MinidumpLinuxMapInfo<'_> {
     }
 
     pub fn memory_range(&self) -> Option<Range<u64>> {
-        // final address is inclusive afaik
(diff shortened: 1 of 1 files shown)


gabrielesvelto (approved):
Very nice catch

PR #1164, opened 2026-07-20
Gradual off-by-one bitflip confidence detractor

schopin-mozilla:
Originally this branch was about removing low-scoring bitflip candidates, however it turns out that the current heuristics can't score low enough, so I folded it with the rest of the off-by-one work.


schopin-mozilla:
Running stackwalk against ~350 crashes, both using the version in `main` and from this branch, this branch resulted in one crash having all its bitflip candidates discarded, another one having its max reduced from 0.625 to 0.394. Since `main` already has some fixes, and I reduced the size of the detractor window, I'm not terribly surprised.

The crash that had all candidates discarded turns out to be a stack overflow in a third-party DLL.
The other one might very well be a bitflip, though: a candidate lands right next to an address in the registers. I'm happy with that: we just reduced the confidence, but it's still a 39% score!

gabrielesvelto (approved):
LGTM, let's move on from here, gather more data on live crashes then come back to the heuristics for more tweaking if necessary.

PR #1167, opened 2026-08-10
Update the indicatif and  object crates, removed several duplicated and unmaintained dependencies

gabrielesvelto:


afranchuk (approved):
LGTM!

PR #1170, opened 2026-08-31
macos: recognize SIGSYS as a valid error

schopin-mozilla:
We're planning on catching those on the Firefox side (BMO#1724931)

Source: https://github.com/apple-oss-distributions/xnu/blob/f6217f891ac0bb64f3d375211650a4c1ff8ca1ea/bsd/sys/ux_exception.h#L48


gabrielesvelto (approved):
LGTM, thanks

PR #1173, opened 2026-09-21
minidump: don't allow cyclic object info chains

klochek:


minidump/src/minidump.rs
```diff
@@ -1856,8 +1856,18 @@ impl<'a> TryFromCtx<'a, HandleDescriptorContext<'a>> for MinidumpHandleDescripto
...
                     {
-                        object_info_rva = object_info.raw.next_info_rva;
-                        object_infos.push(object_info);
+                        if object_info.raw.next_info_rva == 0
+                            || object_info.raw.next_info_rva > object_info_rva
+                        {
+                            object_info_rva = object_info.raw.next_info_rva;
+                            object_infos.push(object_info);
+                        } else {
+                            // We encountered a non-monotonically increasing offset, so bail.
```
schopin-mozilla, line 1865: Do we know for a fact that jumping backwards denotes a cycle?
FWIW, a sure way to detect a cycle is to have a second iterator that only iterates every other step. If the first and second iterator have ever the same value, you're in a cycle.

klochek: Fair point--I can't find anything for sure that states that backwards offsets are disallowed, and MS's [documentation](https://learn.microsoft.com/en-us/windows/win32/api/minidumpapiset/ns-minidumpapiset-minidump_handle_object_information) makes no claims, either (though it's _probably_ a good bet that they do.) My understanding is that there aren't supposed to be many of these items in a chain (there are only 9 variants,) so I'd be happy to replace this logic with a little HashSet-based visited set.  Thoughts?

